PR #2, opened 2019-03-11
renaming tests to minimize confusion

HDembinski:
I constantly confuse 'test' and 'tests', this helps me.


henryiii:
I like it better. The old_tests will slowly move to tests over time.

PR #3, opened 2019-03-11
implemented generic ND fill

HDembinski:
Generic fill for scalars and arrays. Tests are missing, because I cannot run tests from my build directory. Need to investigate.

You can do any of the following:
```
h1d(1)
h1d([1, 2, 3])
h2d(1, 2)
h2d([1, 2], [3, 4])
h4d([1], [2], [3], [4])
```

### DIFF
--- a/src/histogram/histogram.cpp
+++ b/src/histogram/histogram.cpp
@@ -48,12 +48,15 @@ struct fill_helper {
       ++size;
   }
 
+  // keep function small to minimize code bloat, it is instantiated 32 times :(
+  // TODO: solve this more efficiently on the lower Boost::Histogram level
   template <class N>
   void operator()(N) {
     using namespace boost::mp11;
     // N is a compile-time number with N == arrs.size()
     mp_repeat<std::tuple<double>, N> tp;
-    py::gil_scoped_release gil; // HD: is this really safe?
+    // TODO: only release gil when storage is thread-safe
+    // py::gil_scoped_release gil;
     for (std::size_t i = 0; i < size; ++i) {
       mp_for_each<mp_iota<N>>([&](auto I) {
         // I is mp_size_t<0>, mp_size_t<1>, ..., mp_size_t<N-1>
@@ -77,7 +80,8 @@ struct fill_helper {
       const double* ptr;
       std::size_t size;
     };
-    py::gil_scoped_release gil; // HD: is this really safe?
+    // TODO: only release gil when storage is thread-safe
+    // py::gil_scoped_release gil;
     for (double xi : span{arrs.front().second, size})
       hist(xi); // throws invalid_argument if hist.rank() != 1 
   }
--- a/src/histogram/histogram.cpp
+++ b/src/histogram/histogram.cpp
@@ -15,11 +15,77 @@
 #include <boost/histogram/axis/ostream.hpp>
 #include <boost/histogram/ostream.hpp>
 
+#include <boost/mp11.hpp>
+
 #include <cassert>
 #include <vector>
 #include <sstream>
+#include <tuple>
+#include <cmath>
 
 namespace bh = boost::histogram;
+
+template <class Histogram>
+struct fill_helper {
+  fill_helper(Histogram& h, py::args args) : hist(h) {
+    auto dim = args.size();
+    if (dim == 0)
+      throw std::invalid_argument("at least one argument required");
+
+    arrs.reserve(dim);
+    for (std::size_t i = 0; i < dim; ++i) {
+      auto a = py::cast<py::array_t<double>>(args[i]);
+      if (a.ndim() > 1)
+        throw std::invalid_argument("array dim must be 0 or 1");
+      arrs.emplace_back(a, a.data());
+    }
+
+    size = arrs[0].first.size();
+    for (std::size_t i = 1; i < dim; ++i)
+      if (arrs[i].first.size() != size)
+        throw std::invalid_argument("arrays must have same length");
+    if (size == 0) // handle scalars by setting size to 1
+      ++size;
+  }
+
+  template <class N>
+  void operator()(N) {
+    using namespace boost::mp11;
+    // N is a compile-time number with N == arrs.size()
+    mp_repeat<std::tuple<double>, N> tp;
+    py::gil_scoped_release gil; // HD: is this really safe?
+    for (std::size_t i = 0; i < size; ++i) {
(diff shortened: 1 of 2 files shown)


henryiii:
I'll test the peformance, review, and merge tomorrow (at ACAT). I like the look so far. I might suggest moving to a helper header to keep the files clean.

Run `ctest`, `make test`, or `python -m pytest ../tests/filename.py` from the build dir. I'll add setup.py support and a boost outer folder at some point.

src/histogram/histogram.cpp
```diff
@@ -15,12 +15,78 @@
...
+  // specialization for N=1, implement potential optimizations here
+  void operator()(boost::mp11::mp_size_t<1>) {
+    // compilers often emit faster code for range-based for loops
+    struct span {
+      const double* begin() const { return ptr; }
+      const double* end() const { return ptr + size; }
+      const double* ptr;
+      std::size_t size;
+    };
+    py::gil_scoped_release gil; // HD: is this really safe?
```
henryiii: Yes. See 
https://github.com/astrofrog/fast-histogram/pull/31 for a discussion.

HDembinski:
> Run ctest, make test, or python -m pytest ../tests/filename.py from the build dir. I'll add setup.py support and a boost outer folder at some point.

I know how to run tests :). They don't find histogram.

src/histogram/histogram.cpp
HDembinski: I read the discussion you referenced, but I am not convinced that this is never causing a problem. If you run `histogram::__call__` from two threads, then they could try to increment the same counter at the same time. Without a thread-safe storage, I don't think you can release the gil.

"Shouldn't cause a problem" and "will only rarely cause a problem" are not good enough.

HDembinski: Integer increments are usually not atomic. An increment on the unlimited storage is definitely not atomic.

henryiii: The gil means that Python can't do *anything* else while it is running this code. If you run in threads, this will stop Python from doing anything else completely unrelated to histograms. Most Numpy functions release the gil while running for this reason. If you want to take a mutex here to keep threads from filling the same histogram, you can do so, but don't hold the gil.

henryiii: (And I think for now just not supporting simultaneous fills is fine - I'd rather suspect that pandas doesn't handle `inplace=True` from multiple threads either*)

*: If it does, we really should add the mutex.

HDembinski: Releasing the gil and adding a mutex is a good solution for now.

HDembinski:
I tried `make test` and `pytest ../tests` from the build folder, which both didn't find `histogram`. `pytest` on my system is linked to the pytest from python3.7. The command `python3 -c "import histogram"` works in the build directory.

src/histogram/histogram.cpp
henryiii: Note to self for later: We will need a version that does not add a mutex when we try atomics.

HDembinski: Adding a mutex is not so easy. It should be one mutex per histogram, not just a global mutex for all histograms. So the mutex instance needs to be attached to the histogram instance, better, to the storage. At this point, it is the same work to implement a proper thread-safe storage.

HDembinski: Suggested intermediate solution so that we can move on with this PR: I comment out the gil release for now, adding a comment to put it back in when filling is safe.

henryiii: A gil lock is worse than a global mutex; it globally locks everything in Python. I think for now we should remain non-thread safe so we can test different thread-safe solutions like atomics, with a TODO flag here for now.

HDembinski: If you want to test a thread-safe solution you just uncomment the gil line.

HDembinski: The main development branch should not have known bugs.

henryiii: I can do that. Long running computations locking up the Python interface is really ugly, so we need to fix this ASAP. Normal Pythno task switching does not work inside compiled code.

HDembinski: I don't understand your alarmism. This is development code, it is not to be used in production. Development code foremost needs to be correct, then it can be optimized.

HDembinski: If you make a benchmark with broken code, the result is meaningless. You should also warn astrofrog.

henryiii: Astrofrog makes the output directly, so there is no way a user can access it, it is safe. The only thing you have that Python can access is the input array, which you read only.

henryiii: That is, the output array is created inside the fill function (since it's just a simple fill function). The problem in boost-histogram comes from making the storage first, and then filling it later, since you could fill in two threads. (because it's a histogram object rather than a numpy-like histogram function)

henryiii:
Performance is as good or better than the old method. Nice! The penalty of using a dynamic number of dimensions is reduced now too, so I think we might be able to drop those tuple specializations at some point.

henryiii:
@HDembinski This works for me:

```bash
docker run --rm -it alpine
apk add git g++ cmake make python3-dev py3-numpy pytest
git clone https://github.com/scikit-hep/boost-histogram.git --recursive --depth=50
cmake -S boost-histogram -B boost-histogram/build
cd boost-histogram/build
cmake --build -j .
ctest --output-on-failure # Finds correct python and histogram library
make test # Basically the same as the above line
python3 -m pytest ../tests # Finds correct histogram library
```

The following things fail:
```bash
pytest ../tests # This doesn't check the current directory like `python -m pytest` does.
```

To fix that, you could add a `conftest.py`, but that should not be part of boost histogram because it's going to be designed to be pip-installable.

src/histogram/histogram.cpp
```diff
@@ -15,12 +15,82 @@
...
+    size = arrs[0].first.size();
+    for (std::size_t i = 1; i < dim; ++i)
+      if (arrs[i].first.size() != size)
+        throw std::invalid_argument("arrays must have same length");
+    if (size == 0) // handle scalars by setting size to 1
+      ++size;
+  }
+
+  // keep function small to minimize code bloat, it is instantiated 32 times :(
+  // TODO: solve this more efficiently on the lower Boost::Histogram level
```
henryiii: By the way, the compile is noticeably slower now and the output .so is 10 MB (not sure how large it was before).

PR #5, opened 2019-03-12
Adding CI

henryiii:
This adds CI with Azure. Also adds setup.py support, and some warning fixes.

I'm testing a couple of versions of setup.py Python per platform (macOS, Linux, and Windows), and running the CMake build on a Linux node. The CMake one is a bit trickier to run with multiple Pythons, since its hard to get the python executable path in the recipe in a completely shell independent way.

PR #6, opened 2019-03-12
Set up CI with Azure Pipelines

eduardo-rodrigues:


eduardo-rodrigues:
@henryiii, true that the template provided by Azure is rather simple and you would need to tailor the YAML file to your specific needs. But why do you not want to start from this and rather closed the PR?

Can I be of any help?

I hope you got the correct rights in Azure. I believe I did things right last night.

henryiii:
I think you did set the permissions correctly, though you also have correctly set up the CI so there wasn't anything else to do. I already had a PR based on Particle and CLI11's Azure setup (#5), and am just continuing on that.

eduardo-rodrigues:
OK, all good so long as the CI gets sorted ;-)-

henryiii:
It is about to, should merge in ~5 minutes. :)

PR #8, opened 2019-03-14
Update license

henryiii:
We probably should get this right. The points that need discussion:

- [x] What year range to put?  2019, 2018-2019, or 2015-2019?
- [x] Should this be BSD or Boost or something else? The files claim one thing and the license file is another
- [x] Hans and I should be the copyright holders.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 BSD 3-Clause License
 
-Copyright (c) 2018, Hans Dembinski
+Copyright (c) 2018-2019, Henry Schreiner and Hans Dembinski
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/include/boost/histogram/python/axis.hpp
+++ b/include/boost/histogram/python/axis.hpp
@@ -1,8 +1,7 @@
-// Copyright 2015-2018 Hans Dembinski and Henry Schreiner
+// Copyright 2018-2019 Henry Schreiner and Hans Dembinski
 //
-// Distributed under the Boost Software License, Version 1.0.
-// (See accompanying file LICENSE_1_0.txt
-// or copy at http://www.boost.org/LICENSE_1_0.txt)
+// Distributed under the 3-Clause BSD License.  See accompanying
+// file LICENSE or https://github.com/scikit-hep/boost-histogram for details.
 
 #pragma once
 
--- a/include/boost/histogram/python/histogram.hpp
+++ b/include/boost/histogram/python/histogram.hpp
@@ -1,8 +1,7 @@
-// Copyright 2019 Hans Dembinski and Henry Schreiner
+// Copyright 2018-2019 Henry Schreiner and Hans Dembinski
 //
-// Distributed under the Boost Software License, Version 1.0.
-// (See accompanying file LICENSE_1_0.txt
-// or copy at http://www.boost.org/LICENSE_1_0.txt)
+// Distributed under the 3-Clause BSD License.  See accompanying
+// file LICENSE or https://github.com/scikit-hep/boost-histogram for details.
 
 #pragma once
 
--- a/include/boost/histogram/python/histogram_fill.hpp
+++ b/include/boost/histogram/python/histogram_fill.hpp
@@ -1,8 +1,7 @@
-// Copyright 2015-2018 Hans Dembinski and Henry Schreiner
+// Copyright 2018-2019 Henry Schreiner and Hans Dembinski
 //
-// Distributed under the Boost Software License, Version 1.0.
-// (See accompanying file LICENSE_1_0.txt
-// or copy at http://www.boost.org/LICENSE_1_0.txt)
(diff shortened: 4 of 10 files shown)


HDembinski:
>  What year range to put? 2019, 2018-2019, or 2015-2019?

I would chose 2018-2019. The new binding code is new code under a new license. If the new license only applies to the new code, it should only target the years since when this new code exists.

> Should this be BSD or Boost or something else? The files claim one thing and the license file is another

The BSL allows to incorporate the code into other projects which are licensed under other terms, it is not a license virus like the GPL. So we are free to put boost-python under BSD. The BSL and the BSD license are practically the same in terms and conditions. The choice here can only be made on soft arguments:
- BSD license is a more common wide-spread license than BSL
- BSL allows for easy incorporation into Boost at a later point without changing the license, but that's not our aim

> I believe The Scikit-HEP Project is the best copyright holder?

Why can't we be copyright holders? I like my name attached to a software :)

henryiii:
1. Okay, that's what I like too.
2. Will do BSD then, that's common elsewhere in SciKit-HEP.
3. I believe it promotes more contributions and long term maintenance, as well as is more professional. @eduardo-rodrigues can comment further. We can add an AUTHORS file if you like that. :)

henryiii:
PS: I see @jpivarski is using copyright IRIS-HEP in uproot, I might need to do that? Or both SciKit-HEP and IRIS-HEP?

henryiii:
New proposal committed.

jpivarski:
I didn't have a strong reason to say that the copyright was IRIS-HEP, rather than my own name—I haven't heard good advice on that.

But I'd avoid using SciKit-HEP for a copyright, since it's not any kind of entity. IRIS-HEP, at least, is a funded and recognized project. SciKit-HEP is a brand and a GitHub organization.

An option I haven't considered: HSF? I really don't know.

HDembinski:
All Boost libs have the respective authors as copyright holders, so it is not unprofessional. Technically, the copyright holders are the creators unless we transfer our copyright to another legal entity. In my case, I am developing Boost::Histogram and boost-histogram in my free time. If you, Henry, work on boost-histogram as part of your work time, then your employer is probably the copyright holder. That's usually how it works. It should be explained in your work contract.

henryiii:
With CLI11, I consulted a lawyer and the result was:

```
CLI11 1.7 Copyright (c) 2017-2019 University of Cincinnati, developed by Henry
Schreiner under NSF AWARD 1414736. All rights reserved.
```

Might be best to check with a similar lawyer at Princeton? I'll adjust the current statement to my best guess at the moment.

HDembinski:
> CLI11 1.7 Copyright (c) 2017-2019 University of Cincinnati, developed by Henry
Schreiner under NSF AWARD 1414736. All rights reserved.

This looks very reasonable. Technically, you have to ask your employer whether you are allowed to put code you develop under a work contract as open source, except if there is already a standard agreement that this is so. I went through this process when I wanted to make `pyik` Open Source. I developed this while working at the KIT Karlsruhe, and so they are copyright holders and they had to agree to the BSD license.

HDembinski:
I am ok with being named second after you on boost-histogram.

eduardo-rodrigues:
Hi folks, these matters are tricky and I also do not have a legal statement to provide. In these situations I tend to look at what others have done. Take the case of scikit-learn and astropy, which are also community projects such as Scikit-HEP. They both have friendly BSD licences with

> Copyright (c) 2007–2019 The scikit-learn developers.

or 

> Copyright (c) 2011-2017, Astropy Developers

and these make sense to me. This basically means that the copyright is hold by all code contributors. It has a sense of community, which I like. Had one seen DIANA-HEP or others, why would some physicist contribute knowing that it won't have any rights on the code it commits? Seems very odd to me.

Bottom line, I can only say what I prefer, and that's the above. Reminds me that I need to tweak a bit the particle copyright statement, actually ;-).

HDembinski:
@eduardo-rodrigues I think for a small project such as boost-histogram, it is ok to put the main authors. What you say makes sense when you have a big project, such as numpy, scipy, etc. with hundreds of people working on it. It also makes sense for `hist`, the pythonic, notebook-friendly interface on top of boost-histogram, but not for boost-histogram.

boost-histogram is a very specialized module. It will be written mainly by Henry and myself. It can never grow into a large project with many developers, because it has a very narrow scope. Its interface is restricted to Boost::Histogram.

It is like with full-author-list papers and short-author-list papers. Projects with a narrow focus and a small developer base such as this one are an opportunity for Henry and myself to show off what we can do as developers, when there are no external restrictions.

eduardo-rodrigues:
I certainly buy your opinion for what concerns small projects, @HDembinski. Fair enough.

Now I wonder if I should just write Henry and I for the `Particle` package, as it is also a well-scoped package written (at least so far) entirely by us two.

henryiii:
@HDembinski, please verify that this is fine.

PR #11, opened 2019-03-18
Adding a slighly more comprehensive test

henryiii:
This fixes a lifetime bug (something named `lightweight_view` should have made it pretty obvious that a `keep_alive` was needed!), as well as enables string bins (categories) to be accessed. Still no fill for non-double axis. It would also be nice if there was a trait that could detect a real axis from a category one.

### DIFF
--- a/src/histogram/axis.cpp
+++ b/src/histogram/axis.cpp
@@ -43,7 +43,6 @@ py::class_<A> register_axis_by_type(py::module& m, const char* name, const char*
     .def(py::self != py::self)
     
     .def("index", &A::index, "The index at a point on the axis", "x"_a)
-    .def("bin", &A::bin, "The bin contents", "idx"_a, py::keep_alive<0, 1>())
     .def("value", &A::value, "The value for a fractional bin in the axis", "i"_a)
     .def("size", &A::size, "Returns the number of bins, without over- or underflow")
     .def("extent", [](const A& self){return bh::axis::traits::extend(self);},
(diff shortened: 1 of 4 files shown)


src/histogram/axis.cpp
```diff
@@ -56,10 +55,17 @@ py::class_<A> register_axis_by_type(py::module& m, const char* name, const char*
     
     ;
     
+    // This could be a constexpr if, but no cost for being runtime (since pybind config runs once)
+    if(std::is_same<A, axis::category_str>::value) {
```
HDembinski: More generically, you could check whether the bin method returns a reference and then use keep_alive.

HDembinski: I am not sure why this should have caused you problems. According to docs, the default policy "return_value_policy::automatic" should have worked, although it is not efficient.
https://pybind11.readthedocs.io/en/stable/advanced/functions.html?highlight=keep_alive

henryiii: You need the keep_alive, otherwise when you run, for example, `repr(hist.axis(0).bin)`, the axis may be destroyed before the repr is called, causing it to produce `<bin [inf, inf]>`. It's  harder to reproduce in testing since pytest "remembers" the parts of the expression, but I did come up with a slightly more convoluted test that triggers the problem.

henryiii: > More generically, you could check whether the bin method returns a reference and then use keep_alive.

Is there a way to do that in C++14 that is not deprecated in C++17 and removed in C++20? In Boost perhaps?

henryiii: Nevermind, I got it without using `result_of`.

PR #14, opened 2019-03-26
Henryiii macos 10.14.4

henryiii:
Fixing a warning for macOS 10.14.4, and making warnings as errors optional (really only needed for CI).
Also adding a standard metadata type (could be changed from a string eventually).

### DIFF
--- a/.ci/azure-cmake-steps.yml
+++ b/.ci/azure-cmake-steps.yml
@@ -17,7 +17,7 @@ steps:
 
 - task: CMake@1
   inputs:
-    cmakeArgs: .. -DCMAKE_BUILD_TYPE=Debug
+    cmakeArgs: .. -DCMAKE_BUILD_TYPE=Debug -DBOOST_HISTOGRAM_ERRORS=ON
   displayName: 'Configure'
 
 - script: cmake --build . -j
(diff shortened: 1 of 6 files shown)


HDembinski:
Hi Henry, you could use a Python string directly as the metadata.

henryiii:
An std::string is about the same, and a bit more natural for c++ for now. I need to see if Unicode support for example is better with a python string though.

HDembinski:
? But this is a Python module, why would you want to translate back and forth between a Python string and std::string.

HDembinski:
Unicode will definitely work better with a Python string. And you don't have to convert anything when users construct their axis types.

HDembinski:
This is one of those situations where it is actually useful that you can chose the metadata type statically to whatever you like.

henryiii:
Yes, you are right. Since I don’t need the value in C++ it’s better to use py::str.

henryiii:
I am considering the possibility of allowing our user to attach any py object, as that way you could attach a dictionary of information or something like that, maybe units.  I initially made a custom struct to make sure that I caught all of the places that needed changing.

HDembinski:
That would be cool as well, and it is basically the same performance. Then you would use a pybind11::object instead of pybind11::str, I think.

henryiii:
I will probably rename this to metadata again, just to keep the API is similar to B::H as possible, and because I now allow arbitrary python to be attached.  Hist will probably call this label.

HDembinski:
Yes, I like that also better.

PR #16, opened 2019-03-27
Atomics

henryiii:
This adds `dense_atomic_int` as an atomic, threadsafe storage. It also explicitly makes normal storages not threadsafe, just like Boost::Histogram, allowing users to use their own filling methods, like making several histograms and joining them later.

Names in the C++ code match the Python names a bit more closely, as well.


src/histogram/histogram.cpp
```diff
@@ -49,8 +49,9 @@ py::class_<bh::histogram<A, S>> register_histogram_by_type(py::module& m, const
     .def(py::self + py::self)
     .def(py::self == py::self)
     .def(py::self != py::self)
-    .def(py::self *= double())
-    .def(py::self /= double())
+    // fail on atomics
+    // .def(py::self *= double())
+    // .def(py::self /= double())
```
HDembinski: At some point, code should be added which statistically checks what operations the accumulator/the storage supports and then enable those. This can be done based on boost::mp11::mp_valid.

See
https://github.com/boostorg/histogram/blob/7301d6a4c2d2500245648388ab7a9709985a1f4f/include/boost/histogram/detail/meta.hpp#L137
and the following lines 215, 218, 221, 224, 227.

You would use these here like so:
```
if (has_operator_rmul<histogram_t, double>::value)
   hist.def(py::self *= double());
```

HDembinski: It may be worthwhile to duplicate the code in your library, so that you don't rely on boost::histogram's details.

henryiii:
> until I can provide thread-safe storage

Threadsafe storage would be nice and might support growth and use cases were someone already was working in threads, but it would not provide faster filling on multiple cores; this does.

henryiii:
Also, you might be able to use LHCb's counter API trick:

```cpp
{
auto tmp = hist.local();

tmp(...) // fill the local histogram, does not (yet) change hist

} // On tmp destruction, hist is locked, temp is added to it, then unlocked
```

HDembinski:
> > until I can provide thread-safe storage
> 
> Threadsafe storage would be nice and might support growth and use cases were someone already was working in threads, but it would not provide faster filling on multiple cores; this does.

You misunderstand. I am talking about a thread-safe storage based on atomics, obviously.

HDembinski:
How the thread-safety works in an implementation detail.

HDembinski:
You cannot make a thread-safe storage without additional locking, the atomics are not enough, anyway. A thread-safe storage will combine atomics with a storage-wide mutex. The latter is used when the storage has to be resized, which has to happen sequentially. Other operations which affect all counters at once, like copying the storage and adding another storage or scaling the storage, also need locking that cannot be provided by atomics alone.

HDembinski:
The reason why my PR is progressing slowly is because I am thinking about all these things and I want to implement the thread-safety as an option in the storage_adaptor, so that you can use thread-safe vector-based, array-based, and map-based storages.

henryiii:
> I am talking about a thread-safe storage based on atomics, obviously.

I don't think that was obvious, but that's great.

> the atomics are not enough

Yes, due to resizing, etc.

> The reason why my PR is progressing slowly is because I am thinking about all these things

Waiting for a good design is worth it!

> as an option in the storage_adaptor ... vector-based, array-based, and map-based storages

Very cool!

src/histogram/histogram.cpp
henryiii: I am locked to specific checkouts of Boost::Histogram, so in general this is pretty safe for development for now. Once we hit beta stage or so, I could internalize details, or maybe we could mark this with a comment mentioning that it's used directly by the Python bindings as well, so if someone changes it they know to make the required changes in boost-histogram. ;)

HDembinski: I think it is ok to duplicate the code, to disentangle the two works. Ideally, boost-histogram should only depend on the public interface of boost::histogram. Right now we cannot have that, but for the future it would be nice.

PR #29, opened 2019-04-09
Henryiii pickle

henryiii:
This uses built-in pickle support, combined with Python types, to do serialization.  Standard PyBind11 caveats for picking in Python 2 apply (must be cPickle, must use a reasonably high pickle protocol).

Also provides high quality copy machinery. TODO: add proper deep copy support for histograms (plain axis already fixed).


henryiii:
Not quite ready for merge, just trying to get the CI to run.

PR #31, opened 2019-04-10
introducing try_cast

HDembinski:
try_cast behaves somewhat like a visitor for pybind11::objects.

### DIFF
--- a/include/boost/histogram/python/pybind11.hpp
+++ b/include/boost/histogram/python/pybind11.hpp
@@ -15,6 +15,9 @@
 #include <type_traits>
 #include <sstream>
 
+#include <boost/mp11/list.hpp>
+#include <boost/mp11/algorithm.hpp>
+
 // Allow command line overide
 #ifndef BOOST_HISTOGRAM_DETAIL_AXES_LIMIT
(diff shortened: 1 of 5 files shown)


henryiii:
Two changes that I can do later:

* I don't like `[&]`, it's better to list the items captured explicitly
* This should be in its own header or in the histogram header, since not all .cpp files will use it and it's best to avoid extra work for the compiler if possible. (Plus it is cleaner)

Otherwise, it looks great.

henryiii:
Also, the boost variant includes should be removed. Not that it changes much now, but it's cleaner and better if Boost::Histogram changes to a new variant impl.

HDembinski:
> Two changes that I can do later:
> 
>     * I don't like `[&]`, it's better to list the items captured explicitly

Fair enough. I can fix this.

>     * This should be in its own header or in the histogram header, since not all .cpp files will use it and it's best to avoid extra work for the compiler if possible. (Plus it is cleaner)

Ok, what name do you suggest?

henryiii:
Naming header files after what they have in them is always good. :)

henryiii:
That looks fantastic, thanks!

PR #33, opened 2019-04-12
Comparison and other updates

henryiii:
This adds:

* Support for comparison of axis and histograms with metadata attached
* Support for `flow` in sum
* Using released Boost 1.70 for dependencies
* Warnings are restored, and the warning about C++17 not being supported in Python 2 is now hidden.
* Added benchmarking mode for pytest


henryiii:
@martinRenou, thanks! That's similar to what I've implemented. In my case I've implemented `PyGILState_Check`, so hopefully it will be a cleaner removal when Python 2 support is dropped. :)

PR #47, opened 2019-04-22
Adding shortcut for regular and variable axes

henryiii:
This makes the following equivalent:

```python
bh.histogram((10, 0, 1, "x"), [0, .5, .6, .9, 1.2])
```

```python
bh.histogram(bh.axis.regular(10, 0, 1, "x"), bh.axis.variable([0, .5, .6, .9, 1.2]))
```

A final string can be detected and used as metadata.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,10 +9,6 @@ Python bindings for [Boost::Histogram][] ([source][Boost::Histogram source]), a 
 >
 > Join the [discussion on gitter][gitter-link] to follow the development!
 
-[Boost::Histogram]:        https://www.boost.org/doc/libs/develop/libs/histogram/doc/html/index.html 
-[Boost::Histogram source]: https://www.boost.org/doc/libs/develop/libs/histogram/doc/html/index.html 
-[fastest libraries]:       https://iscinumpy.gitlab.io/post/histogram-speeds-in-python/
-
 
 
 ## Installation
@@ -36,8 +32,8 @@ hist = bh.histogram(bh.axis.regular(2, 0, 1),
                     bh.axis.regular(4, 0.0, 1.0))
 
 # Filling can be done with arrays, one per dimension
-hist([.3, .5, .2],
-     [.1, .4, .9])
+hist.fill([.3, .5, .2],
+          [.1, .4, .9])
 
 # Numpy array view into histogram counts, no overflow bins
 counts = hist.view()
@@ -46,18 +42,27 @@ counts = hist.view()
 ## Features
 
 * Many axis types (all support `metadata=...`)
-    * `bh.axis.regular(n, start, stop, flow=True, growth=False)`: shortcut
+    * `bh.axis.regular(n, start, stop, underflow=True, overflow=True, growth=False)`: shortcut to make the types below. `flow=False` is also supported.
         * `bh.axis.regular_uoflow(n, start, stop)`: `n` evenly spaced bins from `start` to `stop`
+        * `bh.axis.regular_uflow(n, start, stop)`: `n` evenly spaced bins from `start` to `stop` (no overflow)
+        * `bh.axis.regular_oflow(n, start, stop)`: `n` evenly spaced bins from `start` to `stop` (no underflow)
         * `bh.axis.regular_noflow(n, start, stop)`: `regular` but with no underflow or overflow bins
         * `bh.axis.regular_growth(n, start, stop)`: `regular` but grows if a value is added outside the range
     * `bh.axis.circular(n, start, stop)`: Value outside the range wrap into the range
     * `bh.axis.regular_log(n, start, stop)`: Regularly spaced values in log 10 scale
     * `bh.axis.regular_sqrt(n, start, stop)`: Regularly spaced value in sqrt scale
     * `bh.axis.regular_pow(n, start, stop, power)`: Regularly spaced value to some `power`
-    * `bh.axis.integer(start, stop)`: Special high-speed version of `regular` for evenly spaced bins of width 1
-    * `bh.axis.integer_noflow(start, stop)`: `integer`, but with no under/overflow bins
-    * `bh.axis.integer_growth(start, stop)`: `integer`, but will grow if new values are added
-    * `bh.axis.variable([start, edge1, edge2, ..., stop])`: Uneven bin spacing
+    * `bh.axis.integer(start, stop, underflow=True, overflow=True, growth=False)`: Special high-speed version of `regular` for evenly spaced bins of width 1
+        * `bh.axis.integer_uoflow(start, stop)`: `n` evenly spaced bins from `start` to `stop`
+        * `bh.axis.integer_uflow(start, stop)`: `n` evenly spaced bins from `start` to `stop` (no overflow)
+        * `bh.axis.integer_oflow(start, stop)`: `n` evenly spaced bins from `start` to `stop` (no underflow)
+        * `bh.axis.integer_noflow(start, stop)`: `integer` but with no underflow or overflow bins
+        * `bh.axis.integer_growth(start, stop)`: `integer` but grows if a value is added outside the range
+    * `bh.axis.variable([start, edge1, edge2, ..., stop], underflow=True, overflow=True)`: Uneven bin spacing
+        * `bh.axis.variable_uoflow([start, edge1, edge2, ..., stop])`: `n` evenly spaced bins from `start` to `stop`
+        * `bh.axis.variable_uflow([start, edge1, edge2, ..., stop])`: `n` evenly spaced bins from `start` to `stop` (no overflow)
+        * `bh.axis.variable_oflow([start, edge1, edge2, ..., stop])`: `n` evenly spaced bins from `start` to `stop` (no underflow)
+        * `bh.axis.variable_noflow([start, edge1, edge2, ..., stop])`: `variable` but with no underflow or overflow bins
     * `bh.axis.category_int([1, 2, ...])`: Integer bins
     * `bh.axis.category_int_growth([1, 2, ...])`: Integer bins where new items are added automatically
     * `bh.axis.category_str(["item1", "item2", ...])`: String bins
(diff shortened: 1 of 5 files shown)


HDembinski:
I am not happy with this patch. How do you handle the ambiguous case? `bh.histogram((10, 20, 30))`?

HDembinski:
Does this generate a regular axis or a variable axis? Do you want to treat tuples differently from lists?

henryiii:
Yes, tuples are a shortcut for regular axes, and lists are shortcuts for variable axes. No other types are accepted. `histogram((1,2,3,4,5,6))` is an error. If you want to to make a variable axis with a preexisting value, you can use `boost.histogram.axis.variable`. `histogram` itself is just a shortcut after all.

henryiii:
Discussion can still be open; I needed the patch for the readme updates.

henryiii:
The tuple for a regular axis is very handy, and is fantastic for power users to quickly throw together a histogram for testing or quick uses. It's probably not nearly as common to make a variable binned axis, and that's already longer and more typing, so I could be convinced to drop that.

HDembinski:
I guess it is not surprising by now that I find this a terrible idea.

HDembinski:
Python treats all sequence types the same, so you cannot give some sequence type a special meaning. This is going to cause bugs. Less so when you use it directly, but code evolves, stuff you wrote in main gets put in functions, then those functions are refactored. At some point, someone forgot that tuples are handled differently from lists. It is very very unpythonic and error-prone.

HDembinski:
Also, I don't see the benefit. You can already have shorthand.
```
from boost.histogram.axis import regular as r, variable as v

histogram(r(10, 20, 30), v(1, 2, 3))
```

HDembinski:
If `axis.variable` doesn't  accept `*args`, perhaps that's a better direction to improve.

henryiii:
You just wrote 98 characters instead of 34 or 61. And you skipped importing histogram. How is that a shortcut? And now you have used r and v, two of the most common characters to use for radius and values.

Python is *not* C++. It is meant to be used on the command line. The main Python libraries provide ways to do things quickly. For example, `numpy.array` does *not* accept arbitrary iterables; only lists and tuples. You have to use `numpy.fromiter()`. This is because they needed a quick, no mess way to make multiple dimensional arrays. Unless people live and breath histograms, they will want a simple way to make histograms. And if they do live-and-breath histograms, they won't want to type a ton of characters each time.

Making a simple histogram should not require a series of imports. If people can do:

```python
import numpy as np
np.histogram2d(data, bins=(10,10), range=((-1,1), (-1,1)))
```

vs:

```python
import boost.histogram as bh
hist = bh.histogram(bh.axis.regular(10, -1, 1), bh.axis.regular(10, -1, 1))
hist.fill(data)
```

But, compare:

```python
import boost.histogram as bh
bh.histogram((10, -1, 1), (10, -1, 1))
hist.fill(data)
```

henryiii:
If someone doesn't want the shortcut, they don't have to use it. `histogram` is already just a shortcut for making histograms. If you refactor this at some point, you should use `axis.variable`. However, since it might be controversial, I've dropped lists for now, maybe we can add it back later.

HDembinski:
This is again about "making interfaces hard to use incorrectly". And about writing Pythonic libraries. Giving lists and tuples different meaning is not Pythonic, as I have explained before in another context.

HDembinski:
> You just wrote 98 characters instead of 34 or 61. And you skipped importing histogram. How is that a shortcut? And now you have used r and v, two of the most common characters to use for radius and values.

You write the imports only once at the top of your notebook, you don't need shorthand there.

> Python is not C++. It is meant to be used on the command line.

I am well aware of this, but you cannot drop safety and core Pythonic principles to get short syntax. It is a matter of priorities. I think you must agree that "making interfaces safe to use" is more important than anything else. And if you do, then you accept my counter-argument to your proposal instead of reiterating your position, which I already am well aware of. Just reinforcing what you want doesn't invalidate my counter-argument.

The problem is not I don't see the value in what you want. The problem is that you disregard my counter-argument, which you cannot, if we have the same priorities.

HDembinski:
```c++
import boost.histogram as bh
from bh.axis imort regular as r, variable as v

bh.histogram(r(10, -1, 1), v(1, 2, 3))
hist.fill(data)
```
Is also nice and short.

Besides, you are inconsistent with your position. You treat characters like the cost money, but you insist on five superfluous characters `.fill` which are completely avoidable.

HDembinski:
And you complaint that the `__call__` operator is not returning anything is not true anymore. In Boost::Histogram it returns an iterator to the current cell, similar to ROOT.

henryiii:
Using `__call__` hides the fact it's a histogram when you duck-type. If I run:

    h(2)

It's not clear what h is or what it does. However, if I run

    h.fill(2)

It's quite clear that I expect it to be a histogram. You didn't want `h[x]` to be confusable with arrays. It's also clear that h is now probably changed.

henryiii:
I want it to be easy to use interactively. Typing a bit more interactively is okay, your thought flow is still going. If you have to type a bunch of stuff just to set up your environment, then that's a problem. It's why pylab exists, and `from numpy import *`, and other atrocities. It's why numpy and pandas both provide methods to do things `x.sum()` instead of requiring users to `np.sum(x)`; you don't have to set that up if you already have an array and it's easy to find by pressing "tab".

henryiii:
What about this:

```python
from boost.histogram import histogram
from bh.axis import regular as r, variable as v

# stuff

theta = 1.2
r = 2.3

# more stuff ...

histogram = histogram(r(10, -1, 1), v(1, 2, 3))
histogram(r, theta)
```

Do you see the inherent dangers here? I see at least 2. What's worse, the error messages might not be that helpful.

None of this would be a problem in a strongly-typed language like C++. You would not be able to override r, and histogram would have a clearly defined type. (Though overusing auto can be an issue).

henryiii:
> ```python
> import boost.histogram as bh
> from bh.axis imort regular as r, variable as v
> ```

This is not valid python. You can't import from an object. It has to be a fully qualified name.

```python
from boost.histogram.axis imort regular as r, variable as v
```

HDembinski:
> I want it to be easy to use interactively. Typing a bit more interactively is okay, your thought flow is still going. If you have to type a bunch of stuff just to set up your environment, then that's a problem. It's why pylab exists, and `from numpy import *`, and other atrocities. It's why numpy and pandas both provide methods to do things `x.sum()` instead of requiring users to `np.sum(x)`; you don't have to set that up if you already have an array and it's easy to find by pressing "tab".

Again, I am aware of all that. Our conflict is not about me not understanding you. So repeating your position is not helping.

HDembinski:
```
from boost.histogram import histogram
from bh.axis import regular as r, variable as v

# stuff

theta = 1.2
r = 2.3

# more stuff ...

histogram = histogram(r(10, -1, 1), v(1, 2, 3))
histogram(r, theta)
```

The `histogram = histogram(...)` is unrelated to our discussion. And the mistake with `r` will give you a quite useful error message ("no call operator for float" or something like that)

HDembinski:
Using short names is always a danger, fair enough, but what you are suggesting is worse than this.

henryiii:
You just brought `__call__` into our discussion, this is a problem with `__call__` vs. `fill`.

HDembinski:
```
histogram = histogram(r(10, -1, 1), v(1, 2, 3))
histogram(r, theta)
```
does the same as 
```
histogram = histogram(r(10, -1, 1), v(1, 2, 3))
histogram.fill(r, theta)
```
so how is `.fill` helping here?

HDembinski:
Both work as expected, you only get an error when you try to make another histogram afterwards. I don't see how this is related to our discussion.

henryiii:
Are you objecting to having both shortcuts, or just the one? (tuple for regular axis). I have a PR to remove the list <-> variable axis. Then tuples would be a fast way to build regular axes, since that's what most analysts would want fast access to, and variable axes would have to be explicit.

HDembinski:
About regular(...) vs `(...)`, here you are willing to sacrifice safety at the gain of a few characters. This choice makes me think that you don't really care about safety or readability. But when it is about `.fill`, you take the opposite position. Using `.fill` is not safer than `__call__`, it is just a bit more readable. But you don't care about readability, since you want to use `(...)` instead of `regular(...)`. That's inconsistent.

HDembinski:
The point is, I am ok with `.fill` instead of `__call__`, although I don't personally like it and you are introducing a difference between boost-histogram and boost::histogram.

henryiii:
If we don't have the list syntax, what safety am I sacrificing?

HDembinski:
If you drop the list syntax, the conflict is solved. If every sequence is interpreted as a regular axis, that's ok. I still don't like it and I ask you do introduce this to hist and not boost-histogram, but there is not safety concern then.

HDembinski:
It is still bad design, because the regular axis gets special treatment.

henryiii:
The regular axis is the most common type of histogram, especially for quick use cases. It also gets special treatment in other ways (if you only have regular axes, filling is faster). It's what people are going to be familiar with coming into our package. Compare other libraries:

```python
# Numpy
np.histogram2d(xvalues, yvalues, (10, 10), ((-1, 1), (-1, 1)))
# ROOT
ROOT.TH2D("title", "title", 10, -1, 1, 10, -1, 1)
```

to

```python
bh.histogram((10, -1, 1), (10, -1, 1))
```

or

```python
histogram(bh.axis.regular(10, -1, 1), bh.axis.regular(10, -1, 1))
```

henryiii:
Note that numpy gives special treatment to lists and tuples inside the array function. That's because they are making it reasonably easy to make arrays, and keeping the entry barrier low. Try `np.array(i for i in range(10))`; it does not make an array of 10 values. They explicitly check for tuples and arrays (and the `range` object now, apparently) so that it's easy to construct multidimensional arrays. `array([[1]])` should make a length 1 1D array with a python list in it, but it doesn't, because it's more practical to expand these. `np.fromiter` is provided.

PR #52, opened 2019-04-24
Simpler threaded fills

henryiii:
The threaded code should share implementations during the compile for the hard-to-compile part (at least, it's now only 1 class and file instead of 3).

Dropped special usage of `==` for metadata; `==` is now no longer used in threaded fills.

Grouped and improved benchmarking.

Also removes the mention of the shortcut in #47 (still available for my use, but not really a public API at the moment).


henryiii:
I believe this cuts about a minute off the compile time on Azure.

PR #53, opened 2019-04-25
Adding weighted fills

henryiii:
Currently only on accumulators that do not support unweighted fills.

.view and asarray does not work on accumulator storages, and probably cannot, due to the private memory layout and compiler padding; they are not portable. We *probably* need to replace them with a way to copy the contents into python objects. (Providing ways to print contents and access them via `[]` would help a lot).

I might add manual tags to the register histogram functions instead of setting it up behind the scenes; and it would be nice to be able to compose these requirements somehow (threaded, atomic, weights, and someday samples).

Lots of tests need to be added for the weighted fills, serialization, etc.


HDembinski:
fill2 also has partial weight support

henryiii:
#82 replaces the fill method and replaces this PR.

PR #57, opened 2019-04-30
Adding indexed accessor

### LINKED ISSUE #54: Work on indexed accesses
There should be support for the indexed iterator. Something like this:

```python
for item in hist.indexed():
    item.index(0) # current index along first axis
    item.index(1) # current index along second axis
    item.bin(0) # current bin interval along first axis
    item.bin(1) # current bin interval along second axis
    item.content = 3 # Set and read bin contents  
```

This could provide the same shortcuts axes already provide, like `.bins()` and `.centers()`. So, for example, this would be the way you evaluate a function `f(x,y,...)` and fill a histogram with the values:

```python
for item in hist.indexed():
    item.content = f(*item.centers())
```

Also add a way to access the contents in the bins through numpy arrays:

```python
X, Y = hist.centers() # similar to mgrid output
```

This could remove the need to add indexed; the alternative would be, for functions that accept arrays and for histogram storages that can be converted to buffers (see #35 for an idea to allow the more common syntax for setting, and might support different storages):

```python
hist.view() = f(*hist.centers())
```

And, for functions that do not, but still for storages that can be converted to buffers:

```python
it = np.nditer([hist.view(), hist.centers()], [], [['readwrite'], ['readonly']])
with it:
    for (content, center) in it:
        content[...] = f(*center)
```

This would do two unnecessary copies due to the way `'readwrite'` on nditer was written. If you want bin edges, etc., that has to be added to the lists.

henryiii:
* Add tests ✅ 
* `boost::histogram::axis::polymorphic_bin<double>` needs wrapping ✅ 
* Can `ind.centers()`, etc be supported? ✅ 

Will close #54 .

### DIFF
--- a/.ci/azure-publish-dist.yml
+++ b/.ci/azure-publish-dist.yml
@@ -5,5 +5,4 @@ steps:
     artifactName: '$(Agent.OS)_$(Agent.JobName)_$(python.architecture)'
     targetPath: 'dist'
 
-# You can set FULLBUILD_FORCE to anything non-blank to get the manylinux job to run.
 # Artifacts are attached to the build, and are published with an Azure release pipeline.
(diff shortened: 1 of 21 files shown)


henryiii:
The rest of the helpers (just `.centers` is available currently) need to be added, and .indexed might eventually be hidden or removed, but there are several other nice enhancements (like the mentioned .centers), so going to go ahead and merge.

PR #61, opened 2019-05-07
Fill

HDembinski:
This adds an alternative fill strategy. This can be easily extended to support weights, but I would like to see how it fares in benchmarks first before we proceed.


HDembinski:
This is the idea I was talking about. This fill version may be faster for general histograms. I currently cannot test it, because I had to a massive rebase and pytest fails on the new develop branch. But it was working before.

HDembinski:
I am not sure why the tests didn't trigger, I can see no reason why they shouldn't.

HDembinski:
Sorry, not ready for review yet, I just noticed...

henryiii:
Next push here will build on Azure.

henryiii:
Ping me when ready to benchmark.

HDembinski:
All tests pass on my machine now, you can try the benchmark.

HDembinski:
What's there:
- generic fill implementation for ND histograms without the code bloat of the previous solution
- fast 1D implementation, which should be as fast for static histograms and dynamic histograms, making specializations for static histograms obsolete

Weights and samples:
The ND fill implementation has support for weights and samples, but this is untested. The 1D fill implementation does not have it yet, but it should be clear how to add it after a careful reading of the code. I am running out of time to work on this.

HDembinski:
Code is now included in boost::histogram

PR #64, opened 2019-05-08
CMake checks for dev packages when configuring

### LINKED ISSUE #63: all pytest calls fail after rebasing to latest develop
```
python3 -m pytest ../tests/test_histogram_internal.py
ERROR: usage: pytest.py [options] [file_or_dir] [file_or_dir] [...]
pytest.py: error: unrecognized arguments: --benchmark-disable
  inifile: /Users/hdembins/Code/boost-histogram/tests/pytest.ini
  rootdir: /Users/hdembins/Code/boost-histogram/tests
```
The current directory is a `build` folder inside the repo. `make test` also generates failures for each pytest.

My pytest version is 4.4.1. It was working fine on an older checkout of the develop branch.

henryiii:
As requested by @HDembinski, a CMake user will now get an error sooner if they don't have pytest-benchmark.

TODO: Fix the name so that it says `pip install pytest-benchmark` instead of `pip install pytest_benchmark`. Have I mentioned that dashes in package names are irritating? ✅ 

Closes #63.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,7 +124,8 @@ configure_file(
 include(CTest)
 
 function(python_import PACKAGE_NAME)
-    execute_process(COMMAND ${PYTHON_EXECUTABLE} -c "import ${PACKAGE_NAME}" RESULT_VARIABLE RESULT_PRESENT)
+    string(REPLACE "-" "_" PYTHON_PACKAGE_NAME "${PACKAGE_NAME}")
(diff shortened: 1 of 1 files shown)


henryiii:
@HDembinski , does this look fine?

HDembinski:
It would be nice to have something like `find_python_package` that mirrors `find_package`. It would set `X_FOUND` and one can then conditionally add the test targets when all Python packages are available.
```
find_python_package(pytest)
find_python_package(pytest-benchmark)

if (pytest_FOUND AND pytest_benchmark_FOUND)
  enable_testing()
  add_subdirectory(tests) # or similar
endif()
```

PR #65, opened 2019-05-10
Update to new histogram version

HDembinski:
This version has the private variant implementation, which together with a few other things now allows us to drop most of the external dependencies.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,9 +7,6 @@
 [submodule "extern/core"]
 	path = extern/core
 	url = ../../boostorg/core.git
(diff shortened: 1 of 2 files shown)


HDembinski:
Works like a charm on my computer, now I am going to remove the now superfluous externals...

henryiii:
Azure is configured correctly now for PRs from forks. 👍

henryiii:
Let's benchmark before and after 499f429, so see if there's a difference in any-histogram.

Also TODO: Remove the MPL limit setting.

PR #70, opened 2019-05-12
Docker user and group ID fix

henryiii:
@HDembinski, please see if this patch fixes your formatting becomes root-owned issue..


HDembinski:
Works, thank you!

henryiii:
Perfect! It doesn't seem to be needed on macOS, but doesn't hurt either.

PR #71, opened 2019-05-16
replaced copyable_atomic with thread_safe accumulator

HDembinski:


include/boost/histogram/python/histogram.hpp
```diff
@@ -6,59 +6,61 @@
...
+py::buffer_info make_buffer(bh::histogram<A, bh::dense_storage<T>> &h, bool flow) {
+    const auto &axes = bh::unsafe_access::axes(h);
+    auto &storage    = bh::unsafe_access::storage(h);
+    return detail::make_buffer_impl(axes, flow, &storage[0]);
+}
 
-/// Unlimited storage does not support buffer access
-template <typename A>
+/// Unlimited storage does not support buffer access yet
+template <class A>
```
henryiii: Why change to class? https://blogs.msdn.microsoft.com/slippman/2004/08/11/why-c-supports-both-class-and-typename-for-type-parameters/

HDembinski: class is shorter.

HDembinski: class is also visibly more distinct from `template`

HDembinski: mp11 uses class, and I use class in boost::histogram.

henryiii: Type name actually describes what it is though, more closely than class. Class is there for historical reasons. Not a strong preference, I've just mostly seen and use typename for this.

henryiii: Pybind11 uses typename.

HDembinski: I have a strong preference for class and I am already making a compromise by not requiring this project to follow the style of Boost::Histogram. I prefered `typename` when I was less experienced, but when you do a lot of TMP, you realize the benefits of `class` because it is shorter. pybind11 is a good library, but it cannot compete in with mp11 as a guideline, which was written by someone who is way beyond in skill of either you or me.

HDembinski: I would like to merge this and not discuss style. I have another patch which depends on this patch in the pipeline.

henryiii: Okay. But in the future please follow the current style and then make a PR or issue to change the style if you don't like it. A mix of the two is much worse than just one or the other. I'll change this to use class everywhere; as I said, I don't have a strong preference. (And we may end up on the Boost::Histogram style)

The standard library (libstdc++ and libc++) uses both interchangeably, of course...

PR #72, opened 2019-05-21
add buffer interface for unlimited_storage

HDembinski:


henryiii:
This looks great! Thanks!

PR #75, opened 2019-08-15
Cleanup

henryiii:
Currently:

* `typename` -> `class`
* Add underscore to `make_` functions


henryiii:
Let's go ahead and merge this, so we can go forward with the other plans. There's still more cleanup to be done on private type names, but since some types look like they'll be removed, the renames can be done after or during that work.

PR #77, opened 2019-08-25
Update externals

HDembinski:
update boost stuff to 1.71.0 and boost::histogram to develop

### DIFF
--- a/tests/test_public_hist.py
+++ b/tests/test_public_hist.py
@@ -249,11 +249,26 @@ def test_add_2d_w(flow):
             assert h.at(i, j) == 2 * m[i][j]
 
 def test_repr():
-    h = histogram(regular_uoflow(10, 0, 1), integer_uoflow(0, 1))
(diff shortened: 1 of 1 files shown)


henryiii:
Every Sunday, Azure checks out all dev branches and runs a test to see if we are still compatible with head. Looks like this problem started happening just this week. <https://dev.azure.com/scikit-hep/boost-histogram/_build?definitionId=7>

I'll look into it in the next day or two.

HDembinski:
I already worked out a fix. The tests discovered a bug in Boost::Histogram.

HDembinski:
> Every Sunday, Azure checks out all dev branches and runs a test to see if we are still compatible with head.

I think this is not a good idea except maybe for Boost::Histogram. You don't want to live at head for other externals.

henryiii:
Boost Histogram requires the latest head on the other externals; at least it required >1.69 and >1.70. I don't plan to actually update the internal versions of the other externals to head unless it's needed - it's just for the check.

PR #82, opened 2019-08-25
Fill method

HDembinski:


HDembinski:
This is work in progress, but I don't know how to make a "draft" PR.

HDembinski:
I don't understand the MetaFactory, so I cannot fix the code as it is. I think the same can achieved in a simpler and more straight-forward way, with a histogram class defined in Python which forwards method calls to the compiled implementation. This has many advantages, namely that the Python docs can be written in Python.

henryiii:
You can only make a draft PR when you first make a PR; it's a small downarrow next to the create PR button. WIP will work though.

I'll clean up the conflicts tomorrow if you don't have a chance, from the `make_histogram` -> `_make_histogram` rename and `template<typename` -> `template<class` changes.

henryiii:
Looks like this throws a segfault:

```cpp
auto weight = py::object();
py::isinstance<arrayd>(weight); // SEGFAULT here
```

We can work around the bug for now, but probably a bug in PyBind11.

HDembinski:
Thank you for finishing this. This is a great leap forward.

Also nice how the threaded stuff can now be implemented in pure Python.

I think releasing the gil in the fill method is only safe for storages that support concurrent writes?

include/boost/histogram/python/register_histogram.hpp
```diff
@@ -178,82 +178,72 @@ py::class_<bh::histogram<A, S>> register_histogram(py::module &m, const char *na
...
+                auto vargs = bh::detail::make_stack_buffer<VArg>(bh::unsafe_access::axes(self));
+
+                if(args.size() != self.rank())
+                    throw std::invalid_argument("Wrong number of args");
+
+                unsigned iarg = 0;
+                for(auto arg : args) {
+                    if(py::isinstance<py::buffer>(arg) || py::hasattr(arg, "__iter__")) {
+                        auto tmp       = py::cast<arrayd>(arg);
+                        vargs.at(iarg) = tmp;
```
HDembinski: It is safe to use `vargs[iarg]` here, because the line 193 already checks that iarg is within bounds.

include/boost/histogram/python/register_histogram.hpp
```diff
@@ -178,82 +178,72 @@ py::class_<bh::histogram<A, S>> register_histogram(py::module &m, const char *na
...
 
-        ;
+                // TODO handle weights
+                using storage_t = typename histogram_t::storage_type;
+                bh::detail::static_if<boost::mp11::mp_or<std::is_same<storage_t, storage::profile>,
+                                                         std::is_same<storage_t, storage::weighted_profile>>>(
+                    [&](auto &h) {
+                        auto sample = required_arg<py::object>(kwargs, "sample");
+                        finalize_args(kwargs);
+                        auto sarray = py::cast<arrayd>(sample);
```
HDembinski: We probably need to check here again that sarray has the right dimension. The size should be checked by boost::histogram.

include/boost/histogram/python/register_histogram.hpp
HDembinski: But it can also stay as it is, I think, this is not performance critical.

henryiii: I have a strong tendency to use `.at(...)` everywhere; I've spent lots of my early life debugging problems that turned out to be out of bounds accesses (admittedly, there are better tools now to make this slightly better). However, since this is safe due to the check above, it can be changed to `[]`. But, yes, anything that accesses Python should not be a performance critical section. The if check will only waste a few cycles, accessing Python is in the hundreds.

henryiii:
> I think releasing the gil in the fill method is only safe for storages that support concurrent writes? 

No, releasing the GIL is perfectly safe for any segment of code that does not interact with the Python memory model. The GIL is simply a hack that ensures the Python memory model, such as reference counts, never break. Since compiled code does not participate in the Python task switching, any compiled code that takes longer than a millisecond or so should avoid Python memory interactions and release the GIL. It is "safe" because it will not segfault Python's memory management. This is not a solution to code data races, nor was it meant to be; it's there because you *can't* survive a data race in Python's internal memory counting.

Yes, if you try to fill the same histogram from two threads you will get the same result as if you tried to fill the same histogram from two threads in C++. You *could* add a lock stored in every histogram and hold it if the backend is not atomic, but holding the GIL instead is not a valid fix. I feel this is not needed; it is not done in C++ either; it is assumed that a user who can fill from multiple threads will also know to use atomic storage instead of making everyone pay a lock cost on every fill. Same argument for Python.

If you have two threads, one filling a histogram and one doing something completely unrelated, holding the GIL causes the histogram thread to lock the other thread out completely until it's done.

By the way, take a look at Numpy, which does the same thing:

```python
from concurrent.futures import ThreadPoolExecutor
import numpy as np

a = np.zeros(1_000_000, dtype=np.int)
b = np.ones(1_000_000, dtype=np.int)
c = b * 2

def f(a, v):
    a += v

with ThreadPoolExecutor(max_workers=2) as e:
    e.submit(f, a, b)
    e.submit(f, a, c)

print(f'{sum(a):,} != 3,000,000')
```

Result from two runs:

```
2,995,638 != 3,000,000
2,996,028 != 3,000,000
```

So a user needs to understand data races to do threading in Python. Just like C++. Or any other language.

HDembinski:
Thank you for this detailed explanation. If our behavior is the same as numpy then that's good enough for me.

PR #86, opened 2019-09-04
Making includes more specific

### LINKED ISSUE #80: Don't include master header "boost/histogram.hpp"
The master header "boost/histogram.hpp" includes almost all other headers. This causes all files in this project to be recompiled if any header in boost::histogram changes, even if they are not affected by this change. Some people in boost think we shouldn't even provide master headers because of this issue, but they are very useful to quickly get started and seems to outweigh the potential problems. This project should not include the master header, though.

henryiii:
Waiting on boostorg/histogram#209. Fixes #80.

### DIFF
--- a/include/boost/histogram/python/axis.hpp
+++ b/include/boost/histogram/python/axis.hpp
@@ -7,7 +7,7 @@
 
 #include <boost/histogram/python/pybind11.hpp>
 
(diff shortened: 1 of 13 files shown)


HDembinski:
Cool, thanks!

PR #92, opened 2019-09-10
Version 0.4.0

henryiii:
Bumping the version number, fixing a bug, and adding some docs.


HDembinski:
Since this is still open, can we change the default from int storage to double storage, please?

henryiii:
> Since this is still open, can we...

Of course. Final changes are fine here, I'll just merge it before tomorrow's talks.

PR #93, opened 2019-09-12
Enhance fill method

HDembinski:
`fill` now handles int, double, and string input. I fixed the issue in the mean accumulator, which now supports weights. Passing numpy arrays with strings, e.g. `dtype="S10"` is very inefficient, because the code allocates a string from the heap for each entry in the numpy array. This can be avoided, but needs some changes to the fill method in boost::histogram, which will be done later.


HDembinski:
I explained it here #98

HDembinski:
I only commented it out, not remove it completely.

Issue #98, opened 2019-09-13
Add GitHub Actions reformatter


HDembinski:
I want to remove the clang-format check. It has caused me only trouble and it is annoying to see a change fail just because the format check thinks some headers should be ordered slightly differently.

Automation should save time, and so far this is only causing an annoyance.

HDembinski:
I am not willing to accept a solution which requires me and other potential developer to install docker to produce consistent formatting. This is an unacceptable threshold for contributing.

HDembinski:
I think it is acceptance to ask people to install clang-format-8 and use the .clang-format file in their editor. This is what I did. Apparently, there are still formatting differences between my machine and the clang-format check.

I am extremely annoyed by this. White space is not so relevant that it justifies causing so much trouble.

henryiii:
You do not need Docker. You need the `-sort-includes` flag. Clang format by default tries not to break code.

If you don’t like the check, leave the x red, Azure makes this very clear that it is the format check that failed (and I’ll pull, reformat, and push, just as you already made me do). I have done this very successfully with many contributions to CLI11. It does not impair contributions, rather it encourages them. Or write a bot that will reformat for you, like CMake uses. Maybe GitHub Actions will make this easier? Or argue that we should drop sorting, but I like it because it forces nice clear grouping for the cases where order matters, and sometimes finds bugs where the order matters when it shouldn't. But do not kill the whole check!

Proper, reproducible white space males merge conflicts easier, make code easier to read, and speeds up development by not allowing the programmer to chose or change formatting. Most large software I know of that does not have a formatting requirement is trying to add one after years of dev and it’s hard (LHCb, Blender).

Also, *never* assume developers will do anything. That’s why we have CI. I don’t always use an editor that supports auto formatting, for example. Like Xcode or GitHub online.

I take it you are going to hate me when I add a clang-tidy check too...

henryiii:
Just run the format in the scripts directory that does not have docker in the name. Or fix your editor. Or get in the habit of sorting your includes (`:sort` in VI on a selection).

If I add a new include and you add a new include, if they are the same include and are sorted, there will not be a merge conflict.

HDembinski:
I was not aware that my clang-format does not sort the headers. I thought it does. I also like sorted headers, this is not the issue. The issue is that a PR should not fail on the grounds of whitespace. Whitespace is not cosmetic.

[...]

PR #99, opened 2019-09-13
remove clang-format test

HDembinski:
Addresses #98.

### DIFF
--- a/.ci/azure-test-pipeline.yml
+++ b/.ci/azure-test-pipeline.yml
@@ -9,23 +9,6 @@ variables:
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


henryiii:
Having the check forces this project to live up to a higher standard, so I'm not interested in dropping it. I'd accept a PR that adds a (controllable in some way) auto reformatting tool. If this check fails for someone's PR, I'll go in and reformat their PR; this is much better in your git history than patching it after-the-fact. More than two dozen people have contributed to CLI11 and this model has worked just fine. All major projects *require* clean git histories and all formatting checks passing before merge.

PR #100, opened 2019-09-13
Remove dtype argument

HDembinski:
histograms are not numpy arrays and a storage is not a numpy array. The dtype argument is teaching people the wrong ideas about this class.

### DIFF
--- a/src/histogram/make_histogram.cpp
+++ b/src/histogram/make_histogram.cpp
@@ -25,24 +25,14 @@ void register_make_histogram(py::module &m, py::module &hist) {
             py::object storage = optional_arg(kwargs, "storage", py::cast(storage::double_{}));
 
             // Allow a user to forget to add () when calling bh.storage.item
+            // HD: I am very conflicted over this. It is handy, but blurs the distinction between classes and objects.
+            // For pedagogical reasons, I am against that. "Cuteness hurts",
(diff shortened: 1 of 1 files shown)


henryiii:
Okay, agreed.

henryiii:
Note: you can pass a configured storage. It’s very odd to make an instance of a enum-like flag in Python. This just allows both syntaxes to work.

HDembinski:
> Note: you can pass a configured storage. It’s very odd to make an instance of a enum-like flag in Python. This just allows both syntaxes to work.

I see that and I am conflicted about this. The question is whether we want people to regard them as a kind of enum or as a passed configurable instance. They truly are the latter so shouldn't we insist on the right usage now so that it is less confusing later? Right now, none of the storages actually makes use of the possibility to pass a configuration via the storage ctor, but a future mmapped storage might (you want to give the path of the mmapped file).

henryiii:
There are a limited number of pre-compiled storages in Python, and you cannot add your own; this is very much like an enum. Are these run-time configurable or compile-time configurable storages? If the later, it really is an enum-like structure in Python, since you are picking between storages. 

If we have to teach users that this is not an enum-like structure, that adds extra overhead for users to learn - in Numpy, they use `dtype=np.double`, not `dtype=np.double()`, etc.

(You can convince me to drop this, I just currently think it will reduce confusion and make this easier to learn and don't see a major downside)

PR #105, opened 2019-09-16
Rename src files to match function

### LINKED ISSUE #101: Rename src/histogram/accumulators.cpp to src/register_accumulators.cpp etc
It is generally a good idea to call a file after the class or function if the file only includes a single class or function. For me it would be much more logical to rename accumulators and the other implementation files to register_something so it matches the functions in these files.

henryiii:
Fixes #101.


HDembinski:
Thank you, I had a patch for this, so I could have done it myself, but ok.

Issue #102, opened 2019-09-16
Remove storage interface


HDembinski:
The storage classes have some interface exposed in Python. I would like to remove this. Neither users nor our tests should touch the storage directly. The storage implementation is tested in boost::histogram, so there is no need to test it in Python. Furthermore, storage functionality can always be tested indirectly by testing a histogram that uses this storage.

I don't want any interface exposed on the storages except their constructor. Exposing more is giving users wrong ideas. In general, we have to be stingy with public interface, because everything that is exposed as public we cannot change easily later. Therefore, it is better to expose as little as possible, so we maintain as much flexibility as possible in changing things.

henryiii:
Agreed, probably was just trying to wrap as much as possible at that point.

henryiii:
> Therefore, it is better to expose as little as possible, so we maintain as much flexibility as possible in changing things.

:+1:

henryiii:
Actually, the reason this is here is to allow storage pickling to be unit-tested in Python. Pickling is not tested by boost::histogram, and testing a histogram that uses the storage is an integration test, not a unit test. I can drop it, but I like being able to do the unit test. How about renaming `push_back` to `_push_back`?

HDembinski:
If you test the pickling of the histogram, it also implicitly test the pickling of the storage. That would be my preference if there is no issue with that I cannot see at the moment.

HDembinski:
> testing a histogram that uses the storage is an integration test, not a unit test.

Where you draw the line between these is a bit arbitrary. The histogram class itself is still a unit. The storage is a replaceable component of the histogram, but it has no use on its own. In my understanding, an integration test is about interactions between different independent classes. There is no interaction here, because a storage is a component, it is not an independent class.

Despite the semantics, I

henryiii:
How about removing the `__set/getitem__` and making those private too (would just need a single `_get`)? Since serialization treats storage as a separate unit, I really would like some way to test it without testing all the rest of histogram too. We save time if we know what is broken immediately from the test.

henryiii:
Actually, is there a way to get the storage from a histogram? If we could have `_storage`, then I could pop the storage off and test picking it; would only need `==`.

henryiii:
If not, I'll kill the test and just not have a unit test for storage pickling. The current one doesn't work for unlimited storage, which makes it less useful as a unit test.

henryiii:
[...]

PR #106, opened 2019-09-16
Make push_back private

henryiii:
May address #102.


henryiii:
Should have renamed the PR before merging...

Issue #90, opened 2019-09-10
Nice reprs


henryiii:
We need to provide nice reprs for histograms (and a few other objects need nicer reprs that the current default PyBind11 ones, too). The repr can use Numpy to format a preview of the histogram.

- [x] Have nice, Python-runnable reprs for axes
- Find some way to have a nice bins view or repr (handled a bit differently, but valid)
- [x] Have a nice repr for histogram, maybe also Python runable

henryiii:
This is pretty good in 0.5.2. May receive future work, but the basics listed here are done.

Issue #104, opened 2019-09-16
Rename storages


HDembinski:
I think that a type called `storage::double_`/`storage.double` looks quite weird in C++ and in Python. Since in C++ the storages are not put in a sub-namespace, they shouldn't be in boost-histogram, either. I would like to rename the C++ type to `double_storage` and likewise the corresponding Python type. The storages shouldn't be in their own submodule, because that's not how it is in C++ either. To not put them in a namespace was an intentional choice in C++. Whether it really was the right choice I don't know, but I want boost-histogram to be consistent with boost::histogram.

henryiii:
We cannot reproduce the C++ interface here, because we must be fully compiled. The mapping is `bh::dense_storage<double>` == `bh.storage.double`, which is clear and simple, IMO. This also allows you to ask which storages are available in Python, which is `dir(bh.storage)` or `bh.storage.<tab>`, which is something that doesn't even make sense in the C++ version, as it is templated and you can make your own. We could rename "storage" to "dense_storage" to better mimic C++, but I vote for the shorter and simpler name (could be convinced otherwise).

Please ignore the `double_` in C++, that is just to stay consistent and reduce complexity in C++ bindings, it is not user facing.

henryiii:
By the way, I recommend always starting with the general term, then the more specific one; I really regret every time I've not done this. Alphabetical sorting should put similar items together. For example, I used `azure_test_pipeline` and `azure_build_pipeline`, and now if I do an `ls`, they are not together. `azure_pipeline_*` would have been a much better scheme!

HDembinski:
Ok, I see the point of `dir(bh.storage)` in Python. It still feels awkward to me to call something that is a whole container like one element of that container, e.g. `bh.storage.double`. I would prefer for educational reasons to have `bh.storage.double_storage` or `bh.storage.storage_double`. Another option which would also feel a bit less awkward is to use `bh.storage.dense_double`.

HDembinski:
Just to clarify, I think the name `bh.storage.double` is ok, but if `storage` is a submodule, you can also do `from bh.storage import double` which I find very awkward.

HDembinski:
> I recommend always starting with the general term, then the more specific one

I agree in principle. I had some reasons why I didn't do it for storage, but I don't recall them right now.

henryiii:
[...]

PR #108, opened 2019-09-17
Python-ready repr

henryiii:
Fixes part of #90; the repr's are now stable and provide Python-ready syntax. This does not add a quick way to preview the contents of histograms or the bin edges, might be in a later PR.

### DIFF
--- a/include/boost/histogram/python/axis_ostream.hpp
+++ b/include/boost/histogram/python/axis_ostream.hpp
@@ -1,0 +1,191 @@
+// Copyright 2018-2019 Henry Schreiner and Hans Dembinski
+//
+// Distributed under the 3-Clause BSD License.  See accompanying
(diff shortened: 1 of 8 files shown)


henryiii:
(I forgot to run the python formatter; this is why I need a formatting check!)

henryiii:
I'm going to merge this soon, but @HDembinski please let me know if you have any comments, we can always change it. I can't work on #104 until this is merged.

henryiii:
Always == before release, anyway. And, in this case, repr's are not guarantied to be steady between releases (as LHCb found out, they were depending on the repr of an float, which changed in Python 2.6 or 2.7)

HDembinski:
I would have implemented the `__repr__` in Python, but the repr string for the axes looks good and it `eval`s back to the original class, which is nice. For the histogram, we can't have it `eval` back, so we are more free in the format.

PR #109, opened 2019-09-18
Rename axes::any

### LINKED ISSUE #103: Renaming axes::any
I don't like the name `axes::any`, because the actual type is `std::vector<axis::variant<...>>`. I would prefer the name `axis_vector`. I think we can call this any way we like, because it is not exposed to Python. The problem is that `axes::any` does not clarify that this is a vector.

henryiii:
Fixes #103

### DIFF
--- a/include/boost/histogram/python/register_histogram.hpp
+++ b/include/boost/histogram/python/register_histogram.hpp
@@ -247,7 +247,7 @@ py::class_<bh::histogram<A, S>> register_histogram(py::module &m, const char *na
                 namespace bmp = boost::mp11;
                 static_assert(
                     bmp::mp_empty<bmp::mp_set_difference<
-                        bmp::mp_unique<bmp::mp_transform<::detail::get_axis_value_type, bmp::mp_first<axes::any>>>,
+                        bmp::mp_unique<bmp::mp_transform<::detail::get_axis_value_type, axis_variant>>,
                         bmp::mp_list<double, int, std::string>>>::value,
                     "supported value types are double, int, std::string; new axis was added with different value type");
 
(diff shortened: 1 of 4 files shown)


henryiii:
@HDembinski What about having `axis_variant` and `std::vector<axis_variant>` or `vector_axis_variant`?

include/boost/histogram/python/register_histogram.hpp
```diff
@@ -246,9 +246,10 @@ py::class_<bh::histogram<A, S>> register_histogram(py::module &m, const char *na
 
                 namespace bmp = boost::mp11;
                 static_assert(
-                    bmp::mp_empty<bmp::mp_set_difference<
-                        bmp::mp_unique<bmp::mp_transform<::detail::get_axis_value_type, bmp::mp_first<axes::any>>>,
-                        bmp::mp_list<double, int, std::string>>>::value,
+                    bmp::mp_empty<
+                        bmp::mp_set_difference<bmp::mp_unique<bmp::mp_transform<::detail::get_axis_value_type,
+                                                                                bmp::mp_first<axis_variant_vector>>>,
```
henryiii: I think you could drop the first here if you could directly access `axis_variant`

HDembinski: Ok, having `axis_variant` separate also makes sense.

henryiii:
If you like `axis_variant_vector` better than `vector_axis_variant `, I can change it. I slightly like this one better due to it reading like `std::vector<axis_variant>`. We could even drop this typedef, since it's not needed for very much or much simpler than not having it.

Will merge and change later if needed, to progress with the module rename.

PR #110, opened 2019-09-20
Moving boost.histogram to boost_histogram

henryiii:
The inner components are now protected inside "core"; hidden types are not even accessible except through `core`. It's easier to inject Python now.

CMake recommended version is 3.14 (`file(CREATE_LINK`). The `boost_histogram` contents are symlinked into the build directory, to allow the python files to be edited without rerunning CMake. However, 3.12+ still works, older versions copy instead of link.

Tests are a little better about using the public interfaces now.

Now supports the CMake install target in case someone wants to use that.


henryiii:
Given that this is quite intrusive, let's get it in sooner rather than later then polish as needed.

PR #114, opened 2019-09-21
simpler fetching of version string

HDembinski:


henryiii:
I really don’t like exec, it adds a security risk if the file is compromised and will trigger warnings in some systems.

We could use setup.cfg if that’s common enough to use now.

henryiii:
It looks like exec is allowed here. Requests, the most downloaded python package ever, does it this way.

setup.py
```diff
@@ -17,17 +16,8 @@
...
-def get_version(version_file):
-    with open(version_file) as f:
-        contents = f.read()
-    mo = RE_VERSION.search(contents)
-    if not mo:
-        raise RuntimeError("Unable to find version string in {}.".format(version_file))
-
-    return mo.group(1)
+# get __version__
+exec(open("boost_histogram/version.py").read())
```
henryiii:
```suggestion
about = {}
with open("boost_histogram/version.py"):
    exec(f.read(), about)
```

Then use `about.__version__`. We also need to build the path rather than assuming that this is being run in the current directory (but that's a problem with the current version too).

See https://github.com/psf/requests/blob/0357b0724bd01cc89169fd40ca0132cf3248aff3/setup.py#L60-L62

HDembinski: Sorry, I neither see the advantage of using with here nor of using `about`. This is seems excessively pedantic. You have a point with the path, I am open to suggestions.

HDembinski: And yes, exec is safe here, because we are calling our own code. It is only a security risk when the code is from an untrusted source.

henryiii: There is no guarantee in Python that the file gets closed on this line. Or any line. Python is garbage collected, and the garbage collector may not run every line. Or at all. We can commit it and I can update it later with the Request's like solution.

henryiii: I know it's safe, but some code checking tools (like ones that check for security issues) flag things like uses of exec. Obviously, they are smart enough to handle this one, since tools like request use it, so we are fine here. That's what I was worried about.

henryiii: I would like to use best practices here (and mimic what I saw in `requests`), since someone may later use our setup.py as an example. Also, not expliclty closing a file will throw a `ResourceWarning` if they are enabled in Python 3.2+. Usually they are enabled by things like `unittest`, so are not likely to show up in setup.py, but still enough to make avoiding a with-less open always a recommendation: https://stackoverflow.com/a/8011863/2402816

PR #116, opened 2019-09-22
axis updates

HDembinski:
* removed typetools.hpp, since remove_cvref can be replaced with std::decay_t everywhere
* histogram.rank is a property
* axis.size is a property
* axis.extent replaces axis.size(flow=true)
* axis is iterable
* axis.bin(...) returns tuple or value
* axis.index and axis.value vectorized for category axis


setup.py
```diff
@@ -17,17 +16,8 @@
...
-def get_version(version_file):
-    with open(version_file) as f:
-        contents = f.read()
-    mo = RE_VERSION.search(contents)
-    if not mo:
-        raise RuntimeError("Unable to find version string in {}.".format(version_file))
-
-    return mo.group(1)
+# get __version__
+exec(open("boost_histogram/version.py").read())
```
henryiii: You've mixed PRs. So you'll need to merge the other one first.

PR #118, opened 2019-09-22
Setup updates

henryiii:
This follows up on #14 

* `with` used for reading file
* `exec` used with a dictionary - this makes it clear to humans and IDEs where `__version__` comes from - the short form of exec is like `from x import *`
* Paths are now all compared to setup.py's location rather than the current working dir.
    - I used a simple shim to keep from adding pathlib2 as a dependency for Python2 but still keep the syntax reasonable, IMO.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,9 +1,25 @@
+from __future__ import division
+from setuptools import find_packages
 from setuptools import setup, Extension
 from setuptools.command.build_ext import build_ext
-import sys
 import setuptools
-from setuptools import find_packages
+import sys
+import os
 
+# Change to using pathlib when Python2 support is dropped
+class Path(str):
+    def __truediv__(self, s):
+        return self.__class__(os.path.join(self, s))
+
+
+# Useful paths
+BASE_DIR = Path(os.path.abspath(os.path.dirname(__file__)))
+VERSION_FILE = BASE_DIR / "boost_histogram" / "version.py"
+README_FILE = BASE_DIR / "README.md"
+SRC_DIR = BASE_DIR / "src"
+EXTERN_DIR = BASE_DIR / "extern"
+
+# Official trick to avoid pytest-runner as requirement if not needed
 needs_pytest = {"pytest", "test", "ptr"}.intersection(sys.argv)
 pytest_runner = ["pytest-runner"] if needs_pytest else []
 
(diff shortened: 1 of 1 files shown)


henryiii:
Draft will be removed when I've tested this a bit. @HDembinski, please let me know if this looks reasonable.

henryiii:
I hate it when packages check types exactly rather than using `isinstance` or (better) duck typing! Looks like they fixed it in Python 3.

HDembinski:
1) I like `with` very much in general, but I think it is superfluous when the body of the `with` consists of a single statement.
```
with open(file) as f:
   do_something(f.read())
```
is functionally equivalent to
```
do_something(open(file).read())
```
Once the file has been read, the refcount of the file object goes to zero and Python closes the file. This is a popular one-liner.

Since the two versions are functionally equivalent, it is a matter of style which one to chose. I use `with` when it gives me a benefit, and I use the oneliner when it is appropriate. That's why I am not happy that you change this, but I will not ask you revert. Note that I will continue to write oneliners when they are appropriate.

2) Regarding `exec`. The comment perfectly says what this command does, it loads `__version__`. The file loaded is under full control by us and will never contain anything other than the `__version__` variable. It is therefore absolutely fine to load this variable directly into the globals. The additional `about` dict is not necessary and just adding an unnecessary layer of indirection.

HDembinski:
It is a little thing, but I can argue that the one-liner is better, because it deletes the file object. The `with` statement does not delete the file object.
```py
>>> with open("foo") as f:
...    pass
...
>>> f
<_io.TextIOWrapper name='foo' mode='r' encoding='UTF-8'>
```

henryiii:
#### 1

No, it is not functionally equivalent to `with`. The with statement promises to close the file when you leave the bock. The later statement depends on an *implementation specific detail* of CPython to delete f and the end of the line. Python is not C+; it's garbage collected. When the ref count goes to 0, then the next pass of the garbage collector will delete the file, which will call the `__del__` of the file, which will close the file. A good way to think about it is this: A perfectly valid implementation in the Python language of the garbage collector is a computer with infinite memory.

And, while generally the current versions of CPython are configured to run the garbage collector once per line, *it is configurable*! If you are a performance freak, maybe you increase that to once per second, and then this fails.

> This is a popular one liner

Which is why they added a warning in Python 3 when this is encountered. If you run this with a Python that was built in Debug Mode, this will throw a ResourceWarning. You should not have warnings in your `setup.py`!

In general, you should *never* count on Python to delete anything. In fact, if you look at the Python docs, you'll see a lot of caveats, including this scary one: "It is not guaranteed that __del__() methods are called for objects that still exist when the interpreter exits." (Python 2 docs, they may have improved this one around 3.5-3.7.)

The fact that f has not been deleted doesn't matter, since it has been closed.

#### 2

I assume you have not tried to open `setup.py` in PyCharm, the most popular Python IDE? Or run this with mypy, the Python Software Foundation's type checker? Human readable comments require a) Humans and b) Humans reading comments. 

(PS: result from mypy: `setup.py:118: error: Name '__version__' is not defined`)

Both of these part of a previously validated pattern (The PSF's own Requests package). Why try to change them? We should only change this if there is a reason we are doing something different than what they are, or if they are doing something wrong due to better solutions now being available; neither of which are the case.

HDembinski:
> No, it is not functionally equivalent to with. The with statement promises to close the file when you leave the bock. The later statement depends on an implementation specific detail of CPython to delete the file at the end of the line. Python is not C+; it's garbage collected. When the ref count goes to 0, then the next pass of the garbage collector will delete the file, which will call the __del__ of the file, which will close the file. But the garbage collector does not have to run. A good way to think about it is this: A perfectly valid implementation in the Python language of the garbage collector is a computer with infinite memory.

I know all this and I still think you make a huge issue out of something that is not practically relevant. Do you think people run the setup.py with a customized garbage collector? And even if they do and the fill remains open a little longer, this is not a resource issue. There are only two extra files opened here.

> Which is why they added a warning in Python 3 when this is encountered. If you run this with a Python that was built in Debug Mode, this will throw a ResourceWarning. You should not have warnings in your setup.py!

No user runs setup.py in a Python interpreter built in Debug Mode.

> I assume you have not tried to open setup.py in PyCharm, the most popular Python IDE? Or run this with mypy, the Python Software Foundation's type checker? Human readable comments require a) Humans and b) Humans reading comments.

No, and these tools obviously produce a false positive here.

HDembinski:
The attitude in the Boost community is that automated code checking tools should not be followed blindly. That's why I am also not a fan of turning on every possible warning, if the code gets less readable in turn. Warnings are good for beginners who don't know how the language works. For experts, who know what they are doing, they are mostly just obfuscating code.

HDembinski:
Even if `__del__` is not called when the program exists, you don't have to worry about file descriptors remaining open. When the program exists by any means, all resources are freed.

https://unix.stackexchange.com/questions/430365/what-happens-to-file-descriptors-when-the-process-is-killed

This answer is for unix, but I am confident that OSX and Windows do it the same.

henryiii:
> I know all this and I still think you make a huge issue

My point is this is bad Python, and we should not ever allow bad Python if we know it is bad.

> No user runs setup.py in a Python interpreter built in Debug Mode.

Currently, it's really hard to use debug-mode Python at all because you have to rebuild all extensions. Python 3.8 will finally fix that issue, so debug mode may become more common. I could at least see a user debugging a Numpy memory issue in a debug mode Python, so maybe someday someone will try it with boost-histogram. 

My point though was that this "holdover from pre Python 2.6 days" was bad enough that they added a warning for it.

> you will do ctrl+F and search for it.

I would rather the *code* tell me what it is doing than the *comments*. Comments do not run, so might get moved/invalidated/deleted. Again, I don't want someone to read this and think raw exec is okay in their code. It's much nicer to read a file into a dictionary than into the globals.

In Python, you can almost always trace down every variable to its origin, unless you use a few highly discouraged tools, like `from x import *` and `exec(x)`[^1]. They are very useful in the REPL but should not be used in files. Consider:

```python
from piggiltysquat import *
y = sum([3])
```

You can't tell where `sum` comes from; it could be `builtins`/`__builtin__` or `piggiltysquat`. This is not so important in languages like C++ where overriding something produces an error at compile time. (PS: Ruby also does this but doesn't have this excuse).

I don't believe either change will likely affect users (+/- maybe a warning), but they affect readers and are the right thing to do. Saving lines of code is never worth sacrificing correctness or clarity.

> When the program exits by any means, all resources are freed.

By the same argument, allocated memory is returned when a process is killed. So (C++) it's okay to not have a `delete` for every `new`...

[^1]: Yes, I am aware that you can inject variables into the global namespace. But that is rare and for special cases only and usually evil.

```python
>>> globals()["x"] = 3
>>> x
3
```

PR #119, opened 2019-09-23
Starting axis options

henryiii:
Something like this is what I'm thinking. I'm not fond of namedtuple for this use case due to *some* of the reasons listed [here](http://www.attrs.org/en/stable/why.html#namedtuples), though the current design I've used is similar and could switch to using them. I can't think of anywhere in the standard library they are used like this; normally the Python standard library would use a [SimpleNamespace](https://docs.python.org/3/library/types.html#types.SimpleNamespace), but that's Python 3.3+ only.

NamedTuples are good for functions that return multiple values and maintaining backward compatibility, but they are inherently structured, so support unpacking, comparing, iteration, etc. They are also almost the only place in Python where `_X` names are not intended to be private (due to potential clashes with possible names)!

Needs test changes, etc.

Issues/changes:
* this is no longer static, since properties are usually only available on instances (`ax.__class__.options.overflow` does not work while the equivalent did before)
* this does not trigger the auto completion in iPython (`ax.options.<tab>`), since it is dynamic

### DIFF
--- a/include/boost/histogram/python/axis.hpp
+++ b/include/boost/histogram/python/axis.hpp
@@ -85,7 +85,7 @@ py::array to_values_impl(const A &ax,
     for(auto i = -underflow; i < ax.size() + overflow; i++)
         result.mutable_at(static_cast<std::size_t>(i + underflow)) = ax.value(i);
 
-    return result;
+    return std::move(result);
 }
 
 template <class... Ts>
@@ -141,7 +141,7 @@ py::array to_centers(const A &ax) {
         return b.center();
     });
 
-    return result;
+    return std::move(result);
 }
 
 template <class A>
--- a/include/boost/histogram/python/register_axis.hpp
+++ b/include/boost/histogram/python/register_axis.hpp
@@ -31,7 +31,17 @@ using namespace std::literals;
 
 struct options {
     unsigned option;
-    bool none() const { return option == bh::axis::option::none; }
+
+    options(unsigned value)
+        : option(value) {}
+    options(bool uflow, bool oflow, bool circ, bool grow)
+        : option(
+            uflow * bh::axis::option::underflow | oflow * bh::axis::option::overflow
+            | circ * bh::axis::option::circular | grow * bh::axis::option::growth) {}
+
+    bool operator==(const options &other) const { return option == other.option; }
+    bool operator!=(const options &other) const { return option != other.option; }
+
     bool underflow() const {
         return static_cast<bool>(option & bh::axis::option::underflow);
     }
--- a/include/boost/histogram/python/register_axis.hpp
+++ b/include/boost/histogram/python/register_axis.hpp
@@ -29,6 +29,21 @@
 
 using namespace std::literals;
 
+struct options {
+    unsigned option;
+    bool none() const { return option == bh::axis::option::none; }
+    bool underflow() const {
+        return static_cast<bool>(option & bh::axis::option::underflow);
+    }
+    bool overflow() const {
+        return static_cast<bool>(option & bh::axis::option::overflow);
+    }
+    bool circular() const {
+        return static_cast<bool>(option & bh::axis::option::circular);
+    }
+    bool growth() const { return static_cast<bool>(option & bh::axis::option::growth); }
+};
+
 /// Ensure metadata is valid.
 inline void validate_metadata(metadata_t v) {
     try {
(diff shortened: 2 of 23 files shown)


henryiii:
By the way, if we do want to consider these options to be ordered, say in the same order as the bitset, namedtuple is likely the right tool to use.

henryiii:
Sorry, I forgot to include skip CI, this is very much draft.

include/boost/histogram/python/axis.hpp
```diff
@@ -85,7 +85,7 @@ py::array to_values_impl(const A &ax,
     for(auto i = -underflow; i < ax.size() + overflow; i++)
         result.mutable_at(static_cast<std::size_t>(i + underflow)) = ax.value(i);
 
-    return result;
+    return std::move(result);
```
HDembinski: All pybind11 objects can be passed by value, because they derive from pybind11::handle, which is a smart pointer. Moving makes no sense here. Also in general, it may be detrimental to move, because it prevents NRVO.

include/boost/histogram/python/axis.hpp
```diff
@@ -141,7 +141,7 @@ py::array to_centers(const A &ax) {
         return b.center();
     });
 
-    return result;
+    return std::move(result);
```
HDembinski: Same here, no move please.

include/boost/histogram/python/register_axis.hpp
```diff
@@ -145,7 +160,7 @@ void vectorized_index_and_value_methods(
                         "argument must be string or sequence of strings");
                 }
 
-                return indices;
+                return std::move(indices);
```
HDembinski: No move.

src/register_axis.cpp
```diff
@@ -10,19 +10,22 @@
...
-    opt.attr("growth")    = (unsigned)bh::axis::option::growth;
-}
-
 void register_axes(py::module &ax) {
-    // This factory makes a class that can be used to create axes and also be used in
-    // is_instance
-    py::object factory_meta_py
-        = py::module::import("boost_histogram.utils").attr("FactoryMeta");
+    py::class_<options>(ax, "options")
+        .def_property_readonly("none", &options::none)
```
HDembinski: No need to expose `none`, right?

include/boost/histogram/python/axis.hpp
henryiii: The compiler throws warnings here if you don't put move, since it can't copy elide here. I'll look into it more.

HDembinski: Ok, nevermind, moving actually preserves the refcount. So one saves one decrement and one increment.

henryiii: Clang is very good with move warnings, it can also warning you when you are moving and could use copy elision instead.

src/register_axis.cpp
HDembinski: Yes, the options should be equal comparable.

HDembinski: And pickleable (Is that a word?)

henryiii: Agreed. (I have no idea if it is, works for me though ;) )

henryiii:
@HDembinski You have requirements mentioned in the comments for options, such as "Mutually exclusive with `growth` and `underflow`." Are those programmatically available somewhere? Such as `bool bh::detail::options_valid(...)`? I'll like to make things like `options(circular=True, growth=True)` throw an error.

henryiii:
PS: I'm not planning to expose the unsigned value of options to Python unless you think I should for some reason (though it will be how it pickles). I'm also not currently planning to support `&` and `|`, etc, as I think that's not really what these are supposed to be for.  My planned use:

```python
a.options.underflow == True # check a single property
a.options == bh.axis.options(underflow=True, overflow=True) # check all properties at once, useful for testing
```

henryiii:
Let me know if you see anything or would like changes, otherwise we should merge soon to avoid clashes with your axis work. (We can always change after a merge in develop, as well).

henryiii:
Going to merge, we can change as needed.

HDembinski:
In C++, the exclusiveness of circular and underflow is currently not enforced, but it may be a good idea to static_assert on that. I didn't enforce it so far because I consider this not catastrophically harmful. If you set circular with underflow you waste memory, but the code should still work - I think. This is not tested, of course, and before adding tests for that I should rather make it impossible.

In any case, in C++ it will be a type check, so it is not useful for you here. You need to write a separate validation check.

HDembinski:
> PS: I'm not planning to expose the unsigned value of options to Python

Yes, we should not expose it. This is an implementation detail.

Issue #96, opened 2019-09-12
axis interface cleanup


HDembinski:
- [x] remove axis.bins(...), make axis iterable instead
- [x] expose bin objects, or yield tuples/values? I think yielding tuples/values is better.
- [x] nice `__repr__` for each axis 
- [x] remove axis.update(...) for growing axis [it is not to be called by users and there is no need to unit test this in Python as it is already unit tested in C++]

HDembinski:
Ok, then we agree. We don't expose the C++ bin objects, but instead we provide .edges, .centers, .widths.

henryiii:
And a nice repr, I think that will work well.

HDembinski:
Ok, I see why you didn't have the flow argument for `.centers()`. A center would be meaningless for a bin from [X, infinity)

henryiii:
(Yes, that is why)

henryiii:
> remove axis.update(...) for growing axis

This was requested on Gitter by someone who needed to use it.

> nice `__repr__` for each axis

This is two-part, I've just about finished the axis `__repr__`, but we also need some nice way to show the bins. (Maybe the tuples will be enough here, let's see)

HDembinski:
> This was requested on Gitter by someone who needed to use it.

What for? I am strongly against this and unless it is clear to me that this is a legitimate use, I will assume that this person is just confused about how the library works.

henryiii:
Please go to the Gitter channel and read the messages, starting with Mar 19 17:35. You were active in that discussion.

HDembinski:
If you remember the argument, why don't you post it here? I don't need to convince you, I need to be convinced. I know how users should use boost::histogram and by extension boost-histogram.

Having `axis.update` exposed is only causing confusion. Calling update on an axis that is part of a histogram is not allowed by boost::histogram. The update method requires a mutable axis, while the axis returned by a histogram is const. If `axis.update` is exposed in Python, you need to raise an error when a user calls it on an axis of a histogram. Exposing a method that just raises an error is stupid. While it is technically possible to use a non-const axis independently of the histogram, there are no good use cases for that.

henryiii:
@TaiSakuma was the one that was asking about it. I don't need it and am happy to remove it unless it is truly useful and needed (and yes, it should not be allowed on an axis attached to a histogram).

henryiii:
@TaiSakuma feel free to explain why it is or would be useful if you need it!

TaiSakuma:
[...]

PR #121, opened 2019-09-25
axis update: bins, edges, centers

HDembinski:
Further work on #96.


HDembinski:
I removed the two-argument category axis ctor, because it doesn't exist in C++ anymore. The circular axis used to have ctors with two or even one argument, but these were removed when the circular axis became just a modification of a regular axis.

HDembinski:
Since in Python a string is also a sequence of strings with length one, I added that support for the category axis to make it pythonic.

HDembinski:
I decided to have both `__len__` and `.size`, since numpy.array also has both.

examples/getting_started_listing_03.py
```diff
@@ -12,26 +12,22 @@
...
 h.fill(x, y)
 
-# Get representations of the bin edges
-x = h.axis(0).edges()
-y = h.axis(1).edges()
-
-# creates a view of the counts (no copy involved)
-count_matrix = np.asarray(h)
+# Get numpy.histogram compatible representation of the histogram
+w, x, y = h.to_numpy()
```
henryiii: Thank you for working on the examples!

include/boost/histogram/python/register_axis.hpp
```diff
@@ -227,27 +212,27 @@ py::class_<A> register_axis(py::module &m, const char *name, Args &&... args) {
...
             "Return the options associated to the axis")
+
         .def_property(
             "metadata",
             [](const A &self) { return self.metadata(); },
             [](A &self, const metadata_t &label) { self.metadata() = label; },
             "Set the axis label")
 
+        .def_property_readonly(
+            "size",
```
henryiii: Note that np.ndarray.size returns a tuple, and this doesn't (since an axis is always 1D), but I think this is still fine, and matches Boost.Histogram.

tests/test_axis_internal.py
```diff
@@ -26,7 +26,7 @@ def test_axis_regular_uoflow(axtype, function):
     assert 10 == ax.index(1.01)
     assert 10 == ax.index(23)
 
-    assert 10 == ax.size
+    assert 10 == len(ax)
```
henryiii: `ax.size` is also available and 10, correct?

tests/test_axis_internal.py
```diff
@@ -39,60 +39,28 @@ def test_axis_regular_uoflow(axtype, function):
...
     assert 11 == ax.extent
-    assert 11 == len(ax.edges())
-    assert 12 == len(ax.edges(True))
-    assert 10 == len(ax.centers())
+    assert 10 == len(ax.centers)
     assert ax.options == bh.axis.options(overflow=True)
 
     ax = bh.axis.regular(10, 0, 1, flow=False)
     assert 10 == ax.extent
-    assert 11 == len(ax.edges())
```
henryiii: Can't we still check `ax.edges` here?

tests/test_axis_internal.py
HDembinski: Both should be tested, right. Anyway, the tests here are largely overlapping with test_axis_public.py, I will merge the two files of tests next.

tests/test_axis_internal.py
HDembinski: edges was removed.

HDembinski: I would prefer if the only way to get numpy.histogram like edges is via histogram.to_numpy

HDembinski: histogram.to_numpy will get much more thorough testing, naturally, to cover all these cases.

include/boost/histogram/python/register_axis.hpp
HDembinski: I think you confuse `.size` and `.shape`. `np.array((1,2,3)).size == 3`.

HDembinski: `.size` is always a number, even for Nd arrays.

tests/test_axis_internal.py
henryiii: Good! Tests need cleanup.

include/boost/histogram/python/register_axis.hpp
henryiii: Yes, I am confused. Ignore me, please. :)

tests/test_axis_internal.py
henryiii: Numpy like edges could be useful for plotting and other uses that are not just trying to get a numpy like return out. Numpy itself added a way to get the edges only a couple of versions ago. This seems useful to have. https://docs.scipy.org/doc/numpy/reference/generated/numpy.histogram_bin_edges.html#numpy.histogram_bin_edges 

(This is fine for now, just something to think about)

HDembinski: I am ok with adding edges as a property without `flow=True`. `histogram.to_numpy` should be only way to get edges with `flow=True`, because `to_numpy` guarantees that the shape of the counter matrix matches the edges shape.

henryiii: I think that is okay, especially for axis. Flow is a bit of a special case for edges.

HDembinski:
I believe that axes should have a uniform interface as much as possible in Python. Anticipating how axis.centers and axis.edges is going to be used, I now treat an integer axis like regular(size(), min, max) and a category axis like regular(size(), 0, size()). This convention allows one to return a valid edge array for all axes. For plotting histograms, these are the numbers that are needed. The category axis is an oddball, it can't be helped. Users can get the label for each category via axis[idx] or axis.bin(idx).

HDembinski:
I merge this now, the behavior of axis.edges for integer and category axes is the only thing we didn't discuss before, and this can still be changed in another PR if necessary.

henryiii:
Looks good. I think that is a reasonable interpretation of edges. They remain useful for plotting (which is probably the main use for edges), and have a consistent return type.

PR #123, opened 2019-09-28
Merge axis tests

HDembinski:
I removed the `flow=bool` keyword from all axis inits, since we explicitly use `underflow` and `overflow` now. 

I discovered a segfault in metadata_t comparison which I fixed.

The default value for metadata is now `None` instead of an empty string.

Setting metadata=1 is now allowed in axis inits.

I renamed the axis types so that the C++ types do not start with an underscore. It is not forbidden to start identifiers with underscore and a **small letter inside a namespace**, but the exact rule is subtle and most programmers (including me) prefer to ban identifiers starting with underscore completely just to avoid any possible issue. See
https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier


henryiii:
It is very common use case to turn off flow; it's very verbose to have to write `overflow=False, underfow=False` to do so. `flow=False` was a shortcut to make this simpler and less verbose. We've managed to reduce the need to turn off flow by careful design of the access methods, and I don't think there's much of an issue with performance (need to test filling with a large fraction of events that go in flow bins), so not quite as necessary as it used to be, but it still was a nice shortcut.

include/boost/histogram/python/axis.hpp
```diff
@@ -26,15 +26,12 @@ struct type_caster<bh::axis::variant<Ts...>>
...
     PYBIND11_OBJECT_DEFAULT(metadata_t, object, PyObject_Check);
 
-    bool operator==(const metadata_t &other) const {
-        return py::cast<bool>(this->attr("__eq__")(other));
-    }
-
+    bool operator==(const metadata_t &other) const { return py::object::equal(other); }
     bool operator!=(const metadata_t &other) const {
-        return py::cast<bool>(this->attr("__ne__")(other));
+        return py::object::not_equal(other);
```
henryiii: Nice; noted for future use. :)

include/boost/histogram/python/axis.hpp
```diff
@@ -260,71 +259,67 @@ py::array_t<double> widths(const A &ax) {
...
-using _regular_uoflow = bh::axis::regular<double, bh::use_default, metadata_t>;
-using _regular_uflow  = bh::axis::
-    regular<double, bh::use_default, metadata_t, bh::axis::option::underflow_t>;
-using _regular_oflow = bh::axis::
-    regular<double, bh::use_default, metadata_t, bh::axis::option::overflow_t>;
-using _regular_noflow
-    = bh::axis::regular<double, bh::use_default, metadata_t, bh::axis::option::none_t>;
-using _regular_growth = bh::axis::
-    regular<double, bh::use_default, metadata_t, bh::axis::option::growth_t>;
+// These match the Python names except for a possible underscore
```
henryiii: I was debating adding the underscore here.

By the way, the new Python layout means these are only available from `boost_histogram.core.axis`, and not from the normal `boost_histogram.axis`, so dropping the underscore entirely might be an option. Up to you.

include/boost/histogram/python/axis_ostream.hpp
```diff
@@ -34,18 +34,10 @@ inline const char *axis_suffix(const axis::transform::log &) { return "_log"; }
...
-            if(!oss.str().empty()) {
-                os << ", metadata=" << std::quoted(oss.str());
-            }
-        },
-        [&os](const auto &) { os << ", metadata=" << detail::type_name<T>(); },
-        t);
+template <class OStream>
+void stream_metadata(OStream &os, const metadata_t &t) {
+    if(!t.is_none())
+        os << ", metadata=" << std::quoted(py::cast<std::string>(t));
```
henryiii: Do all python types cast to strings (I assume through repr or str)?

include/boost/histogram/python/register_axis.hpp
```diff
@@ -53,29 +53,6 @@ struct options {
     bool growth() const { return static_cast<bool>(option & bh::axis::option::growth); }
 };
 
-/// Ensure metadata is valid.
-inline void validate_metadata(metadata_t v) {
```
henryiii: We need to add the kargs and get metadata from that, can be in a later PR, though. I can do that.

tests/test_pickle.py
```diff
@@ -56,19 +56,19 @@ def test_accumulators(accum, args, copy_fn):
...
-    (bh.axis.regular_sqrt, (4, 2, 4)),
-    (bh.axis.regular_pow, (4, 2, 4, 0.5)),
-    (bh.axis.circular, (4, 2, 4)),
-    (bh.axis.variable, ([1, 2, 3, 4],)),
-    (bh.core.axis._integer_uoflow, (1, 4)),
-    (bh.core.axis._category_int, ([1, 2, 3],)),
-    (bh.core.axis._category_int_growth, ([1, 2, 3],)),
-    (bh.core.axis._category_str, (["1", "2", "3"],)),
-    (bh.core.axis._category_str_growth, (["1", "2", "3"],)),
+    (bh.core.axis._regular_none, (4, 2, 4, None)),
```
henryiii: We'll be dropping the ability to pass metadata in as a positional argument, so we'll need to handle the final item in this tuple specially or remove it.

HDembinski:
> It is very common use case to turn off flow

No, it should not be turned off by users and making it cumbersome is good. If you don't like it, you can change it in hist.

tests/test_pickle.py
HDembinski: The _axis are internal, they get all arguments as positional. There is no need to make their interface complicated, because the internal axis types are only called by the library.

include/boost/histogram/python/register_axis.hpp
HDembinski: It looks ok to me as it is right now. The test all work. Since the number of positional arguments is fixed for each axis now, there is no ambiguity anymore. I think we should leave it as is.

tests/test_pickle.py
HDembinski: I want to remove the tests of the private axis types. We need to test the user-facing axis anyway and then these test are redundant.

include/boost/histogram/python/axis_ostream.hpp
HDembinski: Casting to string calls `str(obj)`, this is closest to what you hand before. We may want to call `repr(obj)`.

HDembinski: The `repr` streaming needs to be improved anyway. For `repr`, we want to show all axis options and the metadata even if it is none, but we don't show these things when they are the default values. That's ok for `str` but `repr` should be more specific.

HDembinski: This is another topic for another PR

include/boost/histogram/python/axis.hpp
HDembinski: Does it make sense to have a `_core` module and then everything inside can be without underscore?

henryiii: We can. I wasn't sure whether to name it core or _core. I don't currently have a strong reason to go one way vs. the other. I *think* I like _core + no underscores in names inside it a bit better.

henryiii: (also a topic for another PR)

include/boost/histogram/python/register_axis.hpp
henryiii: I'll make another PR or issue, I really don't like allowing someone to pass metadata without a keyword. It's not clear what they mean. Not important for this PR.

tests/test_pickle.py
henryiii: Okay; I don't think it's bad to test private components in unit tests, but only if they only available as an integration test using public interfaces, which is not true here. The main reason these exist was the order in which this was developed.

henryiii:
Is this still true in the case where you are histogramming a small segment of a much larger distribution? It's also nice in the C++ case, where you have overflow bins that end up overflowing because you are using 8 or 16 bits.

But in Python, and if it is always faster, I don't think there's such a good reason to turn them off, *as long as we get the interface right* (which I think we are doing so far). Most users, except those coming from ROOT, will probably not expect flow bins to be there, but will still silently benefit from things like projections.

tests/test_pickle.py
HDembinski: Right, that makes sense.

include/boost/histogram/python/axis.hpp
HDembinski: Right.

HDembinski:
> But in Python, and if it is always faster, I don't think there's such a good reason to turn them off, as long as we get the interface right (which I think we are doing so far). Most users, except those coming from ROOT, will probably not expect flow bins to be there, but will still silently benefit from them for things like projections.

Right, my thoughts exactly.

HDembinski:
> Is this still true in the case where you are histogramming a small segment of a much larger distribution?

Yes, at least how it is implemented right now. There may actually be a faster way that I think about it.

PR #124, opened 2019-10-06
Move make functions to Python, keyword only

henryiii:
All arguments are now keyword only, except the basic ones unique to an axis type. The creation functions are now written in Python. The axis init c++ function is now able to add init arguments like arguments names. All creation functions support keyword arguments now, including private ones.

Not that direct axis creation, like `regular_*`, still allow metadata to be passed by position for now. Help and signature are not ideal yet.

### DIFF
--- a/boost_histogram/axis.py
+++ b/boost_histogram/axis.py
@@ -2,21 +2,77 @@ from .core.axis import regular_log, regular_sqrt, regular_pow, circular, options
 
 from .core import axis as ca
 
-from .utils import FactoryMeta
+from .utils import FactoryMeta, KWArgs
+
+
+# When Python 2 is dropped, this could use keyword
+# only argument syntax instead of kwargs
+def _make_regular(bins, start, stop, **kwargs):
+    """
+    Make a regular axis with nice keyword arguments for underflow,
+    overflow, and growth.
+    """
+    with KWArgs(kwargs) as k:
+        metadata = k.optional("metadata")
+        options = k.options(underflow=True, overflow=True, growth=False, circular=False)
+
+    if options == {"growth", "underflow", "overflow"}:
+        return ca._regular_uoflow_growth(bins, start, stop, metadata)
+    elif options == {"underflow", "overflow"}:
+        return ca._regular_uoflow(bins, start, stop, metadata)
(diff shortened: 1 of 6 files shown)


HDembinski:
This looks very good.

There is only one thing that I find superfluous, having default parameters for the inits of internal classes, e.g. `_regular_X`. I think internal classes should accept only positional arguments, because any kind of convenience like keyword-based arguments with defaults is not necessary, since these classes are never called directly by a user.

henryiii:
Currently, this is needed for the tests, since they call both private and public inits in a parameterized list. There's currently no penalty for having keyword names on the parameters (and there's no need to force them to be keyword only). It also is more descriptive, etc.

If PyBind11 gains support for FastCall, there might be a small performance benefit for not having the keywords present in the new Pythons, but we can always drop it since it's a private constructor (and it looks like there have been some patches recently that allow keyword arguments in FastCall, and you still get the performance boost as long as you don't use them).

I don't expect axis creation in Python to really matter for performance, though.

henryiii:
Merging; we can edit as needed later.

henryiii:
> you reverted this

No, I did not. Your tests were calling both public and private interfaces as well; you were just calling the public interfaces with a positional metadata argument, which is no longer supported. Eventually we should remove the private interface calls here, then it will no longer matter if these have names or not.

> weird to have convenience features for a backend

Names for arguments is very normal in Python. In fact, until Python 3.8 comes out, there is no way in pure Python to make a limited number positional only arguments.

 > A backend does not need to be user-friendly.

But it does need to be developer friendly. Having names on the arguments may help a future developer at some point. *Unless there's a performance penalty*, it is at least microscopically better to have names. And doc strings.

HDembinski:
> Names for arguments is very normal in Python. In fact, until Python 3.8 comes out, there is no way in pure Python to make a limited number positional only arguments.

There is no need to make a backend user-friendly, because it has no users.

henryiii:
Once the tests are fixed, these can be dropped, though I think it's better to always be complete and always provide names on principle. Some day some developer might be digging around in Python, trying to figure out (or remember?) how everything works and will be helped out by the names, so don't see a reason to remove them if they are already there.

PR #125, opened 2019-10-08
Slower, closer to C++ accumulators

### LINKED ISSUE #115: Fix the accumulator interface
The `__init__` and `__call__` interface should be removed.

See arguments for storage interface.

henryiii:
Accumulators are a bit closer to C++ now. They have the same syntax; for now, no vectorized method to create new accumulators or fill accumulators has been added.


henryiii:
Should weight be a keyword only argument, actually? The mean accumulators might be confusing, since they take weight first.

henryiii:
Also, note that vectorize creates an output array of bools that we throw away. The correct solution would be to add a version of vectorize to PyBind11 that does not produce an output (todo if someone wants to tackle it). For now, it's only 20% slower and handles broadcasting, scalars, and such correctly, so should be fine for this use.

HDembinski:
> Should weight be a keyword only argument, actually?

Yes, I think so.

HDembinski:
The init should not accept arrays of values. It is weird and unpythonic to have an overloaded init, it is also a deviation from C++. There should be one way to fill values and that is via the call operator.

henryiii:
The call operator is also a deviation from C++. Sum and weighted sum don’t even support it, and the mean accumulators don’t support arrays in C++. Using the call operator to modify the state is *very* unpythonic.  We could either use keywords of a classfunction to init, but I think the most common use is likely to get an accurate sum or mean from a numpy array.

henryiii:
Overload behavior is seen in Python. Look at `range()`, for example.

```
In [1]: range(10)
Out[1]: range(0, 10)

In [2]: range(10, 0)
Out[2]: range(10, 0)

In [3]: range?
Init signature: range(self, /, *args, **kwargs)
Docstring:
range(stop) -> range object
range(start, stop[, step]) -> range object
```

That looks like an overload to me. Also arguably the most important class/function in Python, `type`:

```
In [4]: type?
Init signature: type(self, /, *args, **kwargs)
Docstring:
type(object_or_name, bases, dict)
type(object) -> the object's type
type(name, bases, dict) -> a new type
```

Not saying we need to keep the overload; one or both of the inits could be moved to a new classfunction or something similar, but they do exist in Python.

henryiii:
Actually, look at the current constructors:

```python
In [1]: import boost_histogram.accumulators as bha

# Sum

In [2]: bha.sum()
Out[2]: sum(0 + 0)

In [3]: bha.sum(1)
Out[3]: sum(1 + 0)

In [4]: bha.sum([1])
Out[4]: sum(1 + 0)

# note that you don't have a way to
# init "small" even in C++ for sum

# Weighted Sum

In [5]: bha.weighted_sum()
Out[5]: weighted_sum(0, 0)

In [6]: bha.weighted_sum(1)
Out[6]: weighted_sum(1, 1)

In [7]: bha.weighted_sum([1])
Out[7]: weighted_sum(1, 1)

In [8]: bha.weighted_sum(1, 2)
Out[8]: weighted_sum(1, 2)

# Mean

In [9]: bha.mean()
Out[9]: mean(0, 0, -0)

In [10]: bha.mean(1)
Out[10]: mean(1, 1, nan)

In [11]: bha.mean([1])
Out[11]: mean(1, 1, nan)

In [12]: bha.mean(1, 2, 3) # Odd?
Out[12]: mean(1, 2, nan)

# Weighted Mean

In [13]: bha.weighted_mean()
Out[13]: weighted_mean(0, 0, nan)

In [14]: bha.weighted_mean(1)
Out[14]: weighted_mean(1, 1, nan)

In [15]: bha.weighted_mean(1, 2)
Out[15]: weighted_mean(1, 2, nan)

In [16]: bha.weighted_mean([1], [2])
Out[16]: weighted_mean(1, 2, nan)

In [17]: bha.weighted_mean(1, 2, 3, 4)
Out[17]: weighted_mean(1, 3, 4)

```

To me, that looks simple, a natural extension of Python, and doesn't really appear overloaded to the user. I don't see a problem here. The only current downside is that you can't do a vectorized operation on an initialized accumulator, but that could be added later if needed.

henryiii:
Removing vector init. Since I'll probably squash the commits when merging, I'll paste the old code here:

```c++
template <class T>
decltype(auto) init_1_vectorized_add() {
    return py::init([](py::object input_1) {
        auto ptr = new T();

        // There is an alternate way to do this using py::array_t<double>
        // and would skip allocating a useless bool array. However, it is
        // only 20% faster and does not handle broadcasting and scalars.
        // Let's piggyback on py::vectorize for now to get those nice features.
        //
        // // This is just as fast as using raw pointers, and more elegant
        // auto r = input_1.unchecked<1>();
        // for(py::ssize_t idx = 0; idx < r.shape(0); ++idx)
        //     (*ptr) += r(idx);

        py::vectorize([](T *p, const double &in_1) {
            (*p) += in_1;
            return false; // Needed for py::vectorize output
        })(ptr, input_1);
        return ptr;
    });
}

template <class T>
decltype(auto) init_1_vectorized_call() {
    return py::init([](py::object input_1) {
        auto ptr = new T();

        py::vectorize([](T *p, const double &in_1) {
            (*p)(in_1);
            return false; // Needed for py::vectorize output
        })(ptr, input_1);
        return ptr;
    });
}

template <class T>
decltype(auto) init_2_vectorized_add() {
    return py::init([](py::object input_1, py::object input_2) {
        auto ptr = new T();

        py::vectorize([](T *p, const double &in_1, const double &in_2) {
            (*p) += T(in_1, in_2);
            return false; // Needed for py::vectorize output
        })(ptr, input_1, input_2);
        return ptr;
    });
}

template <class T>
decltype(auto) init_2_vectorized_call() {
    return py::init([](py::object input_1, py::object input_2) {
        auto ptr = new T();

        py::vectorize([](T *p, const double &in_1, const double &in_2) {
            (*p)(in_1, in_2);
            return false; // Needed for py::vectorize output
        })(ptr, input_1, input_2);
        return ptr;
    });
}

```

henryiii:
Without vector constructors, we loose the ability to use weighted sum in a vector form. I think the vector constructors were closer to the spirit of the C++ code, but they have been removed for now. I've added the vectorized call back to mean family, and vectorized sum (which does not have the ability to do weights).

henryiii:
We could define vectorized constructors in a simpler and non-overloaded way: For any constructor, passing a vector simply constructs N items and sums them. This would even work for 4-element constructors. Would that be acceptable?

So:

```python
weighted_sum([1,2], 3)
```

would be:

```python
w = weighted_sum(1, 3)
w += weighted_sum(2, 3)
```

(we can wait to work on this till after the beta)

HDembinski:
In all the examples you showed, the init did something similar, you just didn't have to specify all the values when some could be inferred. It is bad design to use init to do two very different things: setting up the accumulator with some initial state *or* filling values. It is a mix of responsibilities.

The generic accumulator interface in boost::histogram is `operator()`, because it allows for any number of arguments. To be backward compatible with simple ints and doubles, also `operator+=` is supported. I decided to use `operator+=` for `sum` and `weighted_sum`, but I could have used `operator()` as well. It does not matter which interface is used in C++.

I don't mind if we have operator() also for `sum` and `weighted_sum`, then it is consistent on the Python side.

HDembinski:
I told you before that the accumulators are components of boost::histogram, they were not designed to be used standalone.

henryiii:
I see two possible uses for accumulators beyond teaching purposes:

1. Providing a high-accuracy sum/mean of an array.
2. Providing a running sum/mean that gets updated inside a loop without storing values.

The first would very much benefit from a simple way to apply an accumulator to an array without having to write multiple lines and deal with a stateful interface. The second doesn't really care about vectorization. Accumulators are rather like a single element histogram and seem like a natural thing to find in a histogram package.

I think using the constructor here is rather consistent. If you give a single value, that initializes the state. If you give multiple values, it computes an initial state from the accumulation of the multiple values. It doesn't stop you from further additions to that state later. We could also have a specialized class method here if it really is bad to combine these, but that adds another thing for users to learn. `bh.accumulators.mean(my_array).value` will be fairly familiar for users, whereas `mean = bh.accumulators.mean(); mean(my_array); mean.value` is like nothing else in Python. For the modifying nature, `for x in my_array: mean(x)` is unavoidable but better. `for x in my_array: sum += x` is good since it uses explicit Python modification syntax. And, just like C++, a user could do `for x in my_array: mean += bh.accumulators.mean(x)` to avoid the bad use of the call operator.

I may drop the vectorized call/sum for now until we can decide on a good way to do this. We can add it later if needed.

henryiii:
(This also means we can wait until after https://github.com/pybind/pybind11/pull/1951 )

henryiii:
Another thought: What about a fill method, matching histograms? It would accept vectorized arguments, so wouldn't need to match C++. `bh.accumulators.mean().fill(my_array).value` isn't awful.

henryiii:
This should be completely non-controversial, so will merge. We can continue to discuss some way to augment the C++ binding with a vectorized fill or constructor later.

HDembinski:
Precisely because there is a symmetry between accumulators and histograms, making and using an accumulator must be two steps. boost-histogram uses a stateful histogram object to do the job, not a function like `np.histogram`. You make the histogram and then you fill it with values. Likewise, an accumulator -- for consistency -- is created first and then filled with values.

PR #126, opened 2019-10-08
Updating histogram, adding some tests

henryiii:
Currently failing to pickle/unpickle histogram.


henryiii:
@HDembinski Any thoughts? I'll look at this more carefully tomorrow.

```
_____________________________________________________________________________ test_pickle_0 ______________________________________________________________________________

    def test_pickle_0():
        a = histogram(
            category([0, 1, 2]),
            integer(0, 20),
            regular(2, 0.0, 20.0, underflow=False, overflow=False),
            variable([0.0, 1.0, 2.0]),
            circular(4, 0, 2 * np.pi),
        )
        for i in range(a.axis(0).extent):
            a.fill(i, 0, 0, 0, 0)
            for j in range(a.axis(1).extent):
                a.fill(i, j, 0, 0, 0)
                for k in range(a.axis(2).extent):
                    a.fill(i, j, k, 0, 0)
                    for l in range(a.axis(3).extent):
                        a.fill(i, j, k, l, 0)
                        for m in range(a.axis(4).extent):
                            a.fill(i, j, k, l, m * 0.5 * np.pi)

        io = pickle.dumps(a, -1)
        b = pickle.loads(io)

        assert id(a) != id(b)
        assert a.storage.__class__ == b.storage.__class__
        assert a.storage == b.storage
        assert a.rank == b.rank
        assert a.axis(0) == b.axis(0)
        assert a.axis(1) == b.axis(1)
        assert a.axis(2) == b.axis(2)
        assert a.axis(3) == b.axis(3)
        assert a.axis(4) == b.axis(4)
        assert a.sum() == b.sum()
        assert repr(a) == repr(b)
        assert str(a) == str(b)
>       assert a == b
E       assert histogram(\n  ...orage=double\n) == histogram(\n  ...orage=double\n)
E         Use -v to get the full diff

../tests/test_public_hist.py:425: AssertionError
```

tests/test_public_hist.py
```diff
@@ -690,14 +695,20 @@ def ar(*args):
     with pytest.raises(KeyError):
         a.fill((1, 2), foo=(1, 1))
     with pytest.raises(ValueError):
-        a.fill((1, 2), weight=(1,))
```
HDembinski: I am not sure why this changed. I think it should still fail.

HDembinski: `a.fill((1, 2), weight=1)` is allowed, but `a.fill((1, 2), weight=(1,))` should not be allowed

henryiii: The good thing is:

```python
a.fill((1,2,3), weight=(1,2))
```

still throws, so this is just "broadcasting", but not sure why it started doing so.

henryiii:
Here's the output on my laptop. 0002 is the new version. `[storage-flow]`. `totaltime (scaled time, with fastest=1)`. Unlimited seems to be slower now?... I'll try on a different machine tomorrow.

```
-------------- benchmark '1d-fills': 14 tests --------------
Name (time in ms)                           Median
------------------------------------------------------------
test_1d[double-False] (0001_e0fb9a3)        1.1018 (1.32)
test_1d[double-False] (0002_b7b4faa)        1.0874 (1.30)
test_1d[double-True] (0001_e0fb9a3)         0.8602 (1.03)
test_1d[double-True] (0002_b7b4faa)         0.8361 (1.0)
test_1d[int-False] (0001_e0fb9a3)           1.1063 (1.32)
test_1d[int-False] (0002_b7b4faa)           1.1049 (1.32)
test_1d[int-True] (0001_e0fb9a3)            0.8465 (1.01)
test_1d[int-True] (0002_b7b4faa)            0.8705 (1.04)
test_1d[unlimited-False] (0001_e0fb9a3)     1.2602 (1.51)
test_1d[unlimited-False] (0002_b7b4faa)     1.3726 (1.64)
test_1d[unlimited-True] (0001_e0fb9a3)      0.9814 (1.17)
test_1d[unlimited-True] (0002_b7b4faa)      1.1779 (1.41)
test_numpy_perf_1d (0001_e0fb9a3)           1.5031 (1.80)
test_numpy_perf_1d (0002_b7b4faa)           1.5256 (1.82)
------------------------------------------------------------

--------------- benchmark '2d-fills': 14 tests --------------
Name (time in ms)                            Median
-------------------------------------------------------------
test_2d[double-False] (0001_e0fb9a3)         2.0530 (1.30)
test_2d[double-False] (0002_b7b4faa)         2.0366 (1.29)
test_2d[double-True] (0001_e0fb9a3)          1.6298 (1.03)
test_2d[double-True] (0002_b7b4faa)          1.5762 (1.00)
test_2d[int-False] (0001_e0fb9a3)            2.0549 (1.30)
test_2d[int-False] (0002_b7b4faa)            2.0232 (1.28)
test_2d[int-True] (0001_e0fb9a3)             1.6347 (1.04)
test_2d[int-True] (0002_b7b4faa)             1.5757 (1.0)
test_2d[unlimited-False] (0001_e0fb9a3)      2.1619 (1.37)
test_2d[unlimited-False] (0002_b7b4faa)      2.2714 (1.44)
test_2d[unlimited-True] (0001_e0fb9a3)       1.7056 (1.08)
test_2d[unlimited-True] (0002_b7b4faa)       1.8758 (1.19)
test_numpy_perf_2d (0001_e0fb9a3)           10.5067 (6.67)
test_numpy_perf_2d (0002_b7b4faa)           10.3665 (6.58)
-------------------------------------------------------------
```

[Speed in Milliseconds (ms) of 1d-fills.pdf](https://github.com/scikit-hep/boost-histogram/files/3710052/Speed.in.Milliseconds.ms.of.1d-fills.pdf)

[Speed in Milliseconds (ms) of 2d-fills.pdf](https://github.com/scikit-hep/boost-histogram/files/3710053/Speed.in.Milliseconds.ms.of.2d-fills.pdf)

HDembinski:
Those changes look they are mostly in the noise, the improvement is definitely smaller than the 10% I am seeing C++. If you run the benchmark several times, do you always get the same results?

HDembinski:
What is also weird: in C++, the `unlimited_storage` is way slower than `dense_storage<double>`, while the difference is not that great in your benchmarks. This would suggest that the bottleneck is somewhere else.

henryiii:
For CMake, this defaults to Release, which adds `-O3 -DNDEBUG` for you. For setup.py, this uses the built-in Python settings, which generally also includes `-O3 -DNDEBUG` (it does for me). Neither of them set an arch or use the unsafe math optimizations, so we could do the latter and mention how to do the former.

I'm really suspecting we have a copy somewhere, and that's taking a good portion of our time. That would explain both the similarity in times we see for unlimited as well as the lack of improvement.

henryiii:
We'll need to profile. I played with the values when setting up the fill and it does not seem to be making a copy.

src/register_version.cpp
```diff
@@ -1,12 +0,0 @@
...
-//
-// Distributed under the 3-Clause BSD License.  See accompanying
-// file LICENSE or https://github.com/scikit-hep/boost-histogram for details.
-
-#include <boost/histogram/python/pybind11.hpp>
-
-#include <boost/histogram/histogram.hpp>
-
-void register_version(py::module &m) {
-    m.attr("BOOST_HISTOGRAM_DETAIL_AXES_LIMIT") = BOOST_HISTOGRAM_DETAIL_AXES_LIMIT;
```
henryiii: What about restoring this as `"_AXES_LIMIT"`? Then using it in the histogram creation function to check the maximum number of axes to fix #128?

HDembinski: That would be ok. I am also ok with calling it `_axes_limit`.

HDembinski: Why not making the check in C++, where you have access to `*AXES_LIMIT`?

henryiii: Because in #130 I move histogram and make_histogram to Python. The only thing left would be to add it to each histogram constructor, replacing `py::init<...>()` with custom code that returns a pointer. Not awful, but not necessary quite as nice.

henryiii: Let me know which way you want to go. I will add `_axes_limit` in this PR if you want the check to happen in Python. If you fix this is Boost.Histogram, we can just rely on that exception for now.

henryiii: I'll go this way for now, and change later if you want me to.

henryiii:
Fixes #128.

PR #127, opened 2019-10-10
Faster math, better support for IDEs

henryiii:
Now checking with several different input types.


HDembinski:
You don't have to test `storage.int`. Integer math is as fast as double math on modern CPUs.

henryiii:
The test was for the filling with and without copies - float64 and int64 does not make a copy, while float32 and int32 do.

henryiii:
(We probably have a slight overkill in variations profiled - will probably see if I can trim it down a bit)

PR #130, opened 2019-10-11
Move more to Python

henryiii:
This moves quite a bit more to Python: `make_histogram` and `__getitem__`. Factory functions now get docstrings from the original function wrapped. A few names simplified.


henryiii:
TODO: rename hist to _hist or _hist<something>, since it's not meant for users.

boost_histogram/_hist.py
```diff
@@ -0,0 +1,133 @@
...
+    _core.storage.int: _core.hist._any_int,
+    _core.storage.atomic_int: _core.hist._any_atomic_int,
+    _core.storage.unlimited: _core.hist._any_unlimited,
+    _core.storage.weight: _core.hist._any_weight,
+    _core.storage.mean: _core.hist._any_mean,
+    _core.storage.weighted_mean: _core.hist._any_weighted_mean,
+}
+
+
+def _arg_shortcut(item):
```
henryiii: TODO: This probably should be removed before the beta.

boost_histogram/_hist.py
```diff
@@ -0,0 +1,133 @@
...
+
+def _make_histogram(*args, **kwargs):
+    """
+    Make a histogram with an optional storage (keyword only).
+    """
+
+    with KWArgs(kwargs) as k:
+        storage = k.optional("storage", _core.storage.double())
+
+    # Initialize storage if user has not
```
henryiii: @HDembinski This shortcut is now more reliable, since it will only turn classes into default constructed objects. I can still drop it if you don't like it.

henryiii: (The shortcut is one line below the shown snippet, I picked a poor line to comment)

PR #131, opened 2019-10-12
Adding accumulator fill and kw arguments

### LINKED ISSUE #129: Vectorize accumulators
This is a placeholder for a discussion on the best way to provide a vectorized accumulator interface in Python. This is not directly mappable to C++ (since in C++ a loop is not 100x slower than an array, and C++14 does not really have anything like Numpy).

Possible options so far:

1. Something in the constructor

Possibly nicest Python interface: `bh.accumulor.sum(arr)` creates a pre-filled accumulator from an array.

2.  Specific vectorized function

Probably `bh.accumulor.sum().fill(arr)`, nice symmetry with Histograms.

3. Vectorized call operator

Make very ugly Python: `bh.accumulor.sum()(arr)`, does not play well with duck typing, and modifying a object with a call operator is very ugly (not: mean already has a 1-item call operator). I don't like the idea of adding call operators to sum, where C++ doesn't even have them.

4. Vectorized operators

Like the one above, but uses += arr for sum. It is not possible to add weights for weighted_sum, however.

henryiii:
Closes #129. Provides a `.fill` method and makes weight/variance keyword only.

### DIFF
--- a/src/register_accumulators.cpp
+++ b/src/register_accumulators.cpp
@@ -10,8 +10,46 @@
 #include <boost/histogram/accumulators/sum.hpp>
 #include <boost/histogram/accumulators/weighted_mean.hpp>
 #include <boost/histogram/accumulators/weighted_sum.hpp>
+#include <boost/histogram/python/kwargs.hpp>
 #include <boost/histogram/python/register_accumulator.hpp>
 #include <pybind11/operators.h>
+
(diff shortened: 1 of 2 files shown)


henryiii:
Note: If https://github.com/pybind/pybind11/pull/1196 gets updated and merged, this can be changed to use that instead. I will try to look into this post PyHEP.

PR #132, opened 2019-10-12
Storage class

henryiii:
Adding this simplifies the Python portion of the bindings. @HDembinski, I've made this hidden, but I'm open for name suggestions. I'm not sure if `_storage_class` or `_storage_type` is better - they are somewhat interchangeable in Python.

### DIFF
--- a/boost_histogram/_hist.py
+++ b/boost_histogram/_hist.py
@@ -2,15 +2,15 @@ from .utils import FactoryMeta, KWArgs
 
 from . import core as _core
 
-_histogram_and_storage = {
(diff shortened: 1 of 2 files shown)


henryiii:
Ah, it's called `storage_type` in Boost.Histogram's histogram class. So should it be `_storage_type` or `storage_type`? It is public in C++.

henryiii:
This should be non-controversial in it's hidden form and it's conflicted with the last two PRs. Let me know if you'd like me to name it `bh.histogram.storage_type` instead of `bh.histogram._storage_type`.

PR #141, opened 2019-10-14
Fix projections and fix test for projection, add sum behavior

### LINKED ISSUE #139: Projection issue with UHI
UHI projection is reversed (my fault).

- [x] fix projection axes
- [x] allow axes to be removed by projection
- [ ] copy UHI description to separate repository (local version can mention current status in boost-histogram)

henryiii:
This corrects the projection UHI implementation, and fixes the test so that it actually reports an error if it doesn't work (oops).

Fixes #139.

### DIFF
--- a/boost_histogram/_hist.py
+++ b/boost_histogram/_hist.py
@@ -110,7 +110,7 @@ def _compute_getitem(self, index):
     except RuntimeError:
         pass
 
-    projections = []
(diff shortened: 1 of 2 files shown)


henryiii:
It's bad when you have a bug and your test covering the behavior also has a bug...

PR #150, opened 2019-10-17
Dropping indexed, filling out axes

### LINKED ISSUE #148: Augment `.axis(i)` with `.axes[i]`
It is very odd to have a function in Python that returns each element of something that is inherently a list - it would be more Pythonic to have a property that returns a list. So Python things like iteration and unpacking would work.

```python
ax0, ax1 = h.axes
len(h.axes)
ax0 == h.axes[0]
```

Etc.

henryiii:
This removes `.indexed/bh.indexed`, and replaces it with `.axes.*`. Closes #148 and #54.


henryiii:
@HDembinski What do you think? Also, having both `axes` and `axis`, one with `()` and one with `[]` could be slightly confusing to users. I think I'm okay with it, but it does provide two ways to get the same thing but with different bracket syntax.

henryiii:
And, if you do think it's okay to drop `.axis(i)`, we'll need to put a deprecation warning and drop it in 0.6.0. Same for anything else we show then change/remove if it is a significant. `indexed` can be killed in 0.5.1 since it was not shown.

(by "drop", I mean rename `_axis`, since it is internally used by axes. Also, `axes` might be trivially cached on the python side, while `axis` would not be)

HDembinski:
I don't want not remove `.axis(i)` because that's the interface in C++. Having `.axes` has advantages, so I am ok with having that duplicated.

HDembinski:
`.at` and `__getitem___` are also duplicated, and there are other things duplicated. The duplicated interface have always been proposed by you. I will not agree to remove any C++ interface because you introduce some other interface.

HDembinski:
That being said, there are good reasons to have `__getitem__` and `.axes` in the way they are. I can live with the duplication.

henryiii:
I'm fine to live with it, too, just wanted to make sure we considered it properly. Otherwise, does the PR look fine?

I don't think it's a good idea, but I could always add `__call__` to `axes` then rename it `axis`. Then it would be all in one.

PR #154, opened 2019-10-17
Source dist problem

### LINKED ISSUE #153: Variant2 not included in the source tarball
Add sourcetarball test to tests. If not tested, it is broken...

henryiii:
First adding a test that fails.

Will fix #153.


henryiii:
Thanks @chrisburr!

PR #160, opened 2019-10-18
Version 1.5.1

henryiii:
Fixes a few key bugs, like fill on non-1 strides and removes indexed. `axes` is provisionally available as a replacement for `axis`, nothing has been deprecated yet, though. `bh.project` has not changed names yet.


HDembinski:
👍

PR #163, opened 2019-10-22
Even simpler listings of source files

henryiii:
Followup to #162. Verify this does work in all cases.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -129,7 +129,7 @@ class BuildExt(build_ext):
 
 extras = {
     "test": ["pytest", "pytest-benchmark", "numpy", 'futures; python_version < "3"'],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -134,7 +129,7 @@ def build_extensions(self):
 
 extras = {
     "test": ["pytest", "pytest-benchmark", "numpy", 'futures; python_version < "3"'],
-    "docs": ["Sphinx>=2.0.0", "recommonmark>=0.5.0", "sphinx_rtd_theme"],
+    "docs": ["Sphinx>=2.0.0", "recommonmark>=0.5.0", "shinx_rtd_theme"],
```
henryiii:
```suggestion
    "docs": ["Sphinx>=2.0.0", "recommonmark>=0.5.0", "sphinx_rtd_theme"],
```

Issue #156, opened 2019-10-18
UHI reprs


henryiii:
We should have nice reprs for UHI objects as well.

PR #164, opened 2019-10-22
Adding offset to UHI

### LINKED ISSUE #152: support for `bh.loc(value) + N`
This would just offset the index by N

henryiii:
This also moves the files around just a bit - UHI components are now exported from `uhi.py` instead of `utils.py`, and `_utils.py` is now private. Closes #152


henryiii:
I need this to work on the UHI reprs #156, so will merge.

PR #165, opened 2019-10-22
Adding copy_in

henryiii:
This is the beginning of work for a `_copy_in` function; this will power (at least) `h[...] = np.array(...)`, and maybe operations with histograms and arrays. Going to get this working with doubles first, then generalize.

Currently, double only.


henryiii:
I'm going to merge, so that I have a single branch for the restructuring of #180 if I get to it on the flight to Australia. There's one fix (nicer accumulator printout), and a private feature (double-only `_copy_in`) for testing.

PR #166, opened 2019-10-22
Unify serialization

HDembinski:


HDembinski:
Do we care about being able to unpickle which were written by a previous version of boost-histogram? boost::histogram can do this, but in here we do not save the version information which is required to achieve this. Saving the version would add an additional integer to the tuple of every pickleable boost-histogram object.

To-do:
- [ ] Use array optimization for unlimited storage (pending on changes in boost::histogram)
- [ ] Save version so that it is possible to unpickle older saves with newer versions of boost-histogram

henryiii:
I think it should store the version, but not for the surface reason. I don't expect or recommend pickle for any long term data storage (you can execute arbitrary commands in a pickle file, for instance). Pickling is useful for short term data storage and multiprocessing (since, before 3.8, it was used to communicate between processes - 3.8 is supposed to have a way around but haven't tried it yet).

By the same argument, a tiny bit of extra data won't hurt this usage noticeably either.

However, it looks like we may have a system we can reuse with a different backend. For example, we could possibly use the same code but implement an hdf5 or json backend, storing exactly the same structure but in a different format - and that would be useful (and possibly could allow Python histograms to be read in C++, and vice versa, if a standalone version of the header was available). So eventually we could add a `to_<something>` and provide a header that lets users who have the requirements read it in C++. If the structure is as close as possible, maybe this could be done? Then having the number would be important.

henryiii:
Users may also be copying histograms to different machines that have different versions of boost-histogram and expect it to work. We ran into that with ROOT and the LZMA default (writing files in 6.12.00 and expecting them to be read in 6.06.xx). While pickle protocol version settings will always be a potential problem if they are changing major Python versions, if we can avoid boost-histogram adding another issue, that would be very nice.

HDembinski:
To-do: Write/read class version only once for the first object of a class encountered in one serialization operation.

HDembinski:
I feel like I am rediscovering all the improvements that Robert added to Boost.Serialization over the years. Btw, I think cereal is not really a better library. They boast about using less bits, yes, but they also offer less functionality. It is easy to be more slim when you drop features.

henryiii:
`python/accumulators_ostream.hpp` was added, that's the source of the merge conflicts, FYI.

HDembinski:
Looks like your accumulators_ostream.hpp is broken?

henryiii:
You can't include both copies of ostream. All ostream headers should come from the python folder.

include/boost/histogram/python/register_accumulator.hpp
```diff
@@ -7,7 +7,8 @@
 
 #include <boost/histogram/python/pybind11.hpp>
 
-#include <boost/histogram/python/serializion.hpp>
+#include <boost/histogram/accumulators/ostream.hpp> // TODO: replace with internal repr
```
henryiii:
```suggestion
#include <boost/histogram/python/accumulators_ostream.hpp>
```

src/register_axis.cpp
```diff
@@ -156,15 +156,15 @@ void register_axes(py::module &mod) {
...
+        {{"_category_int", "_category_int_growth"}},
         "Axis with discontiguous integer bins",
         py::init<std::vector<int>, metadata_t>(),
         "categories"_a,
         "metadata"_a = py::none());
 
     register_axis_sub_types<axis::category_str, axis::category_str_growth>(
         mod,
-        {"_category_str", "_category_str_growth"},
+        {{"_category_str", "_category_str_growth"}},
```
henryiii: Doesn't matter, but why are these now double bracketed?

include/boost/histogram/python/regular_numpy.hpp
```diff
@@ -35,9 +35,8 @@ class regular_numpy : public bh::axis::regular<double, bh::use_default, metadata
 
     template <class Archive>
     void serialize(Archive &ar, unsigned version) {
-        static_cast<bh::axis::regular<double, bh::use_default, metadata_t> *>(this)
-            ->serialize(ar, version);
-        ar &serialization::make_nvp("stop", stop_);
+        regular::serialize(ar, version);
```
henryiii: Very nice :)

src/register_axis.cpp
HDembinski: clang-5 was complaining about it.

PR #183, opened 2019-10-31
Wrapped design

### LINKED ISSUE #155: Provide cpp module
There are some pythonisations (mostly related to arrays and indexing) that duplicate the C++ API. We could add a `boost_histogram.cpp` module, and this would provide the exact Boost.Histogram API where possible.

Things to move or provide there:

```python
.at(i) # In Python: [] and bh.underflow/overflow
.axis(i) # Tuple `.axes` in Python
```

This might be supportable alongside `boost_histogram`, or it might inject the methods directly to the existing classes. I'm not sure I'm very excited about creating a new collection of classes -- but if it can be done carefully, it might work (Basically we need Ruby's `extend`...).

### LINKED ISSUE #174: `help(boost_histogram.histogram)` produces unhelpful result
```
Help on FactoryMeta in boost_histogram object:

boost_histogram.histogram = class FactoryMeta(builtins.object)
 |  boost_histogram.histogram(f, types)
 |
 |  Methods defined here:
 |
 |  __call__(self, *args, **kwargs)
 |      Call self as a function.
 |
 |  __init__(self, f, types)
 |      Initialize self.  See help(type(self)) for accurate signature.
 |
 |  __instancecheck__(self, other)
 |      Check if an object is an instance.
 |
 |  ----------------------------------------------------------------------
 |  Data descriptors defined here:
 |
 |  __dict__
 |      dictionary for instance variables (if defined)
 |
 |  __weakref__
 |      list of weak references to the object (if defined)
```
The expectation is to get help on the histogram object.

### LINKED ISSUE #176: Add functions to AxesTuple
The functions on axes can generalize as well, and can be very useful values -> bins or bins -> values in one quick request, rather than calling each function one by one.

For example, on a 2D histogram:

```python
x, y = h.axes.index(a, b)
# VS
x = h.axes[0].index(a)
y = h.axes[1].index(b)
```

henryiii:
This changes the design to hold an internal object, and users only interact with the Python classes.

#### File layout changes

The C++ core is now named `_core`, and non-public Python parts are in `_internal`. The `cpp` module is new (see below). AxesTuple is in now in `_internal/axestuple.py`, `_utils` is now `_internal/kwargs.py` (the old meta utils are gone, no longer needed). `uhi.py` is now `tag.py`.

#### Histogram design:

- `BaseHistogram`: Shared interface
    - `BoostHistogram`: Mimics the C++ interface as closely as possible. Even has `__call__`!
    - `Histogram`: The default interface, with Pythonizations. Caches the axes tuple.

You can convert between the two, or add a new class. If you do:

```python
import boost_histogram.cpp as bh
```

You will now be working with BoostHistograms instead of Histograms. (closes #155)

#### Axes design:

There is a base class `Axis`, along with `Regular`, etc. for the 4 types. The axis object is stored in `_ax`. The function forms are now class methods, `regular.sqrt` for example (they are still available as `regular_sqrt` for the moment, as well). A helper function internally knows how to convert an `axis` object into the correct `Axis` class, and is used when pulling an axis from a histogram.

#### General changes

The underscores on some C++ parts have been removed. An underscore in the C++ bindings indicates that something is different than the C++ code. The C++ `_core` objects have been stripped down a little, and exactly mimic C++ where possible (`.rank()` is a function, etc). The Pythonisation now happens in `Histogram`. Axes tuples have functions now too (closes #176), and are finally tested.

Also, more docs present, closes #174.


henryiii:
Let's go ahead and merge this into develop, and review/changes can be PRs. I need this in so I work on the next set of changes as new branches on the flight to Australia. Anything here can be reverted/modified in develop if needed.

henryiii:
(Actually, @HDembinski, to make it easier to review, I'll leave this open, but build branches off of it on the plane, but will need to merge before proceeding).

henryiii:
Please raise issues if there are parts of the design that could be better! I've already been working off this, and it seems to be working pretty well.

CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
+### IN DEVELOPMENT
+
+#### User changes
+
+* You can now access the functional regular axis directly, `regular_sqrt` becomes `regular.sqrt`, etc.. [#183][]
```
HDembinski: No `regular.sqrt`. The transform should be a passed as a keyword to the regular class.

boost_histogram/_internal/axis.py
```diff
@@ -0,0 +1,373 @@
...
+        elif options == {"overflow"}:
+            self._ax = ca.regular_oflow(bins, start, stop, metadata)
+        elif options == {"circular", "underflow", "overflow"}:
+            self._ax = ca.circular(bins, start, stop, metadata)
+        elif options == set():
+            self._ax = ca.regular_none(bins, start, stop, metadata)
+        else:
+            raise KeyError("Unsupported collection of options")
+
+    @classmethod
```
HDembinski: This is a change that slipped by me. I think we should not have these class methods, since people should use the `transform` keyword.

PR #185, opened 2019-11-02
UHI update

### LINKED ISSUE #135: Partial projection
Add slice/shrink + project support.

henryiii:
This brings UHI more inline with the original proposal from @jpivarski , and removes the internal tags originally required (`.value` and `.offset`). The original proposal is that the indexers would be callable functors, and they would receive the axis when being called. This simplifies `bh.loc`, and also makes it much easier for users to write there own. In fact, `len` naturally works:

```python
h[0:len:bh.project]
```

Will sum without including the flow bins. This also includes the trick proposed by @HDembinski, where cropping is simulated through setting flow bins on a copy to 0, so it now finally allows limits on sums! Closes #135. `bh.sum` is available now, as well, though `bh.project` is still around for the moment.

We need to think a little about the difference in slicing and direct access. Currently,
```python
h[bh.overflow]
```
gets the overflow bin. While:

```python
h[bh.underflow:bh.overflow:bh.project]
```

does not include the overflow bin, following the standard meaning of Python indexing. While this is incorrect (you should leave endpoints off to get the overflow bin). I *think* this is consistent and correct, but might be surprising to users. ~~It also means `bh.end` would only work in `h[bh.end]`, while `h[:bh.end]` would be one less than the end~~ (`len` or `bh.overflow` would be used to slice to the end). I've removed end.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,11 @@
 
 #### User changes
 
+* You can now sum over a range with endpoints [#185][]
 * You can now access the functional regular axis directly, `regular_sqrt` becomes `regular.sqrt`, etc.. [#183][]
 * `h.axis()` and `h.at()` are now completely removed unless you use the cpp version of histogram. [#183][]
 * `h.axes` now has the functions from axis as well. [#183][]
+* `bh.project` has become `bh.sum` [#185][]
 
 
 #### Developer changes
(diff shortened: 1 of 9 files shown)


henryiii:
We can support addition and subtraction on flow tags, which might help. We can also drop end, for missusage reasons, and suggest users use len or bh.overflow for removing the final flow bin in cropped sums. I think that `h[len]` is clearly one past the end (N vs N - 1), and `bh.overflow` is explicit, while end will be misused in one of the two places.

End does but have the same problem in Matlab, where indexing includes the end point.

henryiii:
I have also added `bh.tag.at`, which recovers Boost.Histogram indexing in `[]` indexing; the user-defined callables are quite powerful!

Issue #175, opened 2019-10-25
Retrieving results of non-trivial storages


beojan:
It doesn't seem to be possible to retrieve the results as a NumPy array when a non-trivial storage is used (e.g. `storage.weight`). It's known that `view` isn't supported, but `np.asarray` also crashes the Jupyter kernel.

henryiii:
It is not currently possible - a decision has/had to be made about that the storages produce as a numpy array. I think we do have a plan now, but it has not been implemented yet. There is an incoming feature to Boost.Histogram C++ (I think today) that will help us here.

`h.view()` and` np.asarray(h)` use the same underlying machinery.

As a workaround, you can still iterate over the histogram cells and access them one-by-one as Python objects.

beojan:
> a decision has/had to be made about that the storages produce as a numpy array

I would expect to be able to access each stored quantity as an array (i.e. `.values()` and `.variances()` as for individual cells in the C++ API).

henryiii:
But then the returned item is no longer a numpy array (killing `.asarray`). We could possibly introduce a difference between `asarray` and `.view()` here, though.

Issue #21, opened 2019-03-29
h.view() for all storage types


HDembinski:

HDembinski:
I will try to work on this next, now that I have a good feeling with the threaded support in boost::histogram.

henryiii:
I don't think this will work; I'm pretty sure you cannot provide a view to a boost multi-precision int.

---

If that wasn't there, it would work, but would still be invalidated after filling if the type changed. So:

```python
v = np.asarray(h)
h.fill(0.5)
# Access to v is now ruined
v *= 0.5
# Access to v is now really ruined
```

(The nice thing is it can't shrink, so at least you won't get a segfault, I guess).

We can either drop the unlimited type, or disable these sorts of things and provide a copy method to get data. (The multi precision int will still be a problem)

HDembinski:
The view of an unlimited_storage is always ruined, when you modify the type of the counters in C++. That either needs to be documented, or we always convert the internal counters to double when someone calls a view. I am for the second option.

HDembinski:
I propose the following solution.
* We do not allow `np.asarray(hist)`, calling that will generate an error (how to implement this?). The only way to get a cell view is `hist.view()`.
* When the user calls `hist.view()` and the storage is `unlimited_storage`, we transform the internal types to `double`. Then we forward the view into the double buffer.

This avoids all problems with views becoming ruined.
* It is a safe solution for the user.
* It is simple to implement.
* It is consistent with the concept of unlimited_storage. To the outside world, it presents its cells as doubles. How it counts internally is supposed to be an implementation detail.
* People call `view()` after the histogram is filled to work with the cells, not in between. When they call it, it is e.g. to normalize the histogram. Practically all of these transformations need to convert the counters to double anyway.

henryiii:
I was just thinking about turning off `np.asarray(hist)`; while we could allow it for simple storages, might be better to be consistent. Especially if we change the default to the unlimited storage.

HDembinski:
1) I think we should keep it and rather document the issue with unlimited_storage. In normal usage, I cannot see how this is going to be a problem, because people fill first and then make an array view of the histogram after the filling to use the counters.

2) I am open to not have unlimited storage as the default. The question is whether we want to be memory conserving or fast by default. The unlimited_storage used to be faster in the past, but it got slower over time. I am not sure why. It may be possible to make it faster again, but that probably costs significant time.

HDembinski:
[...]

PR #186, opened 2019-11-03
WIP: Access accumulator storages

henryiii:
This allows direct memory access to non-simple storages, no longer throws error/segfault and fixes the most common error reported by users, #175 and on gitter. Original issue: #21. Currently, it is mostly a proof of concept, as a future plan below must be chosen.

This method *requires* direct memory access into accumulators. This is currently not a public part of Boost.Histogram, so could potentially be broken in the future. This could be solved by:

1. Making this a promised ABI in Boost.Histogram. This would at least need special constructors for some accumulators to allow them to be made from the raw values.
2. Make the internal values public in Boost.Histogram, thereby officially making this available and unchangeable, which would allow `PYBIND11_NUMPY_DTYPE` to be used to make the RecordArray.
3. Copy the accumulator code into boost-histogram, so that we could use option 2) and not depend on cluttering/loosening/promising in Boost.Histogram. This would mean we would have to copy in changes (like to the weight that happened recently).

Direct memory access into accumulators is currently needed for getting and setting accumulator storages with array expressions. In the future, the proposed numbafication project will probably also need direct access to storage data.

Once we have this, we can provide a Python NDArray subclass or proxy object that will keep the details (partially) hidden from users, so they can access arrays of values with the same syntax. as a normal accumulator and perform some operations. 

I locally have started on the wrapper object, but since I can't construct accumulators from the raw values, this keeps `array[i]` from producing an accumulator.

Example of current (unwrapped) version:

```python
In [1]: h = bh.histogram((10,0,1), storage=bh.storage.weight)

In [2]: h.fill(np.random.rand(100), weight=np.random.rand(100))
Out[2]:
Histogram(
  regular(10, 0, 1),
  storage=weight
) # Sum: weighted_sum(value=50.34, variance=35.681)

In [3]: h.view()
Out[3]:
array([(6.7629267 , 5.11955394), (4.25568444, 2.48312512),
       (6.62707634, 5.08164912), (3.273183  , 2.09949326),
       (6.28007822, 4.4779143 ), (5.38847563, 3.6516565 ),
       (8.79954476, 6.84868563), (2.73569681, 1.65033745),
       (1.36933037, 0.86418926), (4.8479874 , 3.40439761)],
      dtype=[('sum_of_weights_', '<f8'), ('sum_of_weights_squared_', '<f8')])
```

### DIFF
--- a/include/boost/histogram/python/histogram.hpp
+++ b/include/boost/histogram/python/histogram.hpp
@@ -7,6 +7,10 @@
 
 #include <boost/histogram/python/pybind11.hpp>
 
+#include <boost/histogram/python/accumulators/mean.hpp>
+#include <boost/histogram/python/accumulators/weighted_mean.hpp>
+#include <boost/histogram/python/accumulators/weighted_sum.hpp>
+
 #include <boost/histogram/detail/axes.hpp>
 #include <boost/histogram/histogram.hpp>
 #include <boost/histogram/unsafe_access.hpp>
@@ -19,46 +23,6 @@ namespace pybind11 {
 template <class T>
 struct format_descriptor<bh::accumulators::thread_safe<T>> : format_descriptor<T> {
     static_assert(std::is_standard_layout<bh::accumulators::thread_safe<T>>::value, "");
-};
-
-/// This descriptor depends on the memory format for mean<double> remaining unchanged
-template <>
-struct format_descriptor<bh::accumulators::mean<double>> {
-    static std::string format() {
-        return std::string("T{"
-                           "d:sum_:"
-                           "d:mean_:"
-                           "d:sum_of_deltas_squared_:}");
-    }
-};
-
-// If made public, could be:
-// PYBIND11_NUMPY_DTYPE(mean, sum_, mean_, sum_of_deltas_squared_);
-
-/// This descriptor depends on the memory format for weighted_mean<double> remaining
-/// unchanged
-template <>
-struct format_descriptor<bh::accumulators::weighted_mean<double>> {
-    static std::string format() {
-        return std::string("T{"
-                           "d:sum_of_weights_:"
-                           "d:sum_of_weights_squared_:"
-                           "d:weighted_mean_:"
-                           "d:sum_of_weighted_deltas_squared_:"
-                           "}");
-    }
-};
-
-/// This descriptor depends on the memory format for weighted_sum<double> remaining
-/// unchanged
-template <>
-struct format_descriptor<bh::accumulators::weighted_sum<double>> {
-    static std::string format() {
-        return std::string("T{"
-                           "d:sum_of_weights_:"
-                           "d:sum_of_weights_squared_:"
-                           "}");
-    }
 };
 
 } // namespace pybind11
--- a/include/boost/histogram/python/histogram.hpp
+++ b/include/boost/histogram/python/histogram.hpp
@@ -13,10 +13,54 @@
 #include <type_traits>
 
 namespace pybind11 {
+
+/// The descriptor for atomic_* is the same as the descriptor for *, as long this uses
+/// standard layout
 template <class T>
 struct format_descriptor<bh::accumulators::thread_safe<T>> : format_descriptor<T> {
     static_assert(std::is_standard_layout<bh::accumulators::thread_safe<T>>::value, "");
 };
+
+/// This descriptor depends on the memory format for mean<double> remaining unchanged
+template <>
+struct format_descriptor<bh::accumulators::mean<double>> {
+    static std::string format() {
+        return std::string("T{"
+                           "d:sum_:"
+                           "d:mean_:"
+                           "d:sum_of_deltas_squared_:}");
+    }
+};
+
+// If made public, could be:
+// PYBIND11_NUMPY_DTYPE(mean, sum_, mean_, sum_of_deltas_squared_);
+
+/// This descriptor depends on the memory format for weighted_mean<double> remaining
+/// unchanged
+template <>
+struct format_descriptor<bh::accumulators::weighted_mean<double>> {
+    static std::string format() {
+        return std::string("T{"
+                           "d:sum_of_weights_:"
+                           "d:sum_of_weights_squared_:"
+                           "d:weighted_mean_:"
+                           "d:sum_of_weighted_deltas_squared_:"
+                           "}");
+    }
+};
+
+/// This descriptor depends on the memory format for weighted_sum<double> remaining
+/// unchanged
+template <>
+struct format_descriptor<bh::accumulators::weighted_sum<double>> {
+    static std::string format() {
+        return std::string("T{"
+                           "d:sum_of_weights_:"
+                           "d:sum_of_weights_squared_:"
+                           "}");
+    }
+};
+
 } // namespace pybind11
 
 namespace detail {
--- a/boost_histogram/_internal/hist.py
+++ b/boost_histogram/_internal/hist.py
@@ -4,6 +4,7 @@ from .kwargs import KWArgs
 
 from .. import _core
 from .axis import _to_axis, Axis
+from .view import _to_view
 from .axistuple import AxesTuple
 from .sig_tools import inject_signature
 from .storage import Double, _to_storage
@@ -137,7 +138,7 @@ class BaseHistogram(object):
         return self.__class__.__name__ + repr(self._hist)[9:]
 
     def __array__(self):
-        return np.asarray(self._hist)
+        return _to_view(np.asarray(self._hist))
         # TODO: .view does not seem to return an editable view
         #        so we have to use the buffer interface here
 
@@ -298,7 +299,7 @@ class Histogram(BaseHistogram):
         """
         Return a view into the data, optionally with overflow turned on.
         """
-        return self._hist.view(flow)
+        return _to_view(self._hist.view(flow))
 
     def reset(self):
         """
--- a/boost_histogram/_internal/view.py
+++ b/boost_histogram/_internal/view.py
@@ -1,0 +1,92 @@
+from __future__ import absolute_import, division, print_function
+
+from ..accumulators import Mean, WeightedMean, WeightedSum
+
+import numpy as np
+
+
+class View(np.ndarray):
+    __slots__ = ()
+
+    def __getitem__(self, ind):
+        sliced = super(View, self).__getitem__(ind)
+        if sliced.shape:
+            return sliced
+        else:
+            return self._PARENT._make(*sliced)
+
+    def __repr__(self):
+        return repr(self.view(np.ndarray))
+
+    def __str__(self):
+        return str(self.view(np.ndarray))
+
+
+class MeanView(View):
+    __slots__ = ()
+    _PARENT = Mean
+    _FIELDS = ("sum_", "mean_", "sum_of_deltas_squared_")
+
+    @property
+    def count(self):
+        return self["mean_"]
+
+    @property
+    def value(self):
+        return self["sum_"]
+
+    # Variance is a computation
+    @property
+    def variance(self):
+        return self["sum_of_deltas_squared_"] / (self["sum_"] - 1)
+
+
+class WeightedSumView(View):
+    __slots__ = ()
+    _PARENT = WeightedSum
+    _FIELDS = ("sum_of_weights_", "sum_of_weights_squared_")
+
+    @property
+    def sum_of_weights(self):
+        return self["sum_of_weights_"]
+
+    @property
+    def sum_of_weights_squared(self):
+        return self["sum_of_weights_squared_"]
+
+
+class WeightedMeanView(View):
+    __slots__ = ()
+    _PARENT = WeightedMean
+    _FIELDS = (
+        "sum_of_weights_",
+        "sum_of_weights_squared_",
+        "weighted_mean_",
+        "sum_of_weighted_deltas_squared_",
+    )
+
+    @property
+    def sum_of_weights(self):
+        return self["sum_of_weights_"]
+
+    @property
+    def sum_of_weights_squared(self):
+        return self["sum_of_weights_squared_"]
+
+    @property
+    def value(self):
+        return self["weighted_mean_"]
+
+    @property
+    def variance(self):
+        return self["sum_of_weighted_deltas_squared_"] / (
+            self["sum_of_weights_"]
+            - self["sum_of_weights_squared_"] / self["sum_of_weights_"]
+        )
+
+
+def _to_view(item):
+    for cls in View.__subclasses__():
+        if cls._FIELDS == item.dtype.fields:
+            return item.view(cls)
+    return item
--- a/include/boost/histogram/python/accumulators/mean.hpp
+++ b/include/boost/histogram/python/accumulators/mean.hpp
@@ -1,0 +1,128 @@
+// Copyright 2015-2019 Hans Dembinski and Henry Schreiner
+//
+// Distributed under the Boost Software License, version 1.0.
+// (See accompanying file LICENSE_1_0.txt
+// or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Based on boost/histogram/accumulators/mean.hpp
+
+#pragma once
+
+#include <boost/assert.hpp>
+#include <boost/core/nvp.hpp>
+#include <boost/histogram/fwd.hpp> // for mean<>
+#include <boost/throw_exception.hpp>
+#include <stdexcept>
+#include <type_traits>
+
+namespace boost {
+namespace histogram {
+namespace python {
+
+/** Calculates mean and variance of sample.
+
+  Uses Welfords's incremental algorithm to improve the numerical
+  stability of mean and variance computation.
+*/
+template <class RealType>
+class mean {
+  public:
+    mean() = default;
+    mean(const RealType &n, const RealType &mean, const RealType &variance) noexcept
+        : sum_(n)
+        , mean_(mean)
+        , sum_of_deltas_squared_(variance * (n - 1)) {}
+
+    void operator()(const RealType &x) noexcept {
+        sum_ += static_cast<RealType>(1);
+        const auto delta = x - mean_;
+        mean_ += delta / sum_;
+        sum_of_deltas_squared_ += delta * (x - mean_);
+    }
+
+    void operator()(const weight_type<RealType> &w, const RealType &x) noexcept {
+        sum_ += w.value;
+        const auto delta = x - mean_;
+        mean_ += w.value * delta / sum_;
+        sum_of_deltas_squared_ += w.value * delta * (x - mean_);
+    }
+
+    template <class T>
+    mean &operator+=(const mean<T> &rhs) noexcept {
+        if(sum_ != 0 || rhs.sum_ != 0) {
+            const auto tmp = mean_ * sum_ + static_cast<RealType>(rhs.mean_ * rhs.sum_);
+            sum_ += rhs.sum_;
+            mean_ = tmp / sum_;
+        }
+        sum_of_deltas_squared_ += static_cast<RealType>(rhs.sum_of_deltas_squared_);
+        return *this;
+    }
+
+    mean &operator*=(const RealType &s) noexcept {
+        mean_ *= s;
+        sum_of_deltas_squared_ *= s * s;
+        return *this;
+    }
+
+    template <class T>
+    bool operator==(const mean<T> &rhs) const noexcept {
+        return sum_ == rhs.sum_ && mean_ == rhs.mean_
+               && sum_of_deltas_squared_ == rhs.sum_of_deltas_squared_;
+    }
+
+    template <class T>
+    bool operator!=(const mean<T> &rhs) const noexcept {
+        return !operator==(rhs);
+    }
+
+    const RealType &count() const noexcept { return sum_; }
+    const RealType &value() const noexcept { return mean_; }
+    RealType variance() const noexcept { return sum_of_deltas_squared_ / (sum_ - 1); }
+
+    template <class Archive>
+    void serialize(Archive &ar, unsigned version) {
+        if(version == 0) {
+            // read only
+            std::size_t sum;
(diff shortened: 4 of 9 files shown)


HDembinski:
Number 3 seems the simplest route. I don't want to expose the internals of the accumulators in boost::histogram.

include/boost/histogram/python/histogram.hpp
```diff
@@ -13,10 +13,54 @@
...
+// If made public, could be:
+// PYBIND11_NUMPY_DTYPE(mean, sum_, mean_, sum_of_deltas_squared_);
+
+/// This descriptor depends on the memory format for weighted_mean<double> remaining
+/// unchanged
+template <>
+struct format_descriptor<bh::accumulators::weighted_mean<double>> {
+    static std::string format() {
+        return std::string("T{"
+                           "d:sum_of_weights_:"
```
HDembinski: The trailing underscore should go away, but I guess that's implied for the final version.

henryiii:
> Number 3 seems the simplest route. I don't want to expose the internals of the accumulators in boost::histogram.

I think so too. I need three things: direct memory access, custom constructors, and standalone transforms for properties - none of which should be in Boost.Histogram accumulators.

Local tests on NDArray subclasses and structured arrays look very promising.

HDembinski:
And the accumulators are very simple, so it is not too bad to duplicate them in boost-histogram. In the long run, the accumulators in boost::histogram and boost-histogram may diverge in functionality, but that is ok and perhaps even desirable, so that pure-Python users have stability.

HDembinski:
> `In [1]: h = bh.histogram((10,0,1), storage=bh.storage.weight)`

We agreed that a storage should not be used like an enum. The call must be
`bh.histogram((10,0,1), storage=bh.storage.weight())`. There is a specific reason for this: there will be storages which accept a run-time configuration. For example, when you want to make storage that is a mmap'd file or to make a storage that accepts a variable number of weights, when the number is only known at run-time.

HDembinski:
Besides, it should be `bh.storage.Weight()` then.

henryiii:
> We agreed that a storage should not be used like an Enum

No, you said you were not sure if you liked the shortcut, you never said we could not have it. It still is not an enum, it just has optional `()`.

However, that was when it was called `weight`. Now that it's called `Weight`, it now looks like a Python class instead of an enum, and the `()` is now natural.

henryiii:
Actually, shoudn't this work just like transform? For transform, `transform=bh.axis.transform.Log` or `transform=bh.axis.transform.Pow(2)` works. And, I was careful to make sure that `transform=bh.axis.transform.Log()` is not an error either - it's really just duck-typing on a method, and the non-configurable ones have a static method. So `bh.storage.Weight` would trigger based on a method that exists regardless of whether the `()` are included. If a new configurable storage is added, then it would require the `(conf)` part, like `Pow`.

henryiii:
(PS: remember, we are talking about the Python objects that will be used to select storages, not the storages themselves)

henryiii:
This is exactly how good decorators work in Python. Users do not understand that

```python
@dec()
def f(): pass
```

is totally different than:

```python
@dec
def f(): pass
```

So, decorators are generally carefully designed to hide this difference. So you can write

```python
@numba.jit
def f(): ...

@numba.jit(nopython=True)
def f(): ...
```

And they both work; if you've ever designed a decorator you'll know it's not trivial to make that happen.

henryiii:
Plumbum's modifers work this way too.

```python
local.ls & FG
```

and 

```python
local.ls & FG()
```

are exactly the same (They support keyword arguments).

HDembinski:
Transforms and storages are classes and they should be passed with `()` for clarity. This is what makes sense from the consistency point of view. I want users to have a clear understanding what is a class, what is a value and so on.

HDembinski:
boost-histogram should not accept a class where a class instance is required. The distinction is already blurred in Python, which is bad.

henryiii:
> The distinction is already blurred in Python

You are thinking in C++. There is no distinction in Python. A class *is* an instance, of `type`. Or, you set a metaclass, an instance of some other class.

For a *user*, they don't really care about what a storage internally is. They want to make a histogram with as few errors and as little friction as possible. For 95% of the users, the existing collection of storages will probably be fine. We can always use `Double()`, etc. in the examples/docs, but should we knowingly cause a common mistake to throw an error when it does not need to just because we want to teach them something that has nothing to do with histograms? Numba could try to teach the difference between decorators and decorator factories, but it doesn't, since that's not the point of the package.

If you want a rational for `storage=Double` also working, it's just this: if you pass a storage type, then it constructs the default storage. You can't do this in C++ because you can't pass a type because it's not an object. You can pass a type through the template arguments, but Python doesn't have that.

> Note: As you know, I'm often influenced by Ruby, and in Ruby, all paratheses are optional if unambiguous. There is no difference between `a.x()` and `a.x`. And `a.x=2` really is just `a.x=(2)`, where `x=` is the method name, etc. It's quite clever, actually.

henryiii:
Also, passing a *constructed* class is really quite odd in Python. You do not write `np.array([], dtype=int())` in numpy, for example. Usually, you pass the class and then the thing you are giving them constructs them for you. When we add `to_dataframe` functionality, we will have something like `frame = to_dataframe(pd.Dataframe)`; `frame = to_dataframe(pd.Dataframe())` would not work because we need to construct the dataframe with our histogram output, not with the default constructor. This is normal in Python.

HDembinski:
> You are thinking in C++. There is no distinction in Python. A class is an instance, of type. Or, you set a metaclass, an instance of some other class.

No, I am not. As you have pointed out, classes are  also special in Python, since there can be only one class "value" in the whole program.

HDembinski:
Your decorator analogy is not convincing. Nobody understands how decorators work, but most people think of classes and instances as separate things.

HDembinski:
The problem is that you sometimes care greatly about consistency and then in other cases you are happy to throw consistency out of the window. You cannot teach an interface if there are no clear rules followed. The `storage` keyword accepts are storage value, therefore you need the braces.

HDembinski:
@jpivarski Could you please join the discussion? Am I thinking too much in terms of C++, if I want to clearly distinguish between classes and instances of classes?

HDembinski:
> Also, passing a constructed class is really quite odd in Python. You do not write np.array([], dtype=int()) in numpy, for example.

The `dtype` keyword does not randomly accept both classes and values of those classes. You cannot pass an instance of `int`, only the `int` class. For the numpy `dtype` class it is the opposite, there you can only pass an instance of the `numpy.dtype` class and not the class itself.

jpivarski:
What's the issue? Starting at the top of this issue, it's about being able to access accumulator storage in Python. We'll need to have a Numpy-like view or copy of the storage bytes with a stable ABI in order to do meaningful things with it, particularly in Numba. But that doesn't seem to be what the discussion is about at the end.

At the end, it looks like the question is whether one should pass a class instance or a class object as an argument to some function. Despite what's been said about there being no distinction (a class object is an instance of `type`, or whatever its metatype happens to be), one of the two usually makes more sense than the other. Usually, class instances make more sense, but one exception is uproot's `outputtype` argument, which takes a class object of the _type_ of container you want to get back. It is because it's parameterizing the _type_ of container you want back that it wants a _type_ as its argument (e.g. `dict`, `list`, `tuple`, `pandas.DataFrame`). Is that the situation you're talking about here, or is it a different situation?

Is the argument list in question in Python or in C++? I don't think it's even possible for a type/class to be an argument of a function in C++; it can only be an argument of a template, but Python has only one "arena," so to speak.

henryiii:
Boost histogram takes the storage *instance* as a parameter when making a histogram. So you could make a std::vector, and then pass the instance to Boost histogram, and it fills it. We provide a set (compiled-in) collection of these storages, and there is no allowed API on them except construction.

I want to take an invalid operation, `storage=Double`, and turn it into the valid `storage=Double()` instead of throwing an error. I am sure we will eventually get an issue or email asking why some bit of code doesn't work because someone forgot to add the `()` (it's better now that Double is clearly a class, but passing classes to methods in this way is rather common, such as in Numpy).

henryiii:
And, yes, we took over an unrelated issue.

henryiii:
Yes, this is in Python, it would be impossible to do in C++. Also in C++, you can use any valid instance, while there's a small compiled list of them in Python.

henryiii:
Not direclty related: @HDembinski, what should happen if I do:

```python
d = Double()
h1 = histogram(..., storage=d)
h2 = histogram(..., storage=d)
```
 
Currently, the "Double" wrapper actually doesn't create the wrapped instance until it's asked, and it creates it on demand, so the above code is valid. Should it be?

jpivarski:
Okay. In the example I gave, uproot's `outputtype`, there's no mixture: this parameter _only_ takes a type, never an instance. The vast majority of arguments take an instance and not a type.

So the proposal here is to have a `storage` parameter that _either_

   * takes an instance like `storage=Double()` and fills that object, changing it in-place, or
   * takes a class like `storage=Double` (or at least a callable returning an instance), calls it, and fills the object that has been created (which is out of the user's reach)?

To me, thinking with a Python background, that sounds normal. It mixes two rather different things, represented by the two bullet points above, but each of them do what the user likely wants. There's precedent—Python exceptions work this way:

```python
raise Exception()   # I pass an instance; that instance is raised
```

and

```python
raise Exception     # I pass a class that gets instantiated; the instance is raised
```

(Or for a more realistic example, `NotImplementedError()` vs `NotImplementedError`.)

Some users, aware of one form, might not realize that the other form is allowed. Some might intend one form and get the other, but the effect is what they wanted anyway. Most users would not want an error message here; it would seem like nit-picking (though it is an admittedly real distinction). The one sense in which this would bother me is if I'm trying to keep a codebase clean—flake8 probably prefers one over the other. But code linting is more strict than everyday use.

HDembinski:
> Not direclty related: @HDembinski, what should happen if I do:
> 
> ```python
> d = Double()
> h1 = histogram(..., storage=d)
> h2 = histogram(..., storage=d)
> ```
> 
> Currently, the "Double" wrapper actually doesn't create the wrapped instance until it's asked, and it creates it on demand, so the above code is valid. Should it be? Does this mean they write to the same place in memory, or does one append after the other?

This is exactly correct. Remember that there will be storages in the future which can be configured, which accept parameters. Then to save typing, you may want to pass a value to several histograms.

```
s = multi_weight_storage(32) # storage with 32 parallel weights
h1 = histogram(..., storage=s)
h2 = histogram(..., storage=s)
```

jpivarski:
> The problem is that you sometimes care greatly about consistency and then in other cases you are happy to throw consistency out of the window. You cannot teach an interface if there are no clear rules followed. The `storage` keyword accepts are storage value, therefore you need the braces. Likewise for transforms.

I'm getting up-to-speed on this conversation. I think of Python as a strongly typed (though dynamic) language. A _good_ function/method declaration requires specific types for each argument, though the dynamism may allow that to be a late decision (during the "formalization" of an existing codebase) and duck typing may allow a general "type class" (in the Haskell sense; sorry for the word overloading!) or "interface" (in the Java/Scala sense).

However, a function/method may take one of several specific types, like a `std::variant`, and it may have different behaviors, depending on those types. If I'm understanding the case correctly here, those types are "class instance" versus "class object" and they each invoke a different behavior, just as "exception instance" and "exception class object" invoke different behaviors for `raise`. Those different behaviors happen to mop up likely accidents, such as a user forgetting to add empty parentheses or accidentally adding unintended parentheses, but the two behaviors are well defined and do what the user likely wants.

In that sense, the function/method is not ill-defined, though the user's use of it may be, from time to time. If the user has wants to police their own behavior, they should use a linter. There are no linters for the usage of this library, though. (Full disclosure: I've never used a linter to enforce `NotImplementedError()` vs `NotImplementedError` in my own code; I think that would be overkill.)

Although class objects are technically on the same footing as class instances, we do think of them differently, but only in the sense that they're different types (a function/method argument might require class objects only, might require class instances only, or might allow either, with different behaviors for each), not like they're different kinds (fundamentally distinct, such as C++ classes and instances, which only exist in different passes—compile-time vs run-time).

henryiii:
> This is exactly correct. Remember that there will be storages in the future which can be configured, which accept parameters. Then to save typing, you may want to pass a value to several histograms.

I am now thoroughly confused. Let's say storage is a std::vector-like instance with 3 values in it. Does boost-histogram do a resize(N) for n bins, with the 3 three being prefilled, or resize(N+3) and then point at just the last portion of the vector? What happens if the vector is resized later?

jpivarski:
> This is exactly correct. Remember that there will be storages in the future which can be configured, which accept parameters. Then to save typing, you may want to pass a value to several histograms.
> 
> ```
> s = multi_weight_storage(32) # storage with 32 parallel weights
> h1 = histogram(..., storage=s)
> h2 = histogram(..., storage=s)
> ```

Does this mean that filling `h1` increments values in both `h1` and `h2`, or are different counters created for each? A Pythonic reading of the above has `h1` and `h2` sharing the same incrementable storage (which would likely lead to a mess).

It may be that this is Python's peculiarity—newcomers to Python often don't assume that memory is shared when it is, but maybe people like me get used to the sharing enough that the above looks like it's dangerously sharing.

If the original intention of the `storage` parameter was to take what it's given and create a new instance—perhaps by copying, i.e. pass by value—then I've been misinterpreting it in my comments comparing it to `raise NotImplementedError()` versus `raise NotImplementedError`. However, copy-by-value is a foreign concept in Python.

HDembinski:
> Does this mean that filling h1 increments values in both h1 and h2, or are different counters created for each? A Pythonic reading of the above has h1 and h2 sharing the same incrementable storage (which would likely lead to a mess).

No, it makes a copy. Think of the storage like a `dtype`, it is a configuration value for the histogram.

HDembinski:
And it must be a value, because you want to able to configure the storage at runtime (although all currently available storages do not make use of this feature yet).

I am very concerned about designing interfaces so that they can be extended in the future and that's why it is like this.

henryiii:
> Think of the storage like a dtype.

Then in Python it *should* look like a class. Not an instance, that's the wrong concept. Your description of a dtype means it should be a "type". dtype is a data type. A configurable type in Python uses `[]`, so it would be `MultiWeightStorage[32]`. I now want to remove the `()` and only have the non-`()` version.

HDembinski:
As I have already pointed out above, `numpy.dtype("i8")` is a **value** not a class.

HDembinski:
We should follow the design of C++ Boost.Histogram, unless there is a more Pythonic way. In C++, the configured storage is passed as a value to the histogram, which makes a copy of it.

henryiii:
>>> type(np.int64)
type
>>> np.array([], dtype=np.int64)
array([], dtype=int64)
>>> np.array([], dtype=np.int64())
array([], dtype=int64)

HDembinski:
Numpy is inconsistent then
```
>>> np.array([1, 2], dtype=int())
  File "<stdin>", line 1, in <module>
TypeError: data type not understood
>>>
```

henryiii:
We can do this in the `cpp` module.  The design of the C++ works around the lack of first class type support for arguments. In Python, there are well-defined idoms for these sorts of things, and we should follow those. Types are classes, not instances.

You had *me* confused because I was thinking of the Python meaning of instances. You will confuse every user as well.

> Numpy is inconsistent then

You passed in a Python 0 instance, not a numpy datatype. Python int's don't have set bit length. They can be stored in any number of bits.

The correct thing to pass in is a type, and numpy supports a type instance to avoid common the common mistake of adding (). It also supports strings, the object returned from `dtype()`, and more.

henryiii:
The design I propose is this:

* `Double` is a Python type that produces storage.double() instances *inside* the Histogram constructor.
* `Double()` probably shoudn't throw an error, mostly because it's natural for Boost.Histogram users.
* `MultiWeightStorage[32]` creates a multiweight storage (future)

In the `cpp` module, this instead exactly mimics Boost.Histogram, even forcing the `()`.

Python doesn't have a copy-by-value concept like C++ does (and copying a vector-like storage by value is a bit odd even in C++), and it does have first-class types that can be passed around in functions, as well as type configuration (stdlib `typing` module and `numba` at least use this).

jpivarski:
> As I have already pointed out above, `numpy.dtype("i8")` is a **value** not a class.

Well... remember how we had a conversation at PyHEP about a middle ground between `public` and `private/protected`? This is a middle ground between values and types.

As you know, class objects and instances "live in the same arena" in Python, unlike C++'s classes and instances. I agree that users _think_ of them differently. Python class objects are _usually_ not passed around as values; they're _usually_ defined in neat files like a C++ program. The fact that they can be used interchangeably with values is usually not taken advantage of.

NumPy's `dtypes` define another type system outside of Python's own type system, and they are values representing types in a similar sense as class objects are values representing types. The same is true of `numba.type.*` and `awkward.type.*`. Unlike Python's builtin types and class objects, calling a `dtype` does not construct a NumPy scalar, and `isinstance(scalar, dtype)` doesn't do anything meaningful, so it's not a type in any sense that the Python language recognizes, but it has the same "social function" that class objects as values have to a user of Python—just referring to a different type system: NumPy's type system. (And `numba.type.*` objects represent types in Numba's type system, as objects in `awkward.type.*` represent types in Awkward's high-level type system, distinguished from its low-level type system... If you dig, Numba's `llvmlite` has its own type system, also represented by objects...)

Having types at runtime allows for the superposition of quite a lot of type systems, many more than the ones recognized by the language itself, and with a much less firm distinction than types in a statically typed language.

> Numpy is inconsistent then
> 
> ```python
> >>> np.array([1, 2], dtype=int())
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> TypeError: data type not understood
> ```

In NumPy's specification for this argument, it wants a NumPy type—not a Numba type or an Awkward type or any other type system—except that it will convert a couple of Python types into NumPy types: `int` → `np.dtype(np.int64)` and `float` → `np.dtype(np.float64)` as a matter of convenience. (The first of these is an approximation: Python `int` is arbitrary precision, not 64-bit, though a CPython `float` is 64-bit.)

I'm a little upset with Python that it accepted `int()` with no arguments (returning `0`, what should be `int(0)`), but oh well.

At no point should NumPy accept a numerical value as an argument to its `dtype` parameter. It wants a (NumPy) type, which is technically a value, but even Python class objects are _technically_ values.

henryiii:
> No, I am not. As you have pointed out, classes are also special in Python, since there can be only one class "value" in the whole program.

No, they are not special. Type is slightly special, because it is it's own type, but other than that, types are not special. They are all instances, even type. I never said can only be one class "value"; you can have two if you like. Classes don't have a `__copy__` method, so you have to stamp them yourself:

```python
B = type(A.__name__, A.__bases__, dict(A.__dict__))
```

(The `__dict__` of a class is actually not a normal dict, for performance reasons, and type is picky). 

I'm also assuming your metaclass is `type`; you could of course have a different one, in which case you'd use that to stamp out a new class. Then the type of your class can be something other than type, too:

```python
>>> class NewType(type): pass
>>> class A(metaclass=NewType): pass
>>> type(A)
__main__.NewType
```

The thing that you were trying to do before was to have each instance have it's own `__bases__` but share a single `__class__`. That you can't do, because Python looks up special methods on `a.__class__` and not `a`.

HDembinski:
I don't care about your metaclasses, Henry. There is no need for adding metaclasses here.

I will not accept that boost-histogram deviates needlessly from boost.histogram. In boost.histogram, you pass the storage by value. The storage works like a `np.dtype`. It tells the histogram statically which storage type to use and because it is passed by value, one can pass run-time information with the constructor of the storage. This is a simple and elegant design.

If you hate the braces so severely, then the obvious compromise is to pre-generate values of the storage types in the storage module.
```
# in storage.py
class Int: ...

int = Int()

class Double: ...

double = Double()
```
Then you can do as before
```
import bh
from bh.storage import int

h = bh.Histogram(..., storage=int)
```
There is no break of strong typing, and now it actually makes sense that `int` is with snake case, because that's how it should be for values in Python.

HDembinski:
You can accept this compromise or not, but I will not change my mind about strong typing. I will not accept a design that allows users to pass both storage types and storage values in the histogram init.

henryiii:
> because it is passed by value, one can pass run-time information with the constructor of the storage.

There is no distinction between run-time and compile-time to Python. The design was built for C++, where there is a difference. 

However, I do rather like your idea. How about this design:

* `class Int` is defined in `_internal/storage.py`
* Instance `int` is provided in `storage.py`
* `class int(Int): pass` is provided in `cpp/storage.py` (provides a correct repr)

This keeps the multiple ways to do things more clearly separated.

If a future MultiWeight was added, it could live alongside the instances, and the camel case wouild be an indication that it needs to be instanciated with parameters.

henryiii:
I'm going to close this PR and open a new one with the same branch so we can discuss the PR code, since this has turned into an issue basically.

henryiii:
I still strongly prefer @jpivarski's proposal. The rules are this:

* The docs and examples all show `storage=Double()`. We publically teach that storages are instances.
* If the object is not a storage instance, then it is called and tried - this turns an error into a success. It has strong precedent in Python, with exceptions, `np.float32` vs. `np.float32()` and more. It is physically impossible to allow in C++, so it will not step on Boost.Histogram's toes.

This is closer to Boost.Histogram. It doesn't affect users how know that storages are instances, it even helps teach storages are instances, but it doesn't nitpick users when there is an obvious way to do what they intended.

Option 2, adding separate lowercase types, is not awful, but to me the above solution is much better.

henryiii:
Remember, we have some users downloading Python/Matplotlib for the first time just to use Boost.Histogram. If we add duplication and/or extra ways to throw errors when we don't need to, we will be causing them to struggle more than they have to.

I am not asking for ridiculous shortcuts like`storage=int`, or `dtype=bh.accumulators.weight`, or the such. I am respecting Boost.Histogram's design, but simply allowing a small Pythonisation that many users will expect and/or use by accident.

HDembinski:
It is not a Pythonization when you violate Python's strong typing. Don't tell me that you are respecting Boost.Histogram's design, when you are violating it. Again, the compromise is to keep instances in the storage module, `int = Int()` etc.

PR #188, opened 2019-11-03
Adding signature decorator

### LINKED ISSUE #144: Signatures in Python 3
In Python 3, we can replace the signature of `**kwargs` with the actual keywords-only argument signature. Until we drop Python 2, this will at least mean Python 3 users are not dragged through the ugliness of Python 2.

Won't be ready for PyHEP, though.

henryiii:
This allows Python 3 to get nice signatures while still supporting Python 2. Closes #144.

The histogram and axes constructors (and fill) all have nice signatures now to go along with the nice docstring. When we drop Python 2.7, the contents of the decorator strings can replace current signatures and the decorator (and most/all of the KWArgs tool) will go away.


henryiii:
Reduced the core computation from a ~ hundred lines to 3 lines with a suggestion from @jpivarski.

PR #192, opened 2019-11-05
PEP 8 naming

### LINKED ISSUE #187: PEP 8 naming
In [PEP 8](https://www.python.org/dev/peps/pep-0008/#class-names), Python classes are supposed to use CamelCase, with only the standard library or classes that are mostly called like functions as exceptions. Other than Numpy (which was making a basic type and originally targeting the standard library), most Python packages follow this convention (Pandas, iMinuit, Particle, Awkward, Matplotlib, etc.). Even more recent additions to the standard library follow this PEP 8 convention as well (Path from pathlib, ContextVar from contextvars, Enum from enum, etc.).

(Fun fact: it's actually enforced in Ruby by the language, you can't name a class starting with a lower case letter)

The proposal would be to make the Python classes CamelCase, with the PyBind11 and `cpp` classes lowercase, like in Boost.Histogram. (currently this is what is done, but it is hidden from the user by importing Histogram as histogram, Regular as regular, etc in `__init__.py/axis.py`). I *think* I would suggest that accumulators and storages remain lower-case, because they are more of a data-type than a full class (and are currently not pure Python classes).

henryiii:
Closes #187.

* Nicer CPP mode, more strict, no warnings with old names
* DeprecatedWarning from all old names
* Tests use all the new names
* Tests use `bh.*` instead of importing directly
* TransformAxis and family

TODO:
- [x] Remove view.py (not intentionally added)
- [x] Fix docs/README
- [x] CHANGELOG
- [x] Axis
- [x] Histogram
- [x] AxisTransform
- [x] Storage
- [x] Accumulators


boost_histogram/_internal/hist.py
```diff
@@ -27,10 +27,11 @@ def _arg_shortcut(item):
     elif isinstance(item, _Axis):
         return item._ax
     else:
-        return item
         # TODO: This currently support raw axis object for old tests.
         # Replace with:
         # raise TypeError("Only axes supported in histogram constructor")
+        # TODO: Currently segfaults if we pass in a non-axis to the C++ code
```
HDembinski: Why is this segfaulting?

boost_histogram/cpp/tag.py
```diff
@@ -0,0 +1,115 @@
...
+    __slots__ = ()
+
+    def __call__(self, axis):
+        return len(axis) + self.offset
+
+
+overflow = overflow()
+
+
+class at(object):
```
HDembinski: I see, at has the cpp behavior and loc has the Pythonic behavior, nice.

HDembinski:
Looking good.

HDembinski:
The storages and accumulators must be CameCase'd, too.

HDembinski:
Perhaps change the name of this PR to "Class names".

boost_histogram/_internal/hist.py
henryiii: I accidentally wrote `super(histogram, self).__init__(self, *args, **kargs)`, which caused a Segfault. After (quite a bit of) digging around, I found that PyBind11 was trying to prepare an an error message, since the wrapper "`histogram`" was not a valid argument to `histogram`, but in doing so, it was calling `__repr__` on `histogram`, and since the repr of histogram expected `self._hist` to be available, it was throwing a Python error, which PyBind11 was not catching, and the uncaught error caused a segfault.

I think it's something that can be fixed in PyBind11; making an error message shouldn't segfault if the item in the error message has a faulty repr.

henryiii: (The non-axis must also have a broken `__repr__`)

boost_histogram/cpp/tag.py
henryiii: The system is quite flexible and powerful; users can make all sorts of things - defining at yourself is trivial; I'm using classes here just to get a nicer repr. Once we support Python projects/rebins (0.7.0 or so), we can do the same for the step argument items as well, that will finish the rest of the original UHI proposal.

Note that PEP 8 does allow these to be lowercase; classes that mostly used like functions or values don't have to be CamelCase.

henryiii:
```
E       return Regular(*args, **kwargs, transform=transform.Log)
E                                     ^
E   SyntaxError: invalid syntax
```

Have I mentioned I don't like Python 2?

HDembinski:
> Have I mentioned I don't like Python 2?

You are very clear and vocal about that ;)

henryiii:
@HDembinski, please look at the new wrapper for Storages. It doesn't use any tricks, it just defines "_get_storage_()" as a class method if the storage is not configurable, since it doesn't need self - this is just good Python method design. If we had a configurable storage, such as a numpy array view, it would simply define `_get_storage_()` as a normal method, and then construction would become required. (In this case, the storage likely would be hidden from the user, since it would be part of some `from_numpy` type function).

henryiii:
Let's merge, we can tweak if needed, but merging helps the view work go forward.

PR #194, opened 2019-11-06
Direct memory access to accumulators

### LINKED ISSUE #21: h.view() for all storage types

### LINKED ISSUE #175: Retrieving results of non-trivial storages

### LINKED ISSUE #195: Access single histogram values (unlimited and atomic_int)
While you can get a .view, you can't actually access or set single values! Probably just a simple conversation registration.

henryiii:
This allows `.view()` to work on non-simple storages. Closes #175 and closes #21.

Features:

* No copy for asarray or view, fixes no-copy views
* Smart array returned, allowing `.view().value` to get the value, for example.
* Accessing a single value from the returned array gives an accumulator
* Single element access now fixed and tested for unlimited and atomic int histograms. Closes #195.

Todo:

- [x] Remove trailing `_`
- [x] Look into making view be no copy too (existing problem)
- [x] Add tests
- [x] Investigating setting elements of the array from instances

Not going to be ready by the CHEP talk, but should be ready for a release end of week / next week.


henryiii:
I have at least fully demonstrated that it can be done before the CHEP talk.

HDembinski:
I am actually very interested in this, because I want to use a weighted histogram in my current data analysis :).

henryiii:
It's progressing fairly well, will update the PR fairly soon with work from the plane.

henryiii:
I realized after PyHEP that the accumulators are something special to boost-histogram for Python users; up until now "double" storage is about the limit of what you could do in Python histograms easily. I think this explains why people are multiple people have run into this being broken - it was the very thing they were excited to have available directly in Python.

@HDembinski, this is ready for review. I'm just adding a few tests. Basic idea: `np.asarray(h)` simply returns a structured array with an exact, no copy view of the data (it can't return a ndarray subclass, AFAICT). The names exposed here have a naming convention mentioned in the docs, but to repeat it here: if a public property exposes a value, that's the name in the record array, otherwise it uses the private name without the trailing underscore.

`.view()` is a bit more powerful; we can return ndarray subclasses here, so we add a few useful properties and slightly nicer reprs. So you can even do:

```python
h.view().variance
```

to get the variance array, even if it is a computed property! If you do not access computed properties, this procedure is *still no copy*! And, unlike np.RecordArray, you don't even pay a lookup penalty when using properties and tab-completion works.

* Also fixed no-copy views from `.view()` for all arrays now, not just the fancy ones.

No copy example:

```python
>>> h = bh.histogram((10,0,1), storage=bh.storage.Weight())
>>> v = h.view()
>>> v.variance
array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])
>>> v.variance[3] = 3 # Only works if variance is not computed (WeightedSum)
>>> h
histogram(
  regular(10, 0, 1),
  storage=weight
) # Sum: weighted_sum(value=0, variance=3)
>>> h.view()
WeightedSumView(
      [(0., 0.), (0., 0.), (0., 0.), (0., 3.), (0., 0.), (0., 0.),
       (0., 0.), (0., 0.), (0., 0.), (0., 0.)],
      dtype=[('value', '<f8'), ('variance', '<f8')])
>>> v[3]
weighted_sum(value=0, variance=3)
```

I don't know about you, but I think this is really cool. :)

henryiii:
One design question: Should `np.asarray(h)` return the whole structured array, or just the most reasonable single value? So `.value`, generally. Since most places a user would actually trigger "asarray" are really when handing a histogram to a function that expects an array, and none of those functions expect a record array, in general, it might make more sense to make something more standard. We already make the assumption that asarray does not include the flow bins. @jpivarski might have an opinion here too (I think he leaned toward record array).

Recap of reason one might want just the single value:

Nobody should be doing this:
```python
v = np.asarray(h) # v = h.view() is better
```
but they may do this:
```python
plt.bar(h.axes[0].centers, h, width=h.axes[0].widths);
```

The latter expects an array of values.

jpivarski:
I was leaning toward record array because I thought the underlying structure was non-negotiably a `struct`, and this is the appropriate Python translation. If `dtype=[('value', '<f8'), ('variance', '<f8')]`, it's easy enough to `np.asarray(h)['value']` or `np.asarray(h)['variance']` to get arrays of values.

"Easy enough" in a program, but not necessarily in quick-and-dirty data analysis. For that, you're going to want something to return just the values or just the variances, either by changing the ABI in C++ (which I think is rather drastic) or by providing methods. I don't think you need to worry about users expecting `np.asarray(h)` to be some simple array of numbers—they know that the histogram has complex structure—I don't think you have to develop defensively for that. I'd say, just provide methods.

The histogram doesn't have to be a single `py::buffer_protocol`. You could make it a non-buffer object with two methods, `values` and `variances`, each of which returns a `py::buffer_protocol` with the appropriate stride. This way, you wouldn't have to change the ABI or present a record array; you could present two numerical arrays that just happen to interleave in memory. Or you could make the histogram a record array buffer and provide methods that each present a non-record view of `values` and `variances`. In fact, these could be implemented on the pybind11 side by casting a histogram as a `py::array` (no `_t`) and calling `attr("__getitem__")("values")` on it.

henryiii:
All histograms provide identical APIs; you would (and can) use `h.view()["value"]` or `h.view().value` to get the value array. That's already impemented. The question is, what should the raw histogram act like in Python to something that expects a numpy array? We already set the offsets and strides to give the `flow=False` view of the histogram, so `np.asarray(h)` is not a raw view over all the data, but rather tries to return what a user would expect. It's really a shortcut, then. We could even drop the `__array__` method and then force `.view()` for users who want the data.

Choices:

1. Drop `__array__`, and force `.view()`
2. Extend the shortcut to always return the most reasonable value if possible (implemented in an earlier commit)
3. Allow the shortcut to return structured arrays (implemented in the current commit). Then `np.asarray(h)["value"]` works  and `np.asarray(h).value` doesn't.

I'll let @HDembinski pick one. I'm slightly biased toward 2, as a user, I think it's the one I would expect, and extends the current idea of providing the most likely portion of the data instead of the complete data. I don't have a strong opinion, though. Technically, we wrap the pybind object in Python, and now that we have fixed `.view()`, the Pybind buffer protocol is rather irrelevant, actually. We could implement it in Python with `.view()` from Pybind object.

henryiii:
PS: In hist, we may have histogram subclasses for each storage, so `.values` might be a usable shortcut there. But boost-histogram (and Boost.Histogram) have a single Histogram class.

henryiii:
(Added an option an reordered)

Also, Jim, the description of how we could do it is pretty much how I'm currently doing it, other than being slightly further down in the wrapping.

Note for future posterity: Pybind11 is missing an `array` constructor that takes a buffer info object *and* a base ownership pointer; this was throwing me off until someone on the pybind11 gitter just suggested using the more verbose constructor that did allow the base pointer, which fixed the problem and allowed the array to be made tied to the reference counting of the histogram and without a copy. We should make a Pybind PR with the missing argument in the constructor.

jpivarski:
Agh—that's what you get for asking me for an opinion about something I haven't been following. You invited me to make some stuff up. What you're doing is probably fine.

henryiii:
A nice argument toward \#1 is that all behavior directly on the histogram is identical - if `plt.bar(h.axes[0].centers, h, width=h.axes[0].widths);` works, it works on all storages. Asking for `.view()` gives you the underlying data in storage, so that may not behave identically.

An argument toward \#2 is that this is just a shortcut and can be made explicit by using `h.view().value` (things after `.view()` are storage dependent, so it will depend on the storage).

I'm neutral here. Hans can pick any of the 4.

HDembinski:
Since I need weights support for my analysis, I implemented this in a branch, see https://github.com/scikit-hep/boost-histogram/tree/weights. This is basically a simpler version of what you are doing here. Maybe you could a quick look. I will not make a PR from this branch, just wanted to submit it to show you.

I think the fields for the internal boost-histogram accumulators can be all public (as was done in my branch). This simplifies the code a bit.

HDembinski:
For me it is not a question about convenience, but about the expectation what `np.asarray` does. My expectation is that this thing gives you all the information. You are right that we already hide the flow bins by default and that was a tricky design decision. The argument for hiding the flow bins by default was that they are almost an implementation detail. We need them to have proper projections, but people should normally not mess with them.

So from that point of view, I think we made the right decision to hide the flow bins in `np.asarray`. Nevertheless, when you have an accumulator that is really a struct, that should show up when you do `np.asarray`.

HDembinski:
So in summary, I would prefer option 4.

henryiii:
> I think the fields for the internal boost-histogram accumulators can be all public

That's what was done here.

> So in summary, I would prefer option 4, it seems the least surprising.

Okay, that is what is currently done here already.

Note this now fixes #195 too, you can now access and set single values from unlimited and atomic storages.

henryiii:
If this passes tests, I'm ready to merge. @HDembinski, the only difference I see in your branch is you replaced the property access to value, etc. I didn't make that change just to keep it closer to Boost.Histogram, this way the internal names and property access matches Boost.Histogram, so we can more easily make a matching change if Boost.Histogram changes. Otherwise it's the same. We can always do that change later, there's no external effect.

tests/test_public_hist.py
```diff
@@ -136,7 +136,7 @@ def test_fill_1d(flow):
         assert get(h, bh.overflow) == 1
 
 
-# TODO: atomic_int not supported
+# TODO: atomic_int, unlimited not supported
```
henryiii: Oops, this is supported now.

henryiii:
Extra note: Ahh, this would simplify the numpy descriptors, too; it might be a reasonable thing to do to clean up eventually. I didn't rename the internal fields because they would collide with the methods, but you are just binding directly to the field, so the method is not needed. Possible cleanup for a later PR.

boost_histogram/_internal/hist.py
```diff
@@ -406,5 +405,5 @@ def __getitem__(self, index):
             )
 
     def __setitem__(self, index, value):
-        indexes = _compute_commonindex(self._hist, index, expand=True)
+        indexes = _compute_commonindex(self._hist, index, value)
```
henryiii: Note to self: See why this changed...

henryiii: This should be False for now. Later, when we support setting with arrays, this will be True again.

It never should be "value"...

include/boost/histogram/python/accumulators_ostream.hpp
```diff
@@ -8,6 +8,10 @@
```
HDembinski: in boost.histogram the  ostream operators must be in a separate header, but here you could just move the code to the corresponding accumulator. it is more obvious this way.

tests/test_storage.py
```diff
@@ -0,0 +1,103 @@
...
+from numpy.testing import assert_array_equal
+
+
+@pytest.mark.parametrize(
+    "storage",
+    [bh.storage.Int, bh.storage.Double, bh.storage.AtomicInt, bh.storage.Unlimited],
+)
+def test_setting(storage):
+    h = bh.Histogram(bh.axis.Regular(10, 0, 1), storage=storage())
+    h[bh.underflow] = 1
```
HDembinski: this looks like you are testing both uhi and the storage. i think it would be better to not also use uhi here.

tests/test_storage.py
```diff
@@ -0,0 +1,103 @@
...
+    assert a[0] == h[0]
+
+    b = np.asarray(h)
+    assert b["value"][0] == h[0].value
+
+    h[0] = bh.accumulators.Mean(count=6, value=3, variance=2)
+
+    assert a[0] == h[0]
+    assert b["value"][0] == h[0].value
+
```
HDembinski: please also test variance and count.

tests/test_storage.py
```diff
@@ -0,0 +1,103 @@
...
+    a = h.view()
+
+    assert a[0] == h[0]
+
+    b = np.asarray(h)
+    assert b["value"][0] == h[0].value
+
+    h[0] = bh.accumulators.WeightedMean(wsum=6, wsum2=12, value=3, variance=2)
+
```
HDembinski: likewise, the full interface should be tested

boost_histogram/_internal/hist.py
HDembinski: Mostly, I was puzzled by the name of the function and I didn't look it up. I think this should be combined with `_at_set`, to make it faster. But that is another PR.

HDembinski: Allocating a temporary array for the linearized indices is not necessary and costs significant performance.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 #### User changes
 
 * Histogram and Axis classes now follow PEP 8 naming scheme (`histogram`->`Histogram`, `regular`->`Regular`, etc.) [#192][]
+* You can now view a histogram with accumulators, with property access such as `h.view().value` [#194][]
 * Added axes transforms [#192][]
 * You can now sum over a range with endpoints [#185][]
 * You can now access the functional regular axis directly, `regular_sqrt` becomes `regular.sqrt`, etc.. [#183][]
```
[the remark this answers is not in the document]
henryiii: This already has gone away in #200 as part of the cleanup.

boost_histogram/_internal/hist.py
```diff
@@ -406,5 +405,5 @@ def __getitem__(self, index):
             )
 
     def __setitem__(self, index, value):
-        indexes = _compute_commonindex(self._hist, index, expand=True)
+        indexes = _compute_commonindex(self._hist, index, expand=False)
```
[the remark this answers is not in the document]
henryiii: Expand expands `...`. It may go away or change drastically when we add proper set support next; but for now, you can't expand when you only support setting a single value at a time. :)

CHANGELOG.md
```diff
@@ -12,6 +13,10 @@
 * Signatures are much nicer in Python 3 [#188][]
 
 
+#### Bug fixes:
+* Unlimited and AtomicInt storages now allow single item access [#194][]
+* `.view()` now no longer makes a copy [#194][]
```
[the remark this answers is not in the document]
henryiii: Given version 0.5.x claimed all more complicated storages had to be accessed per element, and per element access was broken on these two, I'd say it was a bug. Also breaks up the changelog a bit better. And `.view` was really supposed to not copy.

include/boost/histogram/python/accumulators/weighted_sum.hpp
```diff
@@ -0,0 +1,120 @@
...
+    template <typename T>
+    bool operator!=(const T &rhs) const noexcept {
+        return !operator==(rhs);
+    }
+
+    /// Return value of the sum.
+    const RealType &value() const noexcept { return sum_of_weights_; }
+
+    /// Return estimated variance of the sum.
+    const RealType &variance() const noexcept { return sum_of_weights_squared_; }
```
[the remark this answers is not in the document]
henryiii: If I go ahead and clean up the other things, I probably can go ahead and tackle this too. I originally was trying to stay as close to the original code as possible.

boost_histogram/_internal/hist.py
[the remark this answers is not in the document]
henryiii: (Note to self: Given that neither Hans nor I knew what expand meant (and I wrote it), we should rename to `expand_ellipsis`)

CHANGELOG.md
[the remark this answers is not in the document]
henryiii: I think these shortcuts don't need to go in (yet), and would more likely be part of Hist anyway. Let's look at my proposal for transformations and then think about shortcuts later.

This went in before you had the idea to use `transform=`, and I hadn't cleaned it up yet (until #200, anyway)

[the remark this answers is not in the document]
HDembinski: Ok, that's all good then. I am not super against this, but I would really like the shortcuts to come in later, after our 1.0 release.

CHANGELOG.md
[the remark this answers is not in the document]
HDembinski: OK

boost_histogram/_internal/hist.py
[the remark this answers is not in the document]
HDembinski: OK

CHANGELOG.md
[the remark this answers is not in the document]
henryiii: This is one of my points in #204 :) For transforms though, I'm not hating them as much as I did. I'm going to write up the idea.

include/boost/histogram/python/accumulators/weighted_sum.hpp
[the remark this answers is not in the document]
HDembinski: That's very nice, thank you. But here you can change things at your leisure, since it is all private implementation.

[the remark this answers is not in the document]
HDembinski: And if we can make our lives easier...

include/boost/histogram/python/accumulators/weighted_sum.hpp
```diff
@@ -0,0 +1,120 @@
...
+};
+
+} // namespace python
+} // namespace histogram
+} // namespace boost
+
+#ifndef BOOST_HISTOGRAM_DOXYGEN_INVOKED
+namespace std {
+template <class T, class U>
+struct common_type<boost::histogram::accumulators::weighted_sum<T>,
```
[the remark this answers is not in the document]
HDembinski: to clarify, strictly speaking these are wrong now, because you want common_type specializations for your versions of weighted_sum, which these here are common_type specializations for the original boost::histogram accumulators. but never mind all this, just remove them :)

CHANGELOG.md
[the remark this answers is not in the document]
henryiii: We can also experiment with shortcuts in Hist, then back port the best ideas to boost-histogram. Boost-histogram is supposed to be the more stable, base library, after all. ;)

boost_histogram/_internal/hist.py
henryiii: First we should make everything work, then we can optimize. :)

include/boost/histogram/python/accumulators/mean.hpp
```diff
@@ -0,0 +1,138 @@
...
+namespace python {
+
+/** Calculates mean and variance of sample.
+
+  Uses Welfords's incremental algorithm to improve the numerical
+  stability of mean and variance computation.
+*/
+template <class RealType>
+struct mean {
+  public:
```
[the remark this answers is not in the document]
henryiii: Obviously, I changed things in stages. :)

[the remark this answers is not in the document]
henryiii: Oh, that's actually on the methods, I missed those.

include/boost/histogram/python/accumulators_ostream.hpp
henryiii: Ours are customized anyway, so we can. Though just moving it to `python/accumulators/ostream.hpp` would help (and sum's ostream is in there too).

include/boost/histogram/python/accumulators/mean.hpp
```diff
@@ -0,0 +1,138 @@
...
+        return sum_ == rhs.sum_ && mean_ == rhs.mean_
+               && sum_of_deltas_squared_ == rhs.sum_of_deltas_squared_;
+    }
+
+    template <class T>
+    bool operator!=(const mean<T> &rhs) const noexcept {
+        return !operator==(rhs);
+    }
+
+    const RealType &count() const noexcept { return sum_; }
```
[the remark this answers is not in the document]
henryiii: Might as well do it now since I'm changing things anyway.

henryiii:
The accumulator interface is now (not yet pushed) nearly identical to the view interface, which is great. One more thing: should accumulators support `a["value"]` as well? Then `a[x]["value"]` would work regardless of whether x is a purely values or if it contains one or more slices. Also, the order becomes interchangeable, that is, `arr["value"][3] == arr[3]["value"]`, regardless of whether arr came from `.view()` or `np.asarray`. Currently, without the getitem, it does not work on the `.view()` version and does work on the `np.asarray` version.

henryiii:
Currently, we have keyword arguments in the accumulator constructor:

```python
bh.accumulators.WeightedMean(wsum=6, wsum2=12, value=3, variance=2)
```

Should we use the property/field names? This makes the repr *very* long:

```python
bh.accumulators.WeightedMean(sum_of_weights=6, sum_of_weights_squared=12, value=3, variance=2)
```

But I like the consistency.

henryiii:
The final commit adds `accum["value"]`, and allows IPython completion on the name. This keeps the customized view (which returns accumulators instead of structured array "scalars" on single element access) from breaking code that expects the structured array scalars. I can revert the commit if this is a bad idea.

Ready for second review, @HDembinski. This is looking really nice. :)

include/boost/histogram/python/accumulators_ostream.hpp
henryiii: I moved it, and it seems like it's reasonably ordered, up to you if you want it moved further.

include/boost/histogram/python/accumulators/mean.hpp
```diff
@@ -0,0 +1,96 @@
...
+    template <class T>
+    bool operator!=(const mean<T> &rhs) const noexcept {
+        return !operator==(rhs);
+    }
+
+    RealType variance() const noexcept { return sum_of_deltas_squared / (count - 1); }
+
+    template <class Archive>
+    void serialize(Archive &ar, unsigned) {
+        ar &boost::make_nvp("count", count);
```
henryiii: I've changed the archive names to match the internal names.

HDembinski:
> But I like the consistency.

I also like consistency more than the short names. "Readability counts"

include/boost/histogram/python/accumulators/mean.hpp
HDembinski: That's fine. It may become out of sync eventually, but since we have a fresh start here, we can do it right.

include/boost/histogram/python/accumulators/ostream.hpp
```diff
@@ -8,21 +8,23 @@
...
+
 #include <boost/histogram/detail/counting_streambuf.hpp>
 #include <boost/histogram/fwd.hpp>
 #include <iosfwd>
 
-namespace boost {
-namespace histogram {
-
-namespace detail {
+namespace {
```
HDembinski: Anonymous namespace may not be used in headers, see https://wiki.sei.cmu.edu/confluence/display/cplusplus/DCL59-CPP.+Do+not+define+an+unnamed+namespace+in+a+header+file

HDembinski: Strictly speaking it is ok if the header is not included in several cxx files, but this is the case here, I think. Also, it is good to embrace such rules instead of thinking about them case by case...

include/boost/histogram/python/accumulators/ostream.hpp
```diff
@@ -38,51 +40,55 @@ handle_nonzero_width(std::basic_ostream<CharT, Traits> &os, const T &x) {
     return os;
 }
 
-} // namespace detail
+} // anonymous namespace
```
HDembinski: The reason why boost libs use `detail` so much is because unnamed namespaces cannot be used, see previous comment.

HDembinski:
> The accumulator interface is now (not yet pushed) nearly identical to the view interface, which is great. One more thing: should accumulators support `a["value"]` as well? Then `a[x]["value"]` would work regardless of whether x is a purely values or if it contains one or more slices. Also, the order becomes interchangeable, that is, `arr["value"][3] == arr[3]["value"]`, regardless of whether arr came from `.view()` or `np.asarray`. Currently, without the getitem, it does not work on the `.view()` version and does work on the `np.asarray` version.

That sounds excellent, yes.

> The final commit adds `accum["value"]`, and allows IPython completion on the name. This keeps the customized view (which returns accumulators instead of structured array "scalars" on single element access) from breaking code that expects the structured array scalars. I can revert the commit if this is a bad idea.

This is very nice!

include/boost/histogram/python/accumulators/ostream.hpp
henryiii: > Also, it is good to embrace such rules instead of thinking about them case by case.

Sounds good.

henryiii:
Not worried about typing it so much (IPython tab completes it for me anyway), but it is long enough that it can go off the screen when displayed (such as in the GitHub code box above) Still better to be consistent.

henryiii:
Should be ready to go! I've removed the namespace, fixed the spelling, and replaced the keywords.

HDembinski:
Then it is ready to merge! Excellent work.

PR #196, opened 2019-11-07
Warning for unintialized storage

henryiii:
This adds a warning for all uninitialized storages. If a user forgets the (), a warning is raised. In 0.7.0, this will become an error with an explicit, helpful error message.

@HDembinski, I implore you to look over the reasoning given by @jpivarski and me and carefully consider allowing default-constructible storages to be passed in without an error if a user forgets the `()`, either by mistake or by thinking they should pass a type. The function signature would be `storage : Union[Storage, Type[Storage]]` in that case. I will respect your wishes if you do not change your mind.

---

PS: This is the little program you can run through MyPy to verify that the types pass:

```python
from typing import Union, Callable, Type

class Storage: pass

def f(storage : Union[Storage, Type[Storage]]):
    pass

# Also passes with storage : Union[Storage, Callable[[], Storage]]

f(Storage())
f(Storage)
```

```
$ mypy tmp.py
Success: no issues found in 1 source file
```

### DIFF
--- a/boost_histogram/_internal/hist.py
+++ b/boost_histogram/_internal/hist.py
@@ -6,7 +6,7 @@ from .. import _core
 from .axis import _to_axis, Axis
 from .axistuple import AxesTuple
 from .sig_tools import inject_signature
-from .storage import Double, _to_storage
+from .storage import Storage, Double, _to_storage
 
 import warnings
 import numpy as np
@@ -98,7 +98,7 @@ class BaseHistogram(object):
         ----------
         *args : Axis
             Provide 1 or more axis instances.
-        storage : Storage = bh.storage.Double
+        storage : Storage = bh.storage.Double()
             Select a storage to use in the histogram
         """
 
@@ -114,6 +114,16 @@ class BaseHistogram(object):
         # Keyword only trick (change when Python2 is dropped)
         with KWArgs(kwargs) as k:
             storage = k.optional("storage", Double())
+
+        # Check for missed parenthesis
+        if not isinstance(storage, Storage) and callable(storage):
+            storage = storage()
+            if not isinstance(storage, Storage):
+                raise KeyError("Only storages allowed in storage argument")
+            else:
+                warnings.warn(
+                    "Passing in an initialized storage may be removed in the next release. Please add ()."
+                )
 
         storage = storage._get_storage_()
 
--- a/boost_histogram/_internal/storage.py
+++ b/boost_histogram/_internal/storage.py
@@ -6,17 +6,19 @@ from .._core import storage as store
 
 
 class Storage(object):
-    __slots__ = ()
+    __slots__ = "_storage"
+
+    def __init__(self):
+        self._storage = self._STORAGE()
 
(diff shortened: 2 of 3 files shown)


HDembinski:
I think that functions which accept both types and values of that same type are not pythonic. Automatically turning a type into a default constructed value of that type is not pythonic. This is violating the principle of strong typing in Python.

The motivation for this whole debate was to skip a brace for commonly used storages, https://github.com/scikit-hep/boost-histogram/pull/193 offers this feature in a simple and straight-forward way without violating strong typing. The syntax shown at PyHEP remains valid and gets promoted as officially supported by the module.

henryiii:
This is your solution. The one you proposed. I want to have a 1 release phase out, so we respect the users who tried out Boost.Histogram at PyHEP, and help them transition. This is what I've been doing on all other features that have changed. (though this one is a UserWarning, rather than a DeprecationWarning to make extra sure that they see it)

henryiii:
If you *really* don't like the warning for one release, we can issue a clearly worded error immediately.

HDembinski:
I am trying to put your comments together.
* you don't want official support for `h = bh.histogram(..., storage=bh.storage.mean)` anymore
* DepStorageMixin is temporary and will be removed eventually
* A warning will be shown when you try to pass a storage type instead of a value

I am ok with this if
* `h = bh.histogram(..., storage=bh.storage.mean)` is a warning
* `h = bh.histogram(..., storage=bh.storage.Mean)` is an error

HDembinski:
On the other hand, I grew attached to the idea of having `bh.storage.mean`. It is useful and the clutter to tab completion is acceptable.

henryiii:
I do not like two ways to do the same thing. Letting a user forget `()` was being nice to the users. Having two ways to do this is not consistent Boost.Histogram, you are not going to add static instances or something like that. Adding `int` and `Int` is confusing. If we hide Int by leaving it only internal or putting it only in the cpp module, I'm okay with the lower case instances.

StorageDepMixin is temporary and will be removed. `h = bh.histogram(..., storage=bh.storage.Mean)` can become an error.

Why are there merge conficts?

henryiii:
> On the other hand, I grew attached to the idea of having bh.storage.mean. It is useful and the clutter to tab completion is acceptable.

You are constantly changing. I agreed to your plan, and you agreed to in the comment (see above), then you decided to merged your change anyway, which breaks the system I had in place for adding static typing later. There now is now Strorage base class. Either the reprs in cpp or the one in the core class will now be wrong, since there's no wrapping. Why don't you be honest, we have been arguing and therefore you don't want anything I put in, you have to "Win".

**I have *never* merged anything against your will.** You now cannot say that.

henryiii:
I proposed that I could write Boost.Histogram wrappers, and you agreed. They were written completely from scratch (except for the two test files, `test_public_hist`). I have written 80% of the bindings. For anything in the main portion, we have to come to a mutual agreement on the change before we merge. You have complete freedom in `.cpp`.

You deserve to give me that much respect.

PR #221, opened 2019-11-18
Str/Int Categories

### LINKED ISSUE #215: Empty growable categories
Currenty, we cannot make an empty growable category axis. The problem is that we can't know if an empty list is supposed to have strings or ints. Here are some ideas:

```python
bh.Category(str, growth=True)
bh.Category(int, growth=True)
```

Passing a type selects that type of axes but does not construct any str's or int's.

---

```python
bh.Category([], growth=str)
bh.Category([], growth=int)
```

Only growing axes have this issue, and str and int both ducktype as true anyway. Looks weirder than the first one.

---

```python
bh.Category([], dtype=str, growth=True)
bh.Category([], dtype=int, growth=True)
```

Nothing else has a dtype, but this does look rather normal.

---

```python
bh.Category[str]([], growth=True)
bh.Category[int]([], growth=True)
```

Let's be Cython, shall we?

---

```python
bh.StrCategory([], growth=True)
bh.IntCategory([], growth=True)
```

We give up on the single Category, and have two.

In any of these last ones, the `[]` could be optional.

@HDembinski ?

---

We need to make sure IntEnum is well supported, too; it might be nice to allow something like:

```python
class Cat(IntEnum):
    a = 1
    b = 3
bh.Category(Cat)
```

which is kind of in line with the first one. (Currently, `bh.Category(list(Cat))` works and that might be just fine, just something to think about).

henryiii:
Fixes #215 by splitting Category into IntCategory and StrCategory. You can now make an empty category axis.

Some cleanup of the axis creation, and removing internal docstrings.

~~@HDembinski, would a possible workaround for now (in the bindings) be to create the default constructed axis, then set the metadata manually? Would there be any potential issues? It would be nice to be compatible with Boost 1.72's Histogram for the moment.~~ 

~~(Implemented)~~

@HDembinski already fixed this in Boost.Histogram, so now using that here!

### DIFF
--- a/boost_histogram/_internal/axis.py
+++ b/boost_histogram/_internal/axis.py
@@ -504,7 +504,11 @@ class BaseStrCategory(Axis):
             categories = list(categories)
 
         if options == {"growth"}:
-            self._ax = ca.category_str_growth(categories, metadata)
+            if len(categories) == 0:
+                self._ax = ca.category_str_growth()
+                self._ax.metadata = metadata
+            else:
+                self._ax = ca.category_str_growth(categories, metadata)
         elif options == set():
             self._ax = ca.category_str(categories, metadata)
         else:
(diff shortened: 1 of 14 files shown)


HDembinski:
If I understand correctly, this removes Category completely and leaves only IntCategory and StrCategory. I think that's fine for now.

HDembinski:
Looks good to me.

PR #223, opened 2019-11-18
renaming tests

HDembinski:
This improves the naming of tests, at least for me. So far, the filenames do not follow any pattern. The filename of a test should indicate what this is about. It should also be clear whether a test checks public or private interface. Tests of internal interfaces should have a `_internal_` prefix.


henryiii:
Any cleanup of tests is welcome! They were grown organically, unfortunately.

HDembinski:
> Any cleanup of tests is welcome! They were grown organically, unfortunately.

Excellent, my latent OCD is triggered by the current organic entropy...

henryiii:
Note to self or whoever ends up merging: As with any PR that renames files, it should be rebased and merged instead of squashed and merged, so git does not lose track of history. Renames should always be in their own commit.

HDembinski:
> Note to self or whoever ends up merging: As with any PR that renames files, it should be rebased and merged instead of squashed and merged, so git does not lose track of history. Renames should always be in their own commit.

Good point. I removed the commits which do something other than renaming.

PR #224, opened 2019-11-18
Restructuring tests

HDembinski:
**pending on test renaming**

this moves some test code around to other places, the idea is to have all pickle tests in one place and all init tests, etc. there is still some issue with the pickle test to be fixed. seems i cannot just use dumps and loads on py27


henryiii:
Pybind11 does not support pickle protocol 0 (plain text), which is the default protocol in Python 2.

HDembinski:
I will redo this some other time.

PR #229, opened 2019-11-19
Build updates

### LINKED ISSUE #222: Remove test_basic_import.py
Do we really need `test_basic_import.py`? What does an assert of a module do? It always passes, I would think. The name suggests that this test checks whether boost.histogram can be imported, but basically all other tests test this as well. I think this is not useful.

The other test checks whether the version info is the same in boost_histogram and  boost_histogram.version, which is automatically true since it is imported. So, all in all, I think this test can be removed.

### LINKED ISSUE #225: cmake should default to debug build
cmake defaults to release, but cmake builds are only done by developers. doing a debug build by default seems more natural.

### LINKED ISSUE #226: reduce binary size
setup.py builds with `-g`. We should be able to greatly reduce the binary size by either hacking into distutils so that `-g` is not set or by stripping the final shared library
https://linux.die.net/man/1/strip

henryiii:
Closes #225 and closes #226 and closes #222.


setup.py
```diff
@@ -108,7 +108,7 @@ def cpp_flag(compiler):
 class BuildExt(build_ext):
     """A custom build extension for adding compiler-specific options."""
 
-    c_opts = {"msvc": ["/EHsc", "/bigobj"], "unix": []}
+    c_opts = {"msvc": ["/EHsc", "/bigobj"], "unix": ["-g0"]}
```
HDembinski: I didn't know that you can set that flag. It is not documented in Apple Clang.

henryiii:
This drops the manylinux wheel size down to about 1-2 MB from 20 MB. Windows and macOS already were about 1-2 MB. Nice!

setup.py
henryiii: Clang tends to allow GCC opts. This fixed the issue on manylinux, and macOS was already smaller.

Issue #214, opened 2019-11-14
Miscellaneous usability issues


nsmith-:
Even the best laid plans won't survive first contact with the user.  Let me know if you want these broken into separate issues, but most are small.  p.s. this is with v0.5.2

I installed the package and played around with it, first reaction: docstrings are quite sparse, probably due to the large amount of pybind11 shenanigans.
It would be nice if the various axis type inheritances are visible from python, or at least something that tells me `isinstance(x, bh.axis.axis)` or something to that effect.

### 1 (see #216)

First dumb move:
```
ax = bh.axis.regular(20, 0, 1)
ax.value()
```
raises a cryptic `TypeError`, but I just didn't know the function required arguments.

### 2 (see #215)

For growable categories, I hit several obstacles:
```
ax = bh.axis.category([], growth=True)
```
raises `ValueError` saying there must be at least one bin.
Is it not possible to create growable axis with empty starting categories?

### 3
Answered well by https://github.com/scikit-hep/boost-histogram/issues/214#issuecomment-554683890

Playing with indexing,
```
ax = bh.axis.category(['old'], growth=True)
ax.index('new')
# returns 1
ax.extent
# returns 1, why?
```
so I assume we should only treat axis objects as immutable with no actual state w.r.t. filling?
Of course, if I make a histogram using this axis, everything works as expected:
```
h = bh.histogram(bh.axis.category(['old'], growth=True))
h.fill(['hi', 'there'])
h.axis(0).extent
# returns 3
```

### 4 (see #184)

The options for how to provide indices to fill are a bit weird:
```
h = bh.histogram(bh.axis.category(['old'], growth=True))
h.fill('hi')
```
raises `ValueError` about casting strings to `char const&`, so I cannot use plain strings?

### 4.a (#230)

I see the choice to interpret python strings as iterables causes some headache:
```
h = bh.histogram(bh.axis.category([''], growth=True), bh.axis.regular(20, 0, 1))
h.fill('hi', np.arange(4))
```
raises `ValueError: spans must have compatible lengths`
(although even when they are compatible, the string to char array issue comes up again)

### 5 #233

Ok, so sticking to arrays,
```
h = bh.histogram(bh.axis.category([''], growth=True), bh.axis.regular(20, 0, 1))
h.fill(['hi'], np.arange(4))
Segmentation fault: 11
```
uh oh!
but this works ok:
```
[...]

PR #230, opened 2019-11-19
Fix fill issues

HDembinski:
Fixes the issues in #214 reported by @nsmith-
- [x] `h.fill("foo")` must work
- [x] `h.fill(np.array("foo"))` should not work and give a clear error message (ndim==0 not supported)
- [ ] segfault of the following code
```py
h = bh.Histogram(bh.axis.StrCategory([], growth=True), bh.axis.Integer(0, 1))
h.fill(["a"], [1, 2])
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,6 @@
 *~
 *.clangd
 compile_commands.json
+
+# Venvs
+/*venv*
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,7 +93,7 @@ option(BOOST_HISTOGRAM_ERRORS "Make warnings errors (for CI mostly)")
 
 # Adding warnings
 if("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang" OR "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU")
-    target_compile_options(_core PRIVATE -Wall -Wextra -pedantic-errors -Wconversion -Wsign-conversion)
+    target_compile_options(_core PRIVATE -Wall -Wextra -pedantic-errors -Wconversion -Wsign-conversion -Wsign-compare)
     if(BOOST_HISTOGRAM_ERRORS)
         target_compile_options(_core PRIVATE -Werror)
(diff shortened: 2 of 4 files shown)


HDembinski:
Notes:
- segfault only:
  - when StrCategory (other axes work)
  - growth is on
  - argument for the category axis has a single entry
- bugs get hot-fixes for now, but the underlying issues in boost::histogram need to be fixed before the 1.72-beta closes

HDembinski:
Henry if you want, you can merge this now and I make another PR for the segfault. I think I need to fix this in boost::histogram, there is nothing I can do on this level.

henryiii:
Why is the gil release removed?

HDembinski:
> Why is the gil release removed?

It still had my comment whether this is safe or not without an answer. So I was thinking about it again. If another thread resizes the array while we are processing it, the code will segfault. I think you argued the last time we discussed this, that this is acceptable - users should not modify an array in one thread and read it in another thread without a mutex.

HDembinski:
I can revert this, but we need a proper comment otherwise I will stumble over this again.

henryiii:
This is the same behavior as numpy. If you do not release the GIL, the Python interpreter locks up and will do *nothing* else; it can't even context-switch because it can't pause the call like it can with a series of Python commands. You should never make a potentially long-running call that holds the GIL.

A proper comment would be helpful.

henryiii:
And users can always use shared memory and multithreading to get around a GIL lock. It's not a promise of thread safety. It's just a workaround to allow Python reference counting. We should not be changing that in the fill function (and I don't think we are).

HDembinski:
Ah, so we should only release the gil when we do not manipulate the ref counts anymore, correct?

henryiii:
Yes. If we change a refcount with the GIL released, that's a bug.

HDembinski:
feel free to merge, i am not sure how fast i can provide the segfault fix and it makes no sense for this pr to wait

henryiii:
Looks great, thanks!

PR #231, opened 2019-11-20
Allow callback in Transform

### LINKED ISSUE #191: Add circular axis types
We can add a circular varaible axis and circular integer axis. Should be done in conjunction with #211, since this will break the variant serialization (unless added to the end, but I'd like one chance to nicely order it).

### LINKED ISSUE #211: Transforms revamp
Here's my proposal for transforms:

1. We add `CTransform(cpp_forward, cpp_reverse)`, and eventually a pure python version, that take a compiled function pair and Python function piar, respectively. (I've not done this in Pybind11 before, but it looks pretty easy, and you have experience with it in iMinuit that I might call on ;) )
2. We use this to implement `log = CTransform(_log_fn, _exp_fn)`, where we compile these functions internally.
3. We remove `Sqrt()`. We set `sqrt = Pow(0.5)`.

Now, we have no name duplication based on capital letters, we have clear distinctions, and we have more functionality, and simpler C++ axes.

Let's *try* to get this into 0.6.0, since it will be a breaking change to serialization. Not worth holding up the release too long for, though.

henryiii:
This adds a `Function(forward, inverse)` transform that takes CTypes function pointers and performs the transform without touching Python - this includes numba cfunc's.

```python
import numba, math, numpy as np, ctypes

# Pure Python (15x slower than Log())
ftype = ctypes.CFUNCTYPE(ctypes.c_double, ctypes.c_double)
bh.axis.Regular(10, 1, 4, transform=bh.axis.transform.Function(math.log, math.exp, convert=ftype))

# Pure Python: Numpy (90x slower than Log())
bh.axis.Regular(10, 1, 4, transform= bh.axis.transform.Function(np.log, np.exp, convert=ftype))

# C++ (7% slower than Log())
bh.axis.Regular(10, 1, 4, transform=bh.axis.transform.log))

# Numba (7% slower than Log())
@numba.cfunc(numba.float64(numba.float64,))
def exp(x):
    return math.exp(x)

@numba.cfunc(numba.float64(numba.float64,))
def log(x):
    return math.log(x)

bh.axis.Regular(10, 1, 4, transform=bh.axis.transform.Function(log, exp))
```

Simple speed tests look like:

```python
rand = np.random.rand(10_000)*3+1
r = ...
%%timeit
r.index(rand)
```

Internally, `Function` takes a convert function, and stores `forward`, `inverse`, and `convert`. Then on creation or on unpicking it uses `convert` to convert `forward` and `inverse` to function pointers, which it can then call without Python. This allows pickling in the Pure Python case, and in the Numba case. Users provide custom functions to allow custom transforms, such as those loaded from `.so`'s, to be picklable, as well.

- [x] Fix for non Clang/Libc++ builds
- [x] Add tests
- [x] Add documentation
- [x] Work on public interface (and come up with good name for Func, warning or disable pure Python)
- [x] Full `ctypes` support (`cffi` and `PyCapsule` not supported for now)
- [x] Remove Log() specialization and Sqrt(), provide replacements
- [x] Reorder axes to put `regular_uoflow` first
- [x] Remove shortcut methods for now
- [x] Add circular axes for variable and integer, closes #191 

Would close #211.


include/boost/histogram/python/axis.hpp
```diff
@@ -243,6 +245,61 @@ py::array_t<double> widths(const A &ax) {
...
+
+    bool operator==(const func_transform &other) const noexcept {
+        return _forward_ob.equal(other._forward_ob)
+               && _inverse_ob.equal(other._inverse_ob);
+    }
+
+    template <class Archive>
+    void serialize(Archive &ar, unsigned /* version */) {
+        ar &boost::make_nvp("forward", _forward_ob);
+        ar &boost::make_nvp("reverse", _inverse_ob);
```
henryiii:
```suggestion
        ar &boost::make_nvp("inverse", _inverse_ob);
```

HDembinski:
I am super impressed by this. I think Func will be the main way of generating Transforms, so why don't we just call it "Transform"? It would be nice to hide the technical necessity to pass `lambda x: math.log(x)` etc in the implementation layer.

henryiii:
We could do that, but I don't want to confuse Transform the base class for Pow and Transform the flexible transform class. We could call it BaseTransform, though.

The issue with `lambda x: math.log(x)` being required might be fixable in Pybind11. It should not be trying to increase the reference count on a builtin.

I'm going to change the way I get the function pointer to be fully CTypes compatible; then we can take any CTypes callable. The other two supported by LowLevelCallable from SciPy include `cffi` and PyCapsule; `cffi` should be easy to support too, I think (have not used it much), and PyCapsule would be very hard to support (it supports "Functors", basically, which would allow `Pow` to be implemented, but I don't think that's all that important with something like Numba available and would greatly complicate the implementation, I believe).

> Note to self for later, since it's not obvious: The address of a CTypes callable `f.ctypes` is:
> 
> ```python
> ctypes.cast(f.ctypes, ctypes.c_void_p)
> ```
> 
> This avoids the Numba-specific `.address` attribute.

henryiii:
New version (in progress):

* GCC & MSVC support
* Faster function calls

```python
# Pure Python (cannot capsulate builtins) (15x slower than Log())
ftype = ctypes.CFUNCTYPE(ctypes.c_double, ctypes.c_double)
l = ftype(math.log)
e = ftype(math.exp)
bh.axis.Regular(10,1,4, transform=bh.axis.transform.Func(l, e))

# Pure Python: Numpy (90x slower than Log())
l = ftype(np.log)
e = ftype(np.exp)
bh.axis.Regular(10,1,4, transform=bh.axis.transform.Func(l, npe))

# C++ (7% slower than Log())
bh.axis.Regular(10,1,4, transform=bh.axis.transform.Func(bh._core.axis.transform._log_fn, bh._core.axis.transform._exp_fn))

# Numba (7% slower than Log())
@numba.cfunc(numba.float64(numba.float64,))
def exp(x):
    return math.exp(x)

@numba.cfunc(numba.float64(numba.float64,))
def log(x):
    return math.log(x)

bh.axis.Regular(10,1,4, transform=bh.axis.transform.Func(log, exp))
```

henryiii:
New(er) version:
* Custom name support
* Custom transform function support
    - Numba - automatically recognizes and compiles builtins, too!
    - Pure python
    - (INTERNAL) C++ functions from the internal module
    - (default) Use `.ctypes` if available
* Supports pickling when possible (using custom transform function) (Currently does not work when unpicked due to caching not being rerun)

# Pure Python (15x slower than Log())

bh.axis.Regular(10,1,4, transform=bh.axis.transform.Func(math.log, math.exp, convert=bh.axis.transform.convert.python))

# Pure Python: Numpy (90x slower than Log())
bh.axis.Regular(10,1,4, transform=bh.axis.transform.Func(np.log, np.exp, convert=bh.axis.transform.convert.python))

# C++ (7% slower than Log())
bh.axis.Regular(10,1,4, transform=bh.axis.transform.Func(bh._core.axis.transform._log_fn, bh._core.axis.transform._exp_fn))

# Numba (7% slower than Log())
@numba.cfunc(numba.float64(numba.float64,))
def exp(x):
    return math.exp(x)

@numba.cfunc(numba.float64(numba.float64,))
def log(x):
    return math.log(x)

bh.axis.Regular(10,1,4, transform=bh.axis.transform.Func(log, exp))

# Alternate, picklable numba
bh.axis.Regular(10,1,4, transform=bh.axis.transform.Func(math.log, math.exp, convert=bh.axis.transform.convert.numba))

henryiii:
Updated description with the latest version. Just needs some docs and a few axis changes.

henryiii:
Actually, I've been thinking about this and I think I've overstepped the bounds of boost-histogram. The boundary I've mentioned several times between boost-histogram and hist is that boost-histogram has no dependencies. If we allow a function or a class that soft-import numba why not add a plot function that soft imports matplotlib, etc? I think we should simply provide the ability to write the function (it's really trivial), and then let users write them. `.convert.*` would be removed, and the two public functions (`numba` and `python`) would become examples in the documentation. Downstream packages like Hist, Coffea, etc. could add their own numba or python converter function if they wanted to.

This is the entirety of the functions, minus the soft import:

```python
function_type = ctypes.CFUNCTYPE(ctypes.c_double, ctypes.c_double)
def python(value) :
    return function_type(value)

def numba(value):
    # Support math.sqrt and such
    if isinstance(value, types.BuiltinFunctionType):
        return numba.cfunc(numba.double(numba.double))(lambda x: value(x)).ctypes

    return numba.cfunc(numba.double(numba.double))(value).ctypes
```

The remaining part would be the current CTypes interface, and the settable conversion function.

The other possibility is that we could explicitly allow *soft* imports of numba and only numba in the scope of boost-histogram; as eventually we want to have a numba interface as well. In that case, the current design would be fine. @HDembinski, thoughts?

HDembinski:
The latest version looks very nice. The name is also good. Regarding overstepping, I would say the domain of boost-histogram is computation of the actual histogram, so it should focus on that, and we should not depend on more than numpy.

I think you are right, it sounds better to show users in the docs how make their numba-accelerated functions instead of integrating this. I am happy if `.convert` is removed.

henryiii:
@HDembinski, I'm ready to merge, does this look good?

PR #233, opened 2019-11-21
Fix fill segfaults and broadcasting misbehavior

HDembinski:
~Now the broadcasts work like in numpy~, passing arrays of different size is an error, but scalar and array works, the scalar is broadcast. This fixes the segfault of nsmith


HDembinski:
Numpy also broadcasts `foo([1], [1, 2])`, while this is now an error in `fill`, but could support this again with a change in the fill code in boost-histogram without touching boost.histogram.

PR #238, opened 2019-11-23
Stale axis fix

### LINKED ISSUE #220: Stale axes when growth=True and fill changes an axis
Currently, the axes seem to be copies. They should not be.

henryiii:
Needs tests, but I think I've fixed #220.

### DIFF
--- a/include/boost/histogram/python/register_histogram.hpp
+++ b/include/boost/histogram/python/register_histogram.hpp
@@ -171,16 +171,25 @@ register_histogram(py::module& m, const char* name, const char* desc) {
 
         .def(
             "axis",
-            [](const histogram_t& self, int i) {
+            [](const histogram_t& self, int i) -> py::object {
                 unsigned ii = i < 0 ? self.rank() - (unsigned)std::abs(i) : (unsigned)i;
-                if(ii < self.rank())
-                    return self.axis(ii);
+
+                if(ii < self.rank()) {
+                    const axis_variant& var = bh::unsafe_access::axis(self, ii);
+                    return bh::axis::visit(
+                        [](auto&& item) -> py::object {
+                            return py::cast(item, py::return_value_policy::reference);
+                        },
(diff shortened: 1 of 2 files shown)


include/boost/histogram/python/register_histogram.hpp
```diff
@@ -171,16 +171,25 @@ register_histogram(py::module& m, const char* name, const char* desc) {
...
         .def(
             "axis",
-            [](const histogram_t& self, int i) {
+            [](const histogram_t& self, int i) -> py::object {
                 unsigned ii = i < 0 ? self.rank() - (unsigned)std::abs(i) : (unsigned)i;
-                if(ii < self.rank())
-                    return self.axis(ii);
+
+                if(ii < self.rank()) {
+                    const axis_variant& var = bh::unsafe_access::axis(self, ii);
```
HDembinski: this is basically what i had in mind as well. only a minor comment: why use unsafe_access here instead of `self.axis(ii)`?

HDembinski: apart from that this looks very good

henryiii: Leftover from when I first tried to avoid const, since the py::cast is going to ignore the const. Fixed.

tests/test_histogram.py
```diff
@@ -815,3 +816,50 @@ def test_fill_with_sequence_4():
...
+
+
+def test_axes_lifetime():
+    h = bh.Histogram(bh.axis.Regular(10, 0, 1, metadata=2))
+
+    ax = h.axes[0]
+
+    # 2 is the minimum refcount, so the *python* object should be deleted
+    # after the del; hopefully the C++ object lives through the axis instance.
+    assert sys.getrefcount(h) == 2
```
henryiii: This being 2 is a little worrisome, hopefully keepalive is working here.

PR #239, opened 2019-11-23
fill improvements for StrCategory

HDembinski:
Despite the name of the branch, it is not really faster now, but the implementation is improved. Once we upgrade to C++17 and get std::string_view, I would give it another try to make it faster. The current version is allocating a temporary string for each item passed to StrCategory.


HDembinski:
Other features:
- 0-dim array work now in fill and axis.index as expected
- XCategory.value now returns None when an index out of range is passed instead of raising, this is more consistent with the other axes that also return something when you do that

HDembinski:
Other notes:
- Sorry, I had to move the axis types around in axis.hpp.
- I moved the type caster to the headers where they are used to make sure the specialization is visible in every translation unit where the caster is used

HDembinski:
One last comment: the vectorized version of index and value of the category axis lost the ability to handle arrays with ndim > 1, but I was really unhappy with the old implementation, which relied on a lot of details. The new one is straight-forward.

include/boost/histogram/python/axis.hpp
```diff
@@ -87,89 +178,76 @@ decltype(auto) unchecked_bin(const A& ax, bh::axis::index_type i) {
...
-    return result;
-}
-
-/// Utility to convert bins of axis to numpy array
-template <class A>
-py::array bins(const A& ax, bool flow = false) {
-    // this indirection is needed by pybind11
-    return bins_impl(ax, flow);
-}
+// template <class A>
```
henryiii: Why is there lots of commented out code here?

henryiii:
Please squash and merge when ready, maybe with a quick note about why there's a big block of commented code.

include/boost/histogram/python/axis.hpp
HDembinski: I noticed it is not used any more, which surprised me. To indicate this, I uncommented the code. Why are we not using `.bins`?

HDembinski: I wrote this code, but I don't remember why it is not used any more. Apparently, we don't need it. It is probably for the best, since it is redundant: if you want all bins you can do `list(axis)` or `np.array(axis)`.

HDembinski: Looking at it again. With `.bins(True)` you also get the extra flow bins.

HDembinski: Why is `.edges` a property, it has to be a method, so that you can pass `.edges(True)` to get all edges.

HDembinski: Lol, I changed this myself in #121, now why did I do that...

HDembinski: I think the reasoning was to consistently hide the flow bins in the axis interface (unless explicitly requested with `ax[bh.underflow]` etc.). The axis descriptions and the view into the storage memory have to be synchronized anyway, and the only place where we can offer this kind of synchronisation is in `.to_numpy`, so that would be the one place to request the flow bins. Still makes sense, no?

HDembinski: i removed the commented code and added a comment for .edges

PR #243, opened 2019-11-23
Restructure headers

HDembinski:
Pending on my other PR #239. Only the last commit is the actual move.


henryiii:
Make sure the change either are small enough for git to track the rename, or (better) make this two commits, one with only renames, and one with the file changes, and we'll "rebase and merge". I'm happy to split it for you if you'd like.

HDembinski:
Tracking the rename is important, agreed. Let's review this after the other patch is merged and I rebased this. I think git is properly tracking the renames.

HDembinski:
Ah I see, it does not track the rename for all headers. Will do two separate patches then.

HDembinski:
I sorted the headers such that local headers come first then system headers. This allows us to better detect if a header implicitly depends on other headers, which is bad.
https://stackoverflow.com/questions/2762568/c-c-include-header-file-order

HDembinski:
I also did some minor clean up while going through all headers. Some includes were not needed anymore.

HDembinski:
Feel free to merge this if you see no issues.

henryiii:
I think it's fine. I might need to play with the display in IDEs a bit in CMake.

henryiii:
GitHub switched to squash and merge after I changed it, fixing...

henryiii:
Fixed manually.

PR #244, opened 2019-11-24
Faster category str2

HDembinski:
Warnings: needs a rebase after #243 is merged, then will look much simpler. Do not yet review!

Pending on #243 

- ~[ ] needs implementation of index and update that does not create a temporary std::string~
- [x] some benchmarks

### DIFF
--- a/include/array_like.hpp
+++ b/include/array_like.hpp
@@ -1,0 +1,35 @@
+// Copyright 2018-2019 Hans Dembinski and Henry Schreiner
+//
+// Distributed under the 3-Clause BSD License.  See accompanying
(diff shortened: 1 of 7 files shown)


HDembinski:
It turns out that using a string_view is actually slower, at least for short strings, where short-string-optimization can be applied. Nevertheless, the custom `std::vector<std::string>` is useful, it accelerates conversion from a numpy array of type S by a factor of two.

HDembinski:
Now we also have fast fill from array U, which is the type that numpy uses when you do `np.array(("A", "B"))`

PR #254, opened 2019-11-26
test cleanup

HDembinski:


henryiii:
👍 Thanks for the axes cleanup, I forgot that one!

PR #256, opened 2019-11-26
Add density to numpy

henryiii:
Numpy interface was missing density, now it's easy to calculate, so adding it back in. Also fixed a bug where the requested storage type was being ignored.

### DIFF
--- a/boost_histogram/numpy.py
+++ b/boost_histogram/numpy.py
@@ -25,17 +25,19 @@ def histogramdd(
         if bh_warning is not None:
             import warnings
 
-            warnings.warn("bh= has been replaced by bh_object=", FutureWarning)
+            warnings.warn("bh= has been replaced by bh_cls=", FutureWarning)
+            bh_warning = bh.Histogram
         else:
-            bh_warning = False
-        bh_object = k.optional("bh_object", bh_warning)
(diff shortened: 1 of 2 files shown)


HDembinski:
I am not going to review the numpy interface, so feel free to go ahead with this. If you want to review, please give a shout. I only strongly care about the main interface of boost.histogram.

henryiii:
The current `bh=` shortcut is a bit criptic. Here are possible suggestions:

1. `bh=True, bh_storage=bh.storage.Double`: This is what we have now.
2.  `object=True, storage=bh.storage.Double`: This is more descriptive, but not clear that it's boost-histogram specific. Looks like a numpy option. *Unlikely* to ever conflict.
3. `histogram_object=True, histogram_storage=bh.storage.Double`: This still doesn't make it very clear that this specific to boost-histogram, and a bit wordy.
4. `bh_object=True, bh_storage=bh.storage.Double`: This is fairly clear that it comes from boost histogram, and still clear enough.

I'm leaning toward option 4, I think it's a nice blend of length + intent. Thoughts?

henryiii:
Added option 5: `object=bh.Histogram, storage=bh.storage.Double`.

PR #257, opened 2019-11-27
Add way to check the approx compile info

henryiii:
This adds a way for developers to monitor the compile stats. Ideally, you probably should build a single target, like `src/register_histograms.o` or build with a single core, `-j1` (the default with make).


HDembinski:
Looks good. By the way, it would be really nice to show the compiler and compiler version when setup.py is building.

henryiii:
> By the way, it would be really nice to show the compiler and compiler version when setup.py is building.

I'm not in disagreement there! I just haven't sat down and worked up with a good way to do it yet. Suggestions welcome.

PR #259, opened 2019-11-27
Reduce in Python

### LINKED ISSUE #252: new `.reduce` interface
```
    h = bh.Histogram(bh.axis.Regular(20, 1, 5))
    h.fill(1.1)
    hs = h.reduce(bh.algorithm.rebin(0, 4))
    assert_array_equal(hs.view(), [1, 0, 0, 0, 0])
```
`rebin` should not live in algorithm, this makes no sense now since all the algorithms are methods. It still makes sense for the cpp skin. So `bh.cpp.algorithm.rebin` is ok, but for the main Python skin, it should be `bh.tags.rebin(0, 4)`. For a 1D histogram, you should be able to use `bh.tags.rebin(4)` just like in C++. Similar for `shrink` and `shrink_and_rebin`.

henryiii:
This:

* Adds the missing axis=0 shortcut in `cpp` mode
* Adds a warning to the developer Regular axis shortcut
* Removes the `boost_histogram.algorithm` module
* Implements a Python interface to indexing using a dict

Closes #252 

---

#### Examples

```python
s = bh.tag.Slicer()
h.reduce({0: s[::bh.rebin(2)]})   # rebin axis 0 by two
h.reduce({1: s[0:bh.loc(3.5)]})   # slice axis 1 from 0 to the data coordinate 3.5
h.reduce({7: s[0:2:bh.rebin(4)]}) # slice and rebin axis 7
```


henryiii:
Note: we could use the Python `slice()` instead, if you want. So `h.reduce({7: slice(0, 2, 4)})`, `h.reduce({0: slice(None, None, 2)})`, etc. Just a thought. Sadly python has never supported a `slice[:,:,2]` shortcut for `slice(None, None, 2)`, etc, so you would only be gaining the clear 3 element meaning (rather than using a tuple and checking for length), and some similarity to normal indexing (for those that know that indexes with `:` become slices, anyway).

henryiii:
@HDembinski, what you you think of my latest edit to the above comment?

henryiii:
The original proposal is:  4f8a41c
The UHI based proposal (if we go with this, needs a little polish, consider it a proof of concept) is the latest commit: 44dd1b0

I'll revert the commit if we go with option 1. If we go with option two, I'll polish a bit and add the (now missing) docs from 1.

HDembinski:
I like the new version!

HDembinski:
I would only add a few more tests, to make sure that reduce also works for Nd histograms as expected.

HDembinski:
Passing a dict to `[]` also feels less surprising that passing a dict to a function call.

henryiii:
```python
class Slicer:
    def _getitem__(self, item):
        return item
slicer = Slicer()

h[{0: slicer[::bh.rebin(2)]}]   # rebin axis 0 by two
h[{1: slicer[0:bh.loc(3.5)]}]   # slice axis 1 from 0 to the data coordinate 3.5
h[{7: slicer[0:2:bh.rebin(4)]}] # slice and rebin axis 7
```

Would this be a useful utility to put somewhere? Or just a suggestion in the documentation?

HDembinski:
I am not a fan of fancy syntax hacks, but I think you have a point here: if people learned how to use our slicing notation, you naturally want to use them with the dict as well. As you showed, `slice()` has a completely different syntax, so that does not help. Perhaps the slicer can live in `bh.tags`?

HDembinski:
So would this support slicer **and** the tuple syntax? I think would prefer this.

HDembinski:
It looks like you cannot use tuples, but you can use the normal `slice` which is a builtin. I would like to use tuples as well, but it is not super important and no need to delay the release.

henryiii:
Okay. Getting it ready now.

---

I don't like duplicating the interface to support tuples for several reasons:

---

You can't easily generalize tuples to the non-dict interface. This:

```python
h[{0: (None, None, bh.rebin(2))}]
```

Can't be written as this:

```python
h[(None, None, bh.rebin(2))]
```

Since Python internally converts it to:

```python
h[None, None, bh.rebin(2)]
```

which is how you access a 3D histogram's cells. Contrast this to:

```python
s = bh.tag.Slicer()
h[{0: s[::bh.rebin(2)]}]
h[::bh.rebin(2)]
h[s[::bh.rebin(2)]] # A little silly, but the exact generalization works as well
```

---

The tuple has no inherent meaning. We are used to the meanings we have assigned, `(min, max)` and `(min, max, action)`, but anyone not developing boost-histogram would have to learn what we mean. A slice has a known meaning (we are slightly hijacking step as action, but that's a simple extension). And someone learning on 1 or 2D histograms will learn the normal slicing first. And, since tuples would only work inside dict slicing (as seen above), it's a *very* specific usage. It also does not imply a slice the way a `:` does when you read it.

---

The tuple is often more verbose than using `s = bh.tag.Slicer()`, especially if endpoints are left off. This really just becomes two nearly identical ways to do the same thing.

henryiii:
Currently only adding `Slicer` to `bh.tag`, maybe we can later add an instance, but I do rather think that this encourages you to write this without a namespace, 

```python
 s = bh.tag.Slicer()
```

Rather than doing:

```pyhton
from boost_histogram.tag import slicer as s
```
or using `bh.tag.slicer` (long), or `bh.tag.s` (not descriptive).
Open to suggestions there, though.

PR #264, opened 2019-11-28
Add .shape

henryiii:
This is a common part of the numpy syntax that also makes sense on Histograms.


HDembinski:
I went ahead and merged this, it seemed uncontroversial!

Issue #263, opened 2019-11-28
Minor repr issues


HDembinski:
Empty histograms do not display correctly:
```
In [2]: bh.Histogram()
Out[2]:
Histogram(
  ,
  storage=Double())
```

The cpp skin does not use the Pythonic axis repr, but should:
```
import boost_histogram.cpp as bhc

In [11]: bhc.histogram(bhc.axis.regular(10, 0, 1, metadata=None))
Out[11]:
histogram(
  regular(10, 0, 1, metadata="None", options=underflow | overflow),
  storage=double())
```

PR #265, opened 2019-11-28
Fix repr for 1 and 0d

henryiii:
This fixes half of #263, the Histogram part. Follows the original Boost.Histogram design with single line repr for 1D, and prints properly for 0D (and added tests). May need to need to be merged with changes in #262.


henryiii:
@HDembinski, I'm going to leave this to you to merge or combine with #262 so the collision is kept to a minimum.

HDembinski:
Thanks for this. I am waiting for the tests to finish then will merge.

PR #280, opened 2019-12-10
Add single-element UHI to axis

henryiii:
Single item UHI should work on axes indexing.

(Working on examples and docs)

(Slices as a way to slice axes would be very useful too, but should be a 0.7 feature)


henryiii:
(Merging so I can continue to work on the examples, @HDembinski let me know if there are any problems!)

PR #284, opened 2019-12-11
Fix missing inplace add, nicer inplace

henryiii:
Inplace add was simply falling back to out-of-place add; the binding for inplace add was missing.

Also improved the inplace operations to not change the python object ID either.


include/register_histogram.hpp
```diff
@@ -60,6 +60,9 @@ auto register_histogram(py::module& m, const char* name, const char* desc) {
              })
 
         .def(py::self + py::self)
+        // .def(py::self + value_type())
```
henryiii: Not (yet?) supported by Boost.Histogram.

Issue #296, opened 2020-01-02
Select an item or several items from a category axis


henryiii:
We currently have slice and single item selection (implemented internally as a slice + single item sum), but there is no way to select an item or a list of items from a category axes. Basically this:

If I have a `run_number` axis with `{1,2,5,7}`, I would like to be able to select the `5` only or `{1, 5}`. The second one will require new syntax in UHI (probably `[1,5]::bh.sum`). Users 

Also, a list selection syntax, for categories. So select 2 and 7, but not 4, for example. It would return a new, smaller category axis. Once we get full UHI, this could be implemented through a UHI object. Or we could come up with a new syntax, like [2,7]::bh.sum or something like that. Note that in Numpy, this would be a list, but there's no need to combine lists and slices, but we would often like to combine lists and actions.

This is easy to implement if it is supported upstream - is it @HDembinski? Basically a way to give a new category axis that is a subset of an existing one and ask for boost-histogram to copy one to the other would be enough.

Was mentioned in #274, but was not the focus of that issue. Brought up independently on gitter by @paulgessinger.

HDembinski:
Adding this in Boost.Histogram is not so easy, because it requires a large change to `reduce`. Currently, `reduce` can only select ranges. It is not possible to pick the 1, 3, 5, ... bin.

paulgessinger:
How about a range of length one? I’m interested in slicing just one category from a category axis. Would that work, or is that even possible already?

henryiii:
What about adding a "pick" axis bins to Boost.Histogram? The current workaround is just that; and that's why it is breaking for boost-histogram with category axis, but it would be useful for single item selection for any axis. So you could say:


```cpp
hist.pick(/*axis*/ 2, 4)
```

And it would remove the "2" axis, keeping only the contents in the 4th bin for the rest of the axes. (Syntax only for example, probably not ideal - some way to pick multiple axes in one go would be better)

Currently you have to sum over a slice of length 1 (unless I'm missing something).

henryiii:
> How about a range of length one? I’m interested in slicing just one category from a category axis. Would that work, or is that even possible already?

`h[:, bh.loc("signal")]` is short for `h[:, bh.loc("signal"):bh.loc("signal")+1:bh.sum]`. This normally works, but category histograms explicitly disable slicing, even by bin number (which is what the loc turns into).

[...]

PR #298, opened 2020-01-03
Fixing category indexing

### LINKED ISSUE #297: Fix summing over category axes
The slice and sum support broke unsliced sums (`::bh.sum`) over category axes.

henryiii:
Fix issues with indexing and summing category axes. You can now sum over a category axes again (regression in 0.6.0), and you can select a single item from a category axes just like you can from other axes types.

Fixes #297, improves #296 (single item selection now works on a category axis).

This may be improved upstream in Boost.Histogram later, in which case this can be simplified eventually.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,10 @@ Documentation update. Now using development branch of Boost.Histogram again.
 
 #### Bug fixes
 
-* Allow slicing on flowless axes [#288][]
-* Sum repr fixed [#293][]
+* Fix sum over category axes in indexing [#298][]
+* Allow single category item selection [#298][]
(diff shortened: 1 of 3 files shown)


henryiii:
@HDembinski Let me know if this seems like a reasonable workaround for 0.6.2; I assume Boost.Histogram will later add utilities that will make this easier. A single item no longer becomes a slice of length 1, and the trick shown on Gitter is used to produce the output. This also fixes sums over category axes, which were trying to slice in 0.6 and so were failing.

TODO: This internally tests sums over category axes in the tests now, but that should be added as an explicit test in the future to be safe when the algorithm changes.

PR #305, opened 2020-02-23
Fix a syntax bug

LovelyBuggies:
The sample code misses a ')', which can cause Syntax Error.

### DIFF
--- a/docs/usage/storage.rst
+++ b/docs/usage/storage.rst
@@ -23,7 +23,7 @@ a copy.
 
 .. code:: python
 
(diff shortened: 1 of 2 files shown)


LovelyBuggies:
And I think it better to show the experiment result by using 

```python
print(f"{h[bh.loc(.2)]=}\n{h[bh.loc(.3)]=}") # Python 3.8 print
``` 
rather than using 

```python
print(f"{h[bh.loc(.2)]=}\n{h[bh.loc(.2)]=}") # Python 3.8 print
```

PR #307, opened 2020-02-24
Fix bugs

LovelyBuggies:
Fix some bugs in example codes and docs.

PR #309, opened 2020-02-24
Fix CI for Python 3.8.1 replacing 3.8.0

### LINKED ISSUE #308: PR Check Unsuccessful
I pulled a request [#307](https://github.com/scikit-hep/boost-histogram/pull/307) to fix some bugs in the example codes and document. But git shows that the check is unsuccessful, plz help me with this problem! 

```
Run actions/setup-python@v11s
3.8.1 (x64)
Run actions/setup-python@v1
##[error]Version 3.8.0 with arch x64 not found
Available versions:

2.7.17 (x64)
3.5.9 (x64)
3.6.10 (x64)
3.7.6 (x64)
3.8.1 (x64)
```

henryiii:
GitHub Actions seems to delete old patch versions. Closes #308

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -18,7 +18,7 @@ jobs:
         submodules: true
     - uses: actions/setup-python@v1
       with:
-        python-version: '3.8.0'
+        python-version: '3.8'
     - name: Pre-install Numpy
       run: python -m pip install numpy nbconvert ipykernel
     - name: Install kernel
--- a/setup.py
+++ b/setup.py
@@ -130,7 +130,7 @@ class BuildExt(build_ext):
 extras = {
     "test": ["pytest", "pytest-benchmark", "numpy", 'futures; python_version < "3"'],
     "docs": ["Sphinx>=2.0.0", "recommonmark>=0.5.0", "sphinx_rtd_theme"],
-    "examples": ["numpy", "ipykernel", "matplotlib"],
+    "examples": ["numpy", "ipykernel", "matplotlib", "xarray"],
 }
 
 setup(


LovelyBuggies:
Henry, my Python Version is 3.7.0 for issue #309 and may be outdated for `print(f"{}")`function, I have PR #307 still in progress. **What can I do next?** To uninstall my Python 3.7.0 and reinstall OR to change my `.github/workflows/examples.yml` Python version to 3.7 like this?

```yml
jobs:
  examples:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v1
      with:
        submodules: true
    - uses: actions/setup-python@v1
      with:
        python-version: '3.7'
    - name: Pre-install Numpy
      run: python -m pip install numpy nbconvert ipykernel
    - name: Install kernel
      run: python -m ipykernel install --user --name boost-hist
    - name: Build
      run: python -m pip install .[examples]
    - name: Examples
      run: for f in examples/*.py; do python "$f"; done
    - name: Notebooks
      run: jupyter nbconvert --execute --inplace notebooks/*.ipynb
```

P.S. now my `examples.yml` file shows like this: `python-version: '3.8'`. It's puzzling. I did not rebase or merge your [new commit](https://github.com/scikit-hep/boost-histogram/commit/c6908940b86555fe829fe42455f29334138803cc).

henryiii:
This is very odd. The problem is that you don't have `xhistogram`, which I'm importing and comparing against. This is fine and should be added, but *how did the other PRs (like this one)* manage to successfully pass?

Note that the `print(f'{x}')` syntax was introduced in Python 3.6, so we should be fine with 3.6+.

henryiii:
Found the problem. Examples only run if something was changed in the notebook or example folder. So it hasn't been running, and therefore I haven't seen the issue before. We'll fix it in your PR.

henryiii:
A few tips that will help in the future:

* Always make a branch for working on patches. If you need to rebase or merge, it's a bit easier to have "develop" be "origin/develop" and, for example, "henryiii-myfix" be "henryiii/henryiii-myfix" (and it's *much* easier if you are the owner and you need to check out a fork's branch).
* Learn how to do an interactive rebase. They are very powerful and let you craft the perfect history for your branch before merging. This is another reason for branching, you should never edit the history of master and probably not develop.

We don't need to do this necessarily, but here's roughly what you'd do to clean up your history:

```bash
git fetch origin
git rebase -i origin/develop
# Use editor to select actions
# Drop changes that are already done elsewhere, optionally squash the extra commits
```

henryiii:
Note: this really is only important for large PRs where you want to keep some history (rebase and merge). For small PRs, I use the "Squash and merge" option, so history don't matter, it all gets squashed down in to a single commit anyway.

PR #320, opened 2020-02-29
Making dict access equivalent to positional

### LINKED ISSUE #319: Index Error
Get item index error: https://github.com/henryiii/mini_hist/issues/2#issue-572520900

henryiii:
Closes #319. Slightly simpler code, too, better sharing between dict and positional access.

### DIFF
--- a/boost_histogram/_internal/hist.py
+++ b/boost_histogram/_internal/hist.py
@@ -342,14 +342,16 @@ class Histogram(BaseHistogram):
 
         # Support dict access
         if hasattr(index, "items"):
-            return index, index.items()
+            indexes = [slice(None)] * hist.rank()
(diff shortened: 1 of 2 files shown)


ChanchalKumarMaji:
@henryiii my code for `mini_hist` passes the pytest tests in my local system, but fails when I commit in the server, because of this issue. How to fix that, because I think this PR has to be released to become effective and that may take time.

henryiii:
@ChanchalKumarMaji 

I will try to release something (probably 0.6.3 with very minor fixes, or maybe 0.7.0) before the 16th.

I've also changed the recommendation in the mini_hist repository to something that works with current 0.6.2, you could use that, and just note that the other one would work with the latest develop of boost-histogram.

PR #335, opened 2020-03-16
Prepare for 0.7.0

henryiii:
After merging this, will push develop to master and tag a release (if everything builds correctly).


include/bh_python/register_histogram.hpp
```diff
@@ -145,7 +145,7 @@ auto register_histogram(py::module& m, const char* name, const char* desc) {
                     throw std::out_of_range(
                         "The axis value must be less than the rank");
             },
-            "i"_a,
+            "i"_a = 0,
```
henryiii: @HDembinski We can drop this change now, your fix was the correct one.

HDembinski: Ok

PR #338, opened 2020-03-18
Remove unncessary print statement

douglasdavis:
I was playing with the `bh.numpy` API and I noticed this print statement is always reached when `threads` is not `None`. I'm assuming printing an object any time a particular argument is present is not desired behavior


henryiii:
Thanks! That slipped in.

PR #339, opened 2020-03-18
Two minor fixes

HDembinski:


src/boost_histogram/_internal/hist.py
```diff
@@ -192,7 +192,7 @@ def fill(self, *args, **kwargs):
 
         threads = kwargs.pop("threads", None)
 
-        if threads is None or threads is 1:
+        if threads is None or threads == 1:
```
henryiii: If someone passes something that is not comparable to 1, this will throw an error. The value 1 is only instantiated once in CPython, so it is safe to compare with `is`.

However, if something is not compatible with a number, it probably is invalid here, so I'm okay with the change.

src/boost_histogram/cpp/algorithm.py
```diff
@@ -56,13 +56,6 @@ def empty(histogram, flow=False):
     return histogram._empty(flow)
 
 
-def reduce(histogram, *args):
```
henryiii: Just curious, isn't this needed to match the C++ interface?

src/boost_histogram/_internal/hist.py
HDembinski: > If someone passes something that is not comparable to 1, this will throw an error.

Ok, I see your reasoning, but I would argue that we want the error to happen then.

I know that numbers are only instantiated once, but using `is` in this way is not recommended. This goes against duck-typing in Python. If the user passes an object which has `__eq__` defined, it should work. `1.0 is 1` evaluates to false, but `1.0 == 1` evaluates to true.

src/boost_histogram/cpp/algorithm.py
HDembinski: It was defined twice in this file.

HDembinski: same function, just different doc string.

HDembinski:
Ok, I am merging, this seems uncontroversial.

src/boost_histogram/_internal/hist.py
henryiii: > `1.0 is 1` evaluates to false, but `1.0 == 1` evaluates to true

Good point, though one could argue that floats for threads should be disallowed. Numpy for example has been forcing int-like for indexing. (I am not arguing that!) However, a custom int-like object should pass!

src/boost_histogram/cpp/algorithm.py
henryiii: Good catch!

src/boost_histogram/_internal/hist.py
HDembinski: I agree that passing floats is a bit stupid, but you can easily get a float by accident in Python 3:
```
# ok I have 8 cores
n_cores = 8
# and I want to run threads on half of them
n_threads = n_cores / 2
# n_threads is a float now!
```

henryiii: This is incorrect in Python 3, it should be (and this works in Python 2):

```python
n_threads = n_cores // 2
```

This is now what you have to do in Python if you want to use division in an indexing expression in Numpy, as well; Numpy no longer accepts floats.

henryiii: PS: "incorrect" does not mean people will not try it.

Issue #332, opened 2020-03-11
Add test: default value for boost_histogram.cpp.histogram.axis()


HDembinski:
The axis method has a default value for the index of 0.

henryiii:
Hans, changing this does not break the tests, so I'll do it now, but could you please add a test?

HDembinski:
I was not fixed in develop, I fixed it in #340 and added the test.

PR #340, opened 2020-03-18
Add axis default like in C++ and test

HDembinski:
Fix for #332

### DIFF
--- a/src/boost_histogram/_internal/hist.py
+++ b/src/boost_histogram/_internal/hist.py
@@ -279,7 +279,7 @@ class BaseHistogram(object):
             s = repr(self)
         return s
 
-    def _axis(self, i):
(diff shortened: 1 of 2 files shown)


src/boost_histogram/_internal/hist.py
```diff
@@ -279,7 +279,7 @@ def __str__(self):
             s = repr(self)
         return s
 
-    def _axis(self, i):
+    def _axis(self, i=0):
```
henryiii: My mistake, I forgot this just gets assigned to axis!

HDembinski: No problem at all. It is ok if you focus on the main interface and I try to straighten the cpp interface.

PR #343, opened 2020-03-26
Allow cloudpickle to work on histograms

### LINKED ISSUE #342: Cloudpickle fails on boost-histogram
I'm using boost histograms inside Coffea accumulators. I ran into the issue now that boost histograms appear not to be pickleable by cloudpickle (anymore? I think this used to work). 

When running something like this:

```python
import cloudpickle
h = bh.Histogram(bh.axis.Regular(50, 0, 20))
h.fill([1 ,2 , 3, 4, 5])
cloudpickle.dumps(h)
```

I get an error like

```
AttributeError: module 'boost_histogram._core' has no attribute '__path__'
```

Should that even work? If so, any pointers how I can resolve or circumvent this issue? This is especially unfortunate since standard pickle appears to work fine, but Coffea uses cloudpickle.

henryiii:
This is a bit hacky, a better fix should come from cloudpickle or pybind11. Fixes #342.

This is correctly fixed in https://github.com/cloudpipe/cloudpickle/pull/357 - when that is merged and released, this can simply become a test with a new version of cloudpickle.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -103,7 +103,7 @@ extras = {
         "sphinx_copybutton",
     ],
     "examples": ["matplotlib", "xarray", "xhistogram", "netCDF4", "numba"],
-    "dev": ["pytest-sugar", "ipykernel"],
+    "dev": ["pytest-sugar", "ipykernel", "cloudpickle"],
 }
 extras["all"] = sum(extras.values(), [])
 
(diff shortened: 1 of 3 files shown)


henryiii:
I don't think there's any harm in applying this patch, so let's put it in (we can remove it if there is harm or when cloudpickle gets fixed and released).

Issue #289, opened 2019-12-17
Weight fill on integer storage


henryiii:
The `weight=` keyword argument for integer storages (Int64(), AtomicInt64()) currently accepts both integers and doubles, silently converting to integers; it should only accept integers, and provide an error if a floating type is used.

Unlimited does the correct thing and converts it's internal storage to double, so that one is fine.

PR #346, opened 2020-03-28
forbid float weights for int storage for #289

LovelyBuggies:


LovelyBuggies:
@henryiii I have fixed the bugs in #289 . I saw your work and found you focused on underlying. Basically, I found the `self._hist._storage_type` cannot match Python data type like `int`, so I have to add the `PyType = int()` to allow them to communicate. I think I can better the coding style with your help. **But now, the target has been achieved.**

```python 
import boost_histogram as bh
import numpy as np

h = bh.Histogram(bh.axis.Regular(10, 0, 1), storage=bh.storage.Int64())
h.fill([.2, .3], weight=[1, 2])   # this will work
h.fill([.2, .3], weight=[1., 2])  # this will not work
h.fill([.2, .3], weight=[1, 2.])  # this will not work
h.fill([.2, .3], weight=[1., 2.]) # this will not work

w, data = h.to_numpy()
```

This snippet passed in my notebooks.

LovelyBuggies:
By the way, CI failed again. I wonder why this always happens. 😢

src/boost_histogram/_internal/hist.py
```diff
@@ -98,11 +98,11 @@ def __init__(self, *axes, **kwargs):
 
         # Keyword only trick (change when Python2 is dropped)
         with KWArgs(kwargs) as k:
-            storage = k.optional("storage", Double())
+            self._storage = k.optional("storage", Double())
```
henryiii: You can access the storage type from the C++ object, which is better than trying to hang on to it in Python.

src/boost_histogram/_internal/hist.py
```diff
@@ -195,9 +195,14 @@ def fill(self, *args, **kwargs):
             threaded filling.  Using 0 will automatically pick the number of
             available threads (usually two per core).
         """
-
+        # should not allow float weight for int storage
+        if "weight" in kwargs:
+                for w in kwargs["weight"]:
+                    if not isinstance(self._storage._PyType, type(w)):
```
henryiii: This is an incorrect check; weight could be a single value or an array of values, this is only checking for one int rather than an array (I think).

src/boost_histogram/_internal/storage.py
```diff
@@ -17,6 +17,8 @@ def __repr__(self):
 @set_family(MAIN_FAMILY)
 @set_module("boost_histogram.storage")
 class Int64(store.int64, Storage):
+    def __init__(self):
+        self._PyType = int()
```
henryiii: I think this is causing the segfault in the CI, as Int64 is not properly initializing store.int64 (but could be wrong). The fact the tests are segfaulting is slightly bothering me.

src/boost_histogram/_internal/hist.py
henryiii: Interesting, it does work on arrays of ints? Something like this might fix the other PR.

LovelyBuggies: You might misunderstand. As you can see, `for w in kwargs["weight"]:` will iterate every value in weight list. And you can see the demo above in my comments. Every possible broken case is considered.

henryiii:
This PR doesn't fix the Unlimited storage, since internally they still are always converted to floats, while the the other PR does fix that.

src/boost_histogram/_internal/hist.py
henryiii: Ah, okay. I see the loop. Does this work when this is not an array? And this will be horribly slow, since it's a Python loop.

LovelyBuggies:
> This PR doesn't fix the Unlimited storage, since internally they still are always converted to floats, while the the other PR does fix that.

@henryiii Do you expect a build-in exception rather than `raiseError`?

src/boost_histogram/_internal/hist.py
LovelyBuggies: You get the point! Only `weight=[1]` is supported now. `weight=1` is not supported. But it would not be a hard job.

Yep, I have seen the importance to deep into the underlying C++ parts.

PR #357, opened 2020-05-14
Nino/aghast-tutorial

LovelyBuggies:


LovelyBuggies:
@henryiii This is for https://github.com/scikit-hep/aghast/issues/39

- Unify the upper-case notebook title and change the symlinks correspondingly.
- Add aghast.ipynb in boost-histogram/notebooks (not in docs yet).
- Bh <-> Root histogram conversion.

Note that it's not easy to test this branch as I am using the newest code (not the latest version).

LovelyBuggies:
I am using the aghast kernel with the ROOT package installed, so the CI failed. Any suggestions?

henryiii:
If we have to use the ROOT version instead of the uproot version (depends on what Jim has implemented), then we'll need a custom environment for this. I can set that up this weekend if needed.

LovelyBuggies:
Yes, we need to `import ROOT`, because Aghast's `to_root()` will call underlying ROOT parts.

> (literally the one you have at the top of the file)

😂That is Aghast's `environment-test.yaml`. I directly used Aghast's kernel. If you think it's okay, I can complete exactly what I have done in Aghast's conda installation (just change the env and kernel's names). But not sure how to deal with the CI.

henryiii:
I can fix the CI for it this weekend, we can have an `aghast-environment.yaml`.

There should be a `to_uproot`, right? I think Jim said he provided both but didn't remember which was implemented (or how fully they were implemented).

LovelyBuggies:
> I can fix the CI for it this weekend, we can have an aghast-environment.yaml.

Okay.

> There should be a to_uproot, right? I think Jim said he provided both but didn't remember which was implemented (or how fully they were implemented).

There might be, but I am not sure we really need it. Why? This notebook is mainly to introduce conversion between ROOT and BH histograms, and we don't have to include uproot. I use it now because I want to show how to save and load ROOT objects (aghast is not able to do this now). So, to some extent, we don't have to include the uproot's part. In addition, to convert between ROOT and BH histograms, we don't need the uproot intermediates. If you think it's better to show how those packages work as a whole, I can add it.

LovelyBuggies:
Aghast has many powerful functions (to pandas, to binary, etc.), but some of them cannot stand the tests. As I said, **I try to make everything simple and will not use uproot unless I have to**. If you want some specific functions, I will first find and test it in Aghast. If I cannot find it in Aghast or it doesn't stand the test, then I will implement by uproot.

henryiii:
Uproot is pure-Python, while ROOT is a massive compiled library. Both read and write ROOT files. Uproot should, therefore, be the simpler, unless it is a function that is not implemented yet?

(Uproot should be able to read any histogram, it is only limited in writing.)

LovelyBuggies:
> Uproot should, therefore, be the simpler, unless it is a function that is not implemented yet?

Yeah, I know ROOT is complicated than uproot. But didn't Hans say to convert a ROOT object to a BH object? We cannot bypass ROOT, that is, we need to `import ROOT` anyway.

henryiii:
I've used this as a base to add an initial tutorial file to scikit-hep-tutorials, giving credit to you (thanks!). Could you please remove the ROOT parts and just leave the uproot example at the end? We'll leave ROOT for the tutorials site because Conda is not set up for boost-histogram (and it should not be here). Also please look at my changes to the uproot example and copy them here.

Note: Generally most people don't care about using PyROOT, which is weird and hijacks things like tab completion and breaks them, but rather using the ROOT file format. PyROOT may get much better in ROOT 6.22.

LovelyBuggies:
@henryiii I saw your work for scikit-hep tutorials: I think it's great. Can we transfer the work directly to this (**copy** the ipynb file except to remove the `import ROOT`)?

henryiii:
Sure, though we will need remove the JupyterBook specific bits from the markdown (just a bit here and there).

PR #358, opened 2020-05-21
Update to latest dev histogram

henryiii:
Updating to latest BH. Now supports addition of histograms with growing categories that do not match! Using `+=` now recomputes the axes cache.

> Note: It would be nice if the axes objects could know if they are no longer pointing at a valid axes object.

MyPy used for some *very* basic checking, not tested in CI yet. Cleaned up a couple of places at its request, and fixed a few potential bugs!

### DIFF
--- a/src/boost_histogram/__init__.py
+++ b/src/boost_histogram/__init__.py
@@ -24,9 +24,25 @@ __all__ = (
 try:
     from . import _core
 except ImportError as err:
-    if "_core" in err.msg and "boost_histogram" in err.msg:
-        err.msg += "\nDid you forget to compile? Use CMake or Setuptools to build, see the readme"
-    raise err
+    msg = str(err)
+    if "_core" not in msg:
+        raise
+
+    import sys
+
+    new_msg = "Did you forget to compile boost-histogram? Use CMake or Setuptools to build, see the readme."
+    total_msg = "\n".join([msg, new_msg])
+
+    # Python 2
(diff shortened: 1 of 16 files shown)


henryiii:
The broken test is an issue with the 4.0 release of `setuptools_scm` missing a Python 2 wheel (possibly because of its close relationship with `setuptools`, which has already dropped Python 2 support). Will fix by pinning if it can't be fixed otherwise (see https://github.com/pypa/setuptools_scm/issues/442).

src/boost_histogram/_internal/view.py
```diff
@@ -63,9 +63,7 @@ def name(self, value):
     def injector(cls):
         if hasattr(cls, "_FIELDS"):
             raise RuntimeError(
-                "{0} already has had a fields decorator applied".format(
-                    self.__class__.__name__
-                )
+                "{0} already has had a fields decorator applied".format(cls.__name__)
```
henryiii, line 66: MyPy discovered this error.

src/boost_histogram/__init__.py
```diff
@@ -24,9 +24,25 @@
 try:
     from . import _core
 except ImportError as err:
-    if "_core" in err.msg and "boost_histogram" in err.msg:
```
henryiii, line 27: MyPy discovered this error - "msg" is not public API and has a different name in Python 2.

Issue #208, opened 2019-11-12
Full UHI


henryiii:
The full UHI includes arbitrary reducers and rebinnings. This will need to be implemented before 1.0, but depends on #194 and #189; those should go into 0.6.0 so let's nominally target 0.7.0 for a full UHI implementation.

henryiii:
Suggested API (rebinning):


```python
class Rebinner:
    def __call__(self, axis: PlottableAxis) -> Sequence[int] | Mapping[int | str, Sequence[int | str]]:
        """
        axis: axis object
        """
```

On Regular, Variable, Integer axes, this would return a sequence. For example, `Integer(0, 10)`, `[0,5,2,3,0]` (probably the output of `bh.rebin([5,2,3])`) would become `Variable([0,5,7,10])`. First and last groups get put into underflow/overflow (or removed, if no flow bins). Summing this should always equal the total number of bins, error if not. For Cat axes, mappings are used. So for `IntCat([1,3,5,7,9])`, then a valid rebin would be `{0: [1,3], 1: [5,7,9]}}`. Another might be `{"13": [1,3], "579": [5,7,9]}}`.

@jpivarski, I know that reducers were also planned. I think a reducer can't be single argument (It at least needs the full data + the axis it's operating on?), so I think this won't block future implementation of that if needed?

For `bh.rebin`, we could come up with a nice API for common operations: edge based rebins, group-based rebins (making flow parameters, like `bh.rebin([1,2], overflow=3)`), and mappings.

PR #365, opened 2020-05-30
Built-in sum

henryiii:
This would convert to using the built-in sum. @HDembinski , can you sign off on this, just to make sure you agree to it?

When UHI is completed, any vector -> scalar function will be supported here. Rebinning generally will be even tricker, since it has to change the axes. (See #208)


HDembinski:
Looks ok to me.

PR #366, opened 2020-05-31
chore: More checks in place

henryiii:
Adding Flake8, bugbear, and a few other tests, including *very* basic MyPy. Check-manifest added but optional (and mostly run by CI only). Fixed a few things that could be confusing.


henryiii:
Added changes to https://github.com/scikit-hep/scikit-hep.github.io/pull/60 as well.

PR #367, opened 2020-06-01
Nino/aghast tutorial

LovelyBuggies:


LovelyBuggies:
Sorry. I have pushed a wrong commit by mistake in the last branch because the things became odd after your new commits. 

But anyway, I fixed it in an awkward way, that is deleting the old branch and pushing the reset to a new branch. Is there any way to reset the remote without deleting the origin branch?

docs/notebooks/BoostHistogramHandsOn.ipynb
```diff
@@ -1 +0,0 @@
-../../notebooks/BoostHistogramHandsOn.ipynb
```
LovelyBuggies, line 1: Have no idea about this :/

LovelyBuggies: Whitespace issue?

henryiii: Different file permissions? Odd, not sure I've seen this show up this way before...

henryiii: I bet if we squash your history, this will go away.

LovelyBuggies: Alright. See how it will be after the squash.

henryiii: I can do a squash and force-push, is that okay?

LovelyBuggies:
@henryiii Henry, it's okay to review. But CI says he cannot find aghast.

henryiii:
I'll fix it. We need to remove aghast and ROOT; this will just be an example of how to save with uproot. The scikit-hep-tutorials will be where Aghast, ROOT, etc. will be.

LovelyBuggies:
> I'll fix it. We need to remove aghast and ROOT; this will just be an example of how to save with uproot.

Okay.

henryiii:
Somehow the simlinks were changed. Anyway, I've refreshed the example and changed it to be pure uproot. boost-histogram is not supposed to be HEP specific, so we definitely can't expect people to spend hours installing ROOT or changing to Conda. :)

A few tips:

* Let's keep examples focused and simple (I removed the string save example); scikit-hep-tutorials can be less focused.
* We need to revisit this when we simplify hist / bh -> uproot
* Let's make sure all checked-in notebooks have no outputs - they get built by the CI
* We may need to update the uproot->bh conversion in scikit-hep-tutorials, I made it much more general (supports flow bins and could support weights)

PR #370, opened 2020-06-03
feat: Adding addition of 0

### LINKED ISSUE #261: Allow adding constants
Consider the following code (from my analysis), which sums up histograms generated in different threads:
```
h_tot = None
for h in generate_histograms():
  if h_tot is None:
    h_tot = h
  else:
    h_tot += h
```
It would be nicer to be able to write it like this (or something similar)
```
h_tot = 0
for h in generate_histograms():
   h_tot = h_tot + h
```
We currently don't allow adding numbers to histograms. So to get this feature, we should allow this. Adding a number to a histogram would only be allowed if that is supported by the underlying accumulator.

henryiii:
Adding with a false-like, such as 0 or None, is now supported. This allows syntax like `sum(histogram_list)` without answering the hard question of what it means to add a constant to a histogram with flow bins and unevenly sized bins. That can be done much more clearly using the view or `[...] =` syntax.

Closes #261.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,5 +69,5 @@ compile_commands.json
 # Testing
 /.benchmarks/*
 
-# IDEs
-/.idea
+# PyCharm
+.idea
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,7 +220,7 @@ message(STATUS "boost-histogram ${BOOST_HISTOGRAM_VERSION}")
 
 if(BUILD_TESTING)
 
-    python_import(numpy pytest pytest-benchmark six typing)
+    python_import(numpy pytest pytest-benchmark typing)
 
     # Support for running from build directory
     file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/pytest.ini"
--- a/src/boost_histogram/_internal/hist.py
+++ b/src/boost_histogram/_internal/hist.py
@@ -155,12 +155,25 @@ class BaseHistogram(object):
         return self.view(False)
 
     def __add__(self, other):
-        return self.__class__(self._hist.__add__(other._hist))
+        if hasattr(other, "_hist"):
+            return self.__class__(self._hist.__add__(other._hist))
+        else:
+            retval = self.copy()
+            retval += other
+            return retval
 
     def __iadd__(self, other):
-        self._hist.__iadd__(other._hist)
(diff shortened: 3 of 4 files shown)


henryiii:
@HDembinski, thoughts? Is this fine?

HDembinski:
It is weird that adding None and False also works. Adding 0 makes sense, but why allow the others?

henryiii:
Partially because it is natural in construction (you can just check truthiness, and don't have to require a specific type check; any false-like does not affect the histogram when added - you can have `WeightedSum(0,0)` be false-like an then it would be immediately included). We could do type checks though if it's better to restrict things like `None + hist == hist`.

False will always work, as it's a subclass of integer with value 0 (It was just an integer before Python 2.2 or 2.3, IIRC). `1 + False` is valid Python syntax. As is `1 + True == 2`. `isinstance(False, int)` is True.

HDembinski:
Ok, I understand better now. If the argument to `__radd__` is not another histogram, I would again  push the responsibility to what should happen to the numpy view of the internal storage. This way, we get all the numpy broadcasting rules automatically consistent, including broadcasting of scalars.

The question then again is what should happen with the flow bins. The consistent thing is to include them, since the other operators also work on the flow bins.

henryiii:
Okay, will place on hold until we have more support for operators on views. We need at least addition then for all views. Is it legal to add a double/int to each of the 3 accumulators?

HDembinski:
>  Is it legal to add a double/int to each of the 3 accumulators?

That should work for WeightedSum out of the box, but not for other Mean or WeightedMean.

henryiii:
So Mean and WeightedMean should not support `sum(histogram_list)`?

HDembinski:
They should, but they cannot support addition of constants in general, unless we find a reasonable interpretation of adding a number to a mean. So far, we should only allow the addition of 0, which we define to do nothing.

HDembinski:
You probably already know this:
```
import numpy as np
a = np.zeros(3)
a + 0 # ok
a + False # ok, False interpreted as zero
a + None # Error: unsupported operand type(s) for +: 'float' and 'NoneType'
```
We should also raise an error in case someone adds `None`.

henryiii:
Okay, we should just check for int/float 0 and only allow that, then.

PR #384, opened 2020-07-02
fix: AxesTuple slicing keeps type

### LINKED ISSUE #383: Slices of AxesTuple should return AxesTuple
The `AxesTuple` is lost after slicing, this leads to some unintuitive asymmetries.

```py
import boost_histogram as bh

h = bh.Histogram(bh.axis.Regular(10, 0, 1))
type(h.axes) == "boost_histogram._internal.axistuple.AxesTuple"
type(h.axes[:1]) == "tuple" # should return another AxesTuple

h.axes.centers[:1] # ok
h.axes[:1].centers # bummer, does not work, but should
```

henryiii:
Resolves #383.

feat: Slicing on the tuple before the method call controls the shapes of the result.

Also fills in the aghast outputs (missing previously)

### DIFF
--- a/src/boost_histogram/_internal/axistuple.py
+++ b/src/boost_histogram/_internal/axistuple.py
@@ -65,6 +65,15 @@ class AxesTuple(tuple):
             )
         return tuple(self[i].index(values[i]) for i in range(len(values)))
 
+    def __getitem__(self, item):
+        result = super(AxesTuple, self).__getitem__(item)
+        return self.__class__(result) if isinstance(result, tuple) else result
+
+    # Python 2 support - remove after 1.0
+    def __getslice__(self, start, stop):
+        result = super(AxesTuple, self).__getslice__(start, stop)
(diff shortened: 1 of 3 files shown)


src/boost_histogram/_internal/axistuple.py
```diff
@@ -65,6 +65,15 @@ def index(self, *values):
...
         return tuple(self[i].index(values[i]) for i in range(len(values)))
 
+    def __getitem__(self, item):
+        result = super(AxesTuple, self).__getitem__(item)
+        return self.__class__(result) if isinstance(result, tuple) else result
+
+    # Python 2 support - remove after 1.0
+    def __getslice__(self, start, stop):
+        result = super(AxesTuple, self).__getslice__(start, stop)
+        return self.__class__(result) if isinstance(result, tuple) else result
```
henryiii, line 75:
```suggestion
        return self.__class__(result)
```

PR #386, opened 2020-07-02
fix: Type safety for string index

henryiii:
This fixes a type safety issue where types could be mismatched; for example ints could be used in locators for StrCategory, when really only strings make sense.

Simplify testing a bit with a parametrized fixture (see [nice article here](https://levelup.gitconnected.com/advanced-pytest-techniques-i-learned-while-contributing-to-pandas-7ba1465b65eb)). This could probably be used elsewhere. Also improves testing output with IDs for growth.

### DIFF
--- a/src/boost_histogram/_internal/axis.py
+++ b/src/boost_histogram/_internal/axis.py
@@ -13,6 +13,20 @@ import copy
 del absolute_import, division, print_function
 
 
+def _isstr(value):
+    """
+    Check to see if this is a stringlike or a (nested) iterable of stringlikes
+    """
+    str_types = (type(""), type(""))
+
+    if isinstance(value, str_types):
+        return True
+    elif hasattr(value, "__iter__"):
+        return all(_isstr(v) for v in value)
+    else:
+        return isinstance(value, str_types)
+
+
 # Contains common methods and properties to all axes
 @set_module("boost_histogram.axis")
 class Axis(object):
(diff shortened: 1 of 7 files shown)


HDembinski:
I would prefer if you added this kind of convenience to hist. I don't like that this is breaking the general rule for using axis values in getitem. We can always backport features that proved themselves in hist into boost-histogram.

HDembinski:
The feature is definitely useful, but I would like boost-histogram to be less convenient (for now) and instead more rule-based without exceptions.

henryiii:
That's perfectly fine, it was originally targeting Hist but thought it was fairly straightforward shortcut. I'll drop the shortcut from this PR, but will keep the fixes and improvements. It will show up in Hist soon (along with the complex shortcut for numerical axes, of course - together they allow all axes to be indexed directly without extra imports) - @LovelyBuggies is working on shortcuts now.

src/boost_histogram/_internal/axis.py
```diff
@@ -13,6 +13,20 @@
 del absolute_import, division, print_function
 
 
+def _isstr(value):
+    """
+    Check to see if this is a stringlike or a (nested) iterable of stringlikes
+    """
+    str_types = (type(""), type(""))
```
henryiii, line 20: Something is stripping the `u`'s from these...

henryiii: Doesn't seem to be Black or anything else in pre-commit...

Issue #412, opened 2020-07-12
AxesTuple.edges too deeply nested


HDembinski:
```py
h = bh.Histogram(bh.axis.Regular(5, 2, 10), bh.axis.Regular(2, 0, 6))
print(h.axes.edges)
```
produces
```py
[array([[ 2. ],
       [ 3.6],
       [ 5.2],
       [ 6.8],
       [ 8.4],
       [10. ]]), array([[0., 3., 6.]])]
```
I think the arrays should be 1D, not 2D. This would allow one to do, for example,
```py
pyplot.pcolormesh(*h.axes.edges, h.view().T)
```
which currently does not work.

henryiii:
No, this *is* ND data, with NumPy broadcasting optimizations. The reason it does not work is that Matplotlib expects ij ordering instead of xy - which I think was a mistake (at least as the only input).

If you made this 1D, you would loose all ability to make vectorized calculations with this, that is

```python
X, Y = hist.axes.centers
X + 2*Y 
```

 (computes x + 2y for each element, calculations like these are used in computing a projection from a histogram to a plane, which was a lot of my Thesis work…)

```python
X, Y = hist.axes.widths
Volume = X * Y
```

This is how we compute density, etc…

The hint to the error here should be the `.T` - that’s the weird ij indexing. The correct way to do this is:

```python
plt.pcolormesh(h.axes.edges[0].T, h.axes.edges[1].T, h.view().T)
```

This works as it should, since now you’ve fixed the indexing issue - the 2D data wasn’t the problem. We could add a specialized tuple here too, though, and allow:

```python
plt.pcolormesh(*h.axes.edges.T, h.view().T)
```

Or something else like `.ij.centers` and/or `.flat.centers`, etc.

Making this 1D is also more error prone:

```python
plt.pcolormesh(h.axes.edges[1].flatten(), h.axes.edges[0].flatten(), h.view().T)
```

Is a clear mistake, but it works if the axes have the same dimensions. While

```python
plt.pcolormesh(h.axes.edges[1].T, h.axes.edges[0].T, h.view().T)
```

Breaks, like it should, even if the axes are the same dimensions, since it detects the shape issue.


There are, however, some possible fixes from the Matplotlib side. One is that they could add an argument to allow xy indexing - then this would become:
plt.pcolormesh(*h.axes.edges, h, order=“xy”) # you don’t even need the .view() now!

[...]

PR #414, opened 2020-07-13
feat: allow quick transpose of all results

henryiii:
@HDembinski , this is what my suggestion in #412 would look like. There are two possible methods. The first commit simply adds `.T`. The second commit allows any `np.array` action to operate on the tuple. So `h.axes.edges.ravel()` is also valid. (or `.flatten()`, but that always makes a copy). Tab completion still works in IPython (just not static typing). This feels just like the AxesTuple itself, actions are applied to each item in the tuple.

### DIFF
--- a/notebooks/BoostHistogramHandsOn.ipynb
+++ b/notebooks/BoostHistogramHandsOn.ipynb
@@ -450,21 +450,14 @@
    "outputs": [],
    "source": [
     "def plothist2d(h):\n",
-    "    X, Y = h.axes.edges\n",
-    "    return plt.pcolormesh(X.T, Y.T, h.view().T)"
+    "    return plt.pcolormesh(*h.axes.edges.T, h.view().T)"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "This is transposed because it is ij indexed instead of xy. You can also do:\n",
-    "\n",
-    "```python\n",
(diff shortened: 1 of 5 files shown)


notebooks/BoostHistogramHandsOn.ipynb
```diff
@@ -450,21 +450,14 @@
...
    "metadata": {},
    "source": [
-    "This is transposed because it is ij indexed instead of xy. You can also do:\n",
-    "\n",
-    "```python\n",
-    "X, Y = h.axes.edges\n",
-    "X, Y = np.broadcast_arrays(X, Y)\n",
-    "return plt.pcolormesh(X, Y, h)\n",
-    "```"
+    "This is transposed because pcolormesh requires ij indexing instead of xy."
```
HDembinski, line 460: This is not a clear explanation. Better just to forward to the pcolormesh docs, https://matplotlib.org/3.2.2/api/_as_gen/matplotlib.pyplot.pcolormesh.html#axes-pcolormesh-grid-orientation

HDembinski:
Looks great! Minimal change with a lot of gain.

notebooks/BoostHistogramHandsOn.ipynb
henryiii: It doesn't help that it's wrong, either, I had ij and xy backwards. I'll just point at the docs...

PR #415, opened 2020-07-14
fix: nicer error message for indexing failures

henryiii:
Slightly more helpful if you mix up indices; this way you know a bit more about the str/int that was failing.


henryiii:
@HDembinski, just adding a few error messages, so going to merge - let me know if you have any concerns if I merge something small that you don't like, and I'll revert!

HDembinski:
I like helpful error messages, please go ahead. 👍

Issue #416, opened 2020-07-14
Division issues


henryiii:
From @HDembinski . Replacing

```python
h.view()[:] /= h.axes[0].widths * h.sum()
```

with

```python
h /= h.axes[0].widths * h.sum()
```

does not produce an error, but a weird looking wrong result.

And it says

```python
h[:] /= h.axes[0].widths * h.sum()
```

is not implemented yet.

PR #417, opened 2020-07-14
feat: fix operators on arrays, adds support for it

henryiii:
This refactors operator handling, allowing operations with an array to work #416, and unifies some of the operator handling. Histogram *op* Histogram is handled by BH, all others are handled by the view.

### DIFF
--- a/include/bh_python/register_histogram.hpp
+++ b/include/bh_python/register_histogram.hpp
@@ -87,18 +87,6 @@ auto register_histogram(py::module& m, const char* name, const char* desc) {
 
         ;
 
-    def_optionally(hist,
-                   bh::detail::has_operator_radd<histogram_t, double>{},
-                   py::self += double());
-
-    // Atomics for example do not support these operations
-    def_optionally(hist,
-                   bh::detail::has_operator_rmul<histogram_t, double>{},
-                   py::self *= double());
-    def_optionally(hist,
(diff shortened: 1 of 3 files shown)


HDembinski:
For the record, this is also a bug fix, an operation `h /= array` didn't fail and didn't produce the right answer. @henryiii I didn't understand why the variable axis had anything to do with this. Why does division work for Regular but not for Variable?

HDembinski:
I think this should be labeled as "bug-fix and feat" in the title.

henryiii:
> Why does division work for Regular but not for Variable?

Sorry, I didn't explain that, that that's really the point of the PR. 😊 

The problem is not the axes type, but what you are dividing by. `array` was `h.axes.widths`, which has a special property when the axis is Regular - they are all identical values. The problem was this was passing the array to the underlying implementation, and in the process, it took just the first value from that array, then applied it to the histogram. So `h[:] /= h.axes[0].widths * h.sum()` was really doing `h[:] /= h.axes[0].widths[0] * h.sum()` - which happens to be valid for Regular, but is not remotely valid for Variable.

I think you have to pick fix or feat. feat corresponds to a the middle number of a semantic version, and fix corresponds to the final number. This adds a feature, so I was thinking feat was correct. I'll add fix in the non-conventionalcommits part, though, since it is also a fix.

PR #420, opened 2020-07-15
refactor: use custom boolean, faster compile, simpler fill

henryiii:
Suggested by @HDembinski. This should restore the compile times, make it easier to compile `register_histograms.cpp` in under 10 minutes on simulated ARM reliably, and still keeps part of the performance boost by not having state in the axis type.

### DIFF
--- a/tests/test_histogram.py
+++ b/tests/test_histogram.py
@@ -667,6 +667,14 @@ def test_pickle_1():
     assert a == b
 
 
+def test_fill_bool_not_bool():
+    a = bh.Histogram(bh.axis.Boolean())
+
+    a.fill([0, 1, 1, 7, -3])
+
+    assert_array_equal(a.view(), [1, 2])
+
+
 def test_pick_bool():
     a = bh.Histogram(bh.axis.Boolean(), bh.axis.Boolean(metadata={"one": 1}))
 
--- a/tests/test_histogram.py
+++ b/tests/test_histogram.py
@@ -672,7 +672,7 @@ def test_fill_bool_not_bool():
 
     a.fill([0, 1, 1, 7, -3])
 
-    assert_array_equal(a.view(), [1, 2])
+    assert_array_equal(a.view(), [1, 4])
 
 
 def test_pick_bool():
--- a/include/bh_python/axis.hpp
+++ b/include/bh_python/axis.hpp
@@ -112,13 +112,16 @@ class boolean : public bh::axis::integer<int, metadata_t, option::none_t> {
   public:
     explicit boolean(metadata_t meta = {})
         : integer(0, 2, std::move(meta)) {}
+    explicit boolean(const boolean& src,
+                     bh::axis::index_type begin,
+                     bh::axis::index_type end,
+                     unsigned merge)
+        : integer(src, begin, end, merge) {}
     boolean(const boolean& other)
         : integer(other) {}
     bh::axis::index_type index(int x) const noexcept {
-        return static_cast<bh::axis::index_type>(x != 0);
+        return integer::index(x == 0 ? 0 : 1);
     }
-    int value(bh::axis::index_type i) const noexcept { return static_cast<int>(i); }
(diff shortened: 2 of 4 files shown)


tests/test_histogram.py
```diff
@@ -667,6 +667,14 @@ def test_pickle_1():
     assert a == b
 
 
+def test_fill_bool_not_bool():
+    a = bh.Histogram(bh.axis.Boolean())
+
+    a.fill([0, 1, 1, 7, -3])
+
+    assert_array_equal(a.view(), [1, 2])
```
henryiii, line 675: @HDembinski Actually, what should this do for non-zero values? Should this follow Boolean logic, and non-zero is True, or should we make it just like `Integer(0, 2)`, as done here? I'm leaning toward the former, as it's consistent with other booleans. So the result here would be `[1, 4]` if I made that change. I didn't have this test on the other axes, but I expect it would have produced [1, 4] or would have failed with a type error.

henryiii: I just checked the current develop, this test produces `[1, 4]`. So will change this PR to match. That means there was (and still will be) an advantage to using `Boolean` over `Integer(0,2, flow...)`

henryiii:
5m 39s to 5m 11s for the wheel build on Ubuntu. (On CI, so rather variable in timing, but hopefully okay for a rough idea).

HDembinski:
The savings are not as dramatic as your initial post suggested. I would have been surprised if they were. Why don't you implement this in pure Python?

HDembinski:
Otherwise it looks good. Just a comment: You could have lifted the type restriction in other ways, too, by converting the arrays in the fill method to the value types of the corresponding axes.

henryiii:
We have to have this in C++ because we can take a C++ histogram that is returned from from project, reduce, add, and anywhere else a new histogram is returned and reconstruct the Python object from it. Given that design, we must have a 1:1 mapping between C++ axes and Python axes. Since a `Boolean` is not at all the same as a `Integer(0,2, underflow=False, overflow=False)` (it handles all numbers that are not 0 like they are 1),  it's even more important that Booleans don't cast into Integer. (This is also how subclasses, like Hist and hopefully more, are allowed to subclass Axes - when casting, boost-histogram tries to match the axis to the histogram "family"; if a package provides a new family, `h[...]` can return a new histogram with the subclassed axes, instead of the parent axes).

The alternative would be to teach Python about how types map through all possible operations. And if we ever have operations that change the C++ type (like dropping flow bins), it would be even more challenging to get right. This 1:1 mapping system is simpler, less error-prone, and future-proof.

> Converting arrays

This would be slower than just allowing the integers to be correctly processed by the boolean itself, and might require more memory by creating an extra copy (this takes ints, so we often do 1 copy already).

henryiii:
It would be nice to remove the other workaround for the old boolean axes - the reducible issue. This really can be reduced to only two things, all true or all false. I assume I need to implement the special constructor; will look to see if there's documentation about making a reducible axis

henryiii:
What did my initial post suggest? I think the only thing I said was about ARM compilation. The current version fails 1/3 of the time by timing out at 10 minutes; this is about 30 seconds faster on x86 so should be 1 minute or so on simulated ARM, which would make this compile much more reliably.

henryiii:
I've made this reducible and dropped the workaround.

henryiii:
Nice, we now support reduction on a boolean axes! `h[:1]` or `h[1:]` works properly, and you get a `Boolean(<True>)` or `Boolean(<False>)` in the repr. :) We don't provide a way to construct one, as I think they are _almost_ only for consistency in slicing.

Almost being if you wanted a histogram that only saved values when something was either True or False, and didn't have memory storage for the other, I could see this being useful, I suppose. In 1.0 we will support slices on axes, so `bh.axis.Boolean()[:1]` should be the method to get one of these axis without histogram slicing, I think.

henryiii:
Why is this PR not showing all the changes? That's odd.

PR #422, opened 2020-07-15
refactor: boolean axis with reduce and int fill

henryiii:
PR #420 was not updating, so closing and reopening as new PR.


henryiii:
@HDembinski can you take a quick look at this reworked version, especially the C++ code? Centers was not working before, and now I've added reductions so that the old workarounds in Python can be dropped. I think this is good now. We can't add inclusive, since a sliced axis is not inclusive. But we still avoid extra copies or modifications to the source array, so it should be more efficient than manipulating the fill arrays in pure Python.

Issue #431, opened 2020-07-21
Multiplication by a scalar isn't working


beojan:
Multiplication by a scalar isn't working, because a deep copy of the histogram cannot be made:
```
----> 1 2*h2

/usr/lib/python3.8/site-packages/boost_histogram/_internal/hist.py in __rmul__(self, other)
    211
    212     def __rmul__(self, other):
--> 213         return self * other
    214
    215     def __truediv__(self, other):

/usr/lib/python3.8/site-packages/boost_histogram/_internal/hist.py in __mul__(self, other)
    207     # If these fail, the underlying object throws the correct error
    208     def __mul__(self, other):
--> 209         result = self.copy()
    210         return result._compute_inplace_op("__imul__", other)
    211

/usr/lib/python3.8/site-packages/boost_histogram/_internal/hist.py in copy(self, **kwargs)
    500
    501         if deep:
--> 502             return copy.deepcopy(self)
    503         else:
    504             return copy.copy(self)

/usr/lib/python3.8/copy.py in deepcopy(x, memo, _nil)
    151             copier = getattr(x, "__deepcopy__", None)
    152             if copier is not None:
--> 153                 y = copier(memo)
    154             else:
    155                 reductor = dispatch_table.get(cls)

/usr/lib/python3.8/site-packages/boost_histogram/_internal/hist.py in __deepcopy__(self, memo)
    377     def __deepcopy__(self, memo):
    378         other = self.__class__.__new__(self.__class__)
--> 379         other._hist = copy.deepcopy(self._hist, memo)
    380         other.metadata = copy.deepcopy(self.metadata, memo)
    381         other.axes = other._generate_axes_()

/usr/lib/python3.8/copy.py in deepcopy(x, memo, _nil)
    151             copier = getattr(x, "__deepcopy__", None)
    152             if copier is not None:
--> 153                 y = copier(memo)
    154             else:
    155                 reductor = dispatch_table.get(cls)

/usr/lib/python3.8/copy.py in deepcopy(x, memo, _nil)
    144     copier = _deepcopy_dispatch.get(cls)
    145     if copier is not None:
--> 146         y = copier(x, memo)
    147     else:
    148         if issubclass(cls, type):

/usr/lib/python3.8/copy.py in _deepcopy_dict(x, memo, deepcopy)
    228     memo[id(x)] = y
    229     for key, value in x.items():
--> 230         y[deepcopy(key, memo)] = deepcopy(value, memo)
    231     return y
[...]

PR #432, opened 2020-07-23
fix: view ops correctly supported on WeightedSum View

henryiii:
Fix for comment in #431 by @colizz; expands NEP 13 support for WeighedSum View. Scalars like `WeighedSum()` are now supported too!

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -1,0 +1,82 @@
+# -*- coding: utf-8 -*-
+import boost_histogram as bh
+from numpy.testing import assert_allclose
+import pytest
+
+
+@pytest.fixture
+def v():
+    h = bh.Histogram(bh.axis.Integer(0, 4), storage=bh.storage.Weight())
+    h.fill([1, 1, 1, 2, 2, 3])
+    return h.view()
+
+
+def test_basic_view(v):
+    assert_allclose(v.value, [0, 3, 2, 1])
+    assert_allclose(v.variance, [0, 3, 2, 1])
+
+
+def test_view_mul(v):
+    v2 = v * 2
+    assert_allclose(v2.value, [0, 6, 4, 2])
+    assert_allclose(v2.variance, [0, 6, 4, 2])
+
+    v2 = v * (-2)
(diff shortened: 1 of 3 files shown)


henryiii:
@HDembinski and @jpivarski, would you mind checking over my PR to see if the NEP 13 code looks reasonable and the calculations are reasonable?

jpivarski:
It looks good to me, though you're only overriding a few operations, like `+`, `-`, `*`, `/`, `//`, right? In principle, the same effect could be obtained by overriding `__add__`, `__radd__`, etc. I imagine someone might want to `np.sqrt` or `np.exp` the values in the bins, though it's unclear what this should do to the variances unless you know the derivatives of all functions. (And that might be desired as well; see #gradhep in IRIS-HEP Slack.) Perhaps unknown ufuncs (the main reason for NEP13) would apply to the values and drop the variances? Even though this is a lossy operation, I can see people wanting it in the midst of an analysis. Consider how often we write loops over bin contents, applying arbitrary functions...

Lossy operations for unrecognized ufuncs are particularly good if they're not in-place: then you still have the original variances in another object to decide what to do with them.

henryiii:
I would rather the drop be explicit, that is, either `view.values` is used, or if there's a `.values()` method of some sort added, that be used. We can always add more functionality later, since it still throws an error if it's not in this list. But I should make this match via `dtype`, not by `__class__` (will fix).

tests/test_views.py
```diff
@@ -0,0 +1,100 @@
...
+    assert_allclose(v.variance * 2, v2.variance)
+
+    v2 = v + v[1]
+    assert_allclose(v.value + 3, v2.value)
+    assert_allclose(v.variance + 3, v2.variance)
+
+    v2 = v + bh.accumulators.WeightedSum(5, 6)
+    assert_allclose(v.value + 5, v2.value)
+    assert_allclose(v.variance + 6, v2.variance)
+
```
HDembinski, line 98: Perhaps also test radd here?

henryiii: Good call! Wasn't working. 😳

PR #433, opened 2020-07-23
fix: ops do shallow copy

henryiii:
Fix for actual issue in #431. (still need info from @jpivarski as to why a lock is in Histogram metadata, but now simple operations do shallow copies, which was the original intent).

### DIFF
--- a/src/boost_histogram/_internal/hist.py
+++ b/src/boost_histogram/_internal/hist.py
@@ -182,7 +182,7 @@ class Histogram(object):
         return self.view(False)
 
     def __add__(self, other):
-        result = self.copy()
+        result = self.copy(deep=False)
         return result.__iadd__(other)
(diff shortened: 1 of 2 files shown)


jpivarski:
I just caught this: there's a lock in histogram metadata? Which metadata? (Aghast or ROOT?)

The ROOT objects point back to the file they came from (there are a thousand good reasons for that—I'm glad I'm doing it now), and that's something that shouldn't be deeply copied. More to the point, I should define `__deepcopy__` in them in such a way as to break that connection (they're still valid if the `file` property is `None`). It's possible that there's a lock in the `source` within a `file`, but there's a thread pool executor and open file handles, too.

Perhaps the right thing to do is to break that connection before handing them off to boost-histogram/hist.

Issue #437, opened 2020-07-29
fill() prints histogram in 0.10.1


alexander-held:
Hi, since version 0.10.1, calling a `.fill()` command on a `boost_histogram.Histogram` prints the histogram:

```
import boost_histogram as bh

hist = bh.Histogram(bh.axis.Variable([1,2]))
hist.fill([1])
```
prints
```
(array([1]),) None None
```
while in version 0.10.0 there is no output.

---

environment: running via `python:3.8-slim bash` via docker:
```
boost-histogram==0.10.1
numpy==1.19.1
```

henryiii:
Not again! I left a debug output in 0.7.0 too. I will see if I can set up a pre-commit check that looks for print statements.

henryiii:
Here it is:

https://github.com/scikit-hep/boost-histogram/blob/e79ff55a2ca0be241dd508e55f78b26e1d881867/src/boost_histogram/_internal/hist.py#L295

I ignored the print output last time, because it was in a lesser-used function, but this one probably needs a patch sooner rather than later.

henryiii:
(This has been fixed in 0.10.2)

PR #438, opened 2020-07-29
fix: drop fill print statement

henryiii:
For #437. Won't close that issue until 0.10.2. @HDembinski should we push a quick 0.10.2 with just this fix? Or should we wait a little to see if anything else comes up? The printing when filling is really irritating (though it is just an irritant).

I'm still looking into this to see if there's a good way to detect this in tests or pre-commit for the future (0.7.0 had the same issue in a different place). Edit: Now we have CI and pre-commit ensuring this never happens again!

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
   hooks:
   - id: flake8
     exclude: docs/conf.py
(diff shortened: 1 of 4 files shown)


HDembinski:
Sure, it warrants a minor release.

Issue #430, opened 2020-07-17
name/title system


henryiii:
I'd like to propose upstreaming @LovelyBuggies fantastic title/name system from Hist to boost-histogram. There are several reasons for this:

* It has no dependencies
* It can take part in #423 in providing a consistent API for uproot and mplhep (maybe others) (`mplhep.plothist(uproot.to_boost())` would show titles too, like Hist, and @jpivarski probably could add the same api directly to TH* objects in Uproot4)
* It can make dict-indexing, projections, and axes read more clearly if names are used, allowing a high level of clarity as to which axis is being referenced.
* It is completely optional, and does not affect the current usage at all
* It would make conversions to/from hist easier, since currently names and titles get packed into a dict, which then has to be changed going between hist/boost-histogram (especially for uproot, since title and name have to be handled during the conversion)
* We can optimize the storage internally, just keeping the strings directly in the metadata type, rather than storing python dicts and python strings

Quick example of usage:

```python
h = bh.Histogram(
    bh.axis.Regular(10, -1, 1, name="x"),
    bh.axis.Regular(20, -2, 2, name="y"),
)
h.fill(x=np.random.normal(size=1_000_000), y=np.random.normal(size=1_000_000))

h.project("y");

h[{"x": sum, "y": bh.rebin(2)}]

h.axes["x"].label = "x [μm]"
h.axes["y"].label = "y [cm]"
mplhep.hist2dplot(h);
```

The proposed design, open for discussion:

* Axes gain a new `name` parameter.
    * This probably should be a read-only parameter (which is not how Hist currently works)
    * When histograms are created, they make sure all names are ether empty or unique
    * Names are used anywhere axis numbers can be used.
        * Empty-name axes must be accessed by number.
* Axes gain a new `label` parameter
   * Will override the name for plotting libraries, so titles can be attached to the histogram rather than added afterwords in the plot.
    * If the label is not set, the name is used if available.
* The metadata slot in boost-histogram's C++ becomes a struct, with two strings and an arbitrary Python object.

@HDembinski, what do you think?

HDembinski:
It seems very useful and rather consistent. The use of names seems Pythonic. My only complaint is use of title instead of label. I would prefer label.

Some thoughts: why not go one step further and allow an axis named x to be accessed as h.axes.x?

henryiii:
[...]

PR #450, opened 2020-09-02
feat: arbitrary properties

henryiii:
This allows the classes to either be or behave like dict classes. See #430. Picking and accessing `.metadata` is backwards compatible.


henryiii:
@HDembinski This was your idea, but does the implementation look okay? Actually axis a true `__dict__` class doesn't work, due to specialized `__dict__` handling (not a true dict) and having to put the `_ax` in `__dict__`; but this imitates it for axes and uses `__dict__` for Histograms. It is backward compatible with `metadata` as well, unpicking moves it to the new format.

src/boost_histogram/_internal/axis.py
```diff
@@ -37,6 +37,31 @@ def __copy__(self):
...
+        if item == "_ax":
+            Axis.__dict__[item].__set__(self, value)
+            return
+        elif self._ax.metadata is None:
+            self._ax.metadata = {}
+
+        self._ax.metadata[item] = value
+
+    def __dir__(self):
+        metadata = list(self._ax.metadata) if self.ax.metadata is not None else []
```
HDembinski, line 62: That only works if `self._ax.metadata` is a dict, right? What if metadata is a number or a string?

src/boost_histogram/_internal/axis.py
```diff
@@ -37,6 +37,31 @@ def __copy__(self):
         other._ax = copy.copy(self._ax)
         return other
 
+    def __getattr__(self, item):
+        if item == "_ax":
+            return Axis.__dict__[item].__get__(self)
+        elif self._ax.metadata is not None and item in self._ax.metadata:
```
HDembinski, line 43: Does it make sense to init `self._ax.metadata` as an empty dict, to save this check for every attribute access?

src/boost_histogram/_internal/axis.py
HDembinski: I see below that you remove the possibility to set `metadata` by hand, so perhaps this is ok then, but it should be made sure that `metadata` is always a dict.

src/boost_histogram/_internal/axis.py
```diff
@@ -525,12 +531,14 @@ def __init__(self, categories, **kwargs):
...
         if options == {"growth"}:
-            self._ax = ca.category_str_growth(tuple(categories), metadata)
+            self._ax = ca.category_str_growth(tuple(categories))
         elif options == set():
-            self._ax = ca.category_str(tuple(categories), metadata)
+            self._ax = ca.category_str(tuple(categories))
         else:
             raise KeyError("Unsupported collection of options")
 
+        self.metadata = metadata
```
HDembinski, line 540: It looks like I can still set metadata to any type here. I am personally fine with removing the metadata keyword from init, it is a breaking change, however.

src/boost_histogram/_internal/hist.py
```diff
@@ -162,12 +159,24 @@ def _generate_axes_(self):
 
         return AxesTuple(self._axis(i) for i in range(self.ndim))
 
-    def _new_hist(self, _hist):
+    def _new_hist(self, _hist, memo=NOTHING):
```
HDembinski, line 162: Not sure what this does...

src/boost_histogram/_internal/hist.py
```diff
@@ -376,24 +385,38 @@ def _reduce(self, *args):
...
+        """
+        Version 0.8: metadata added
+        Version 0.11: version added and set to 0. metadata/_hist replaced with dict.
+
+        ``dict`` contains __dict__ without "axes" and "_hist"
+        """
+        local_dict = copy.copy(self.__dict__)
+        del local_dict["axes"]
+        state = {"dict": local_dict, "version": 0}
         return state
```
HDembinski, line 403: state should be a tuple rather than a dict.

src/boost_histogram/_internal/hist.py
```diff
@@ -376,24 +385,38 @@ def _reduce(self, *args):
...
 
     def __setstate__(self, state):
-        self._hist = state["_hist"]
-        self.metadata = state.get("metadata", None)
-        self.axes = self._generate_axes_()
+        if "version" not in state:
```
HDembinski, line 406: If state is a tuple, you can place the new version number at the front of the tuple. To discriminate between old and new, you can check whether `state` is a dict or a tuple.

src/boost_histogram/_internal/axis.py
```diff
@@ -444,24 +454,22 @@ def __init__(self, start, stop, **kwargs):
...
+            self._ax = ca.integer_oflow(start, stop)
+        elif "circular" in options and "growth" not in options:
+            self._ax = ca.integer_circular(start, stop)
         elif options == set():
-            self._ax = ca.integer_none(start, stop, metadata)
+            self._ax = ca.integer_none(start, stop)
         else:
             raise KeyError("Unsupported collection of options")
 
+        self.metadata = metadata
```
HDembinski, line 471: It looks like you are not passing metadata to the cpp instance anymore. Is that correct? If cpp code cannot see metadata, it cannot pass it on in methods like `.project` or the wrapper over `reduce`. `operator==` also compare metadata.

HDembinski: In am not against a solution that handles metadata completely on the Python side and removes metadata_t in C++, but then the python interface over project and reduce, operator== etc need to handle metadata explicitly (before it was handled automatically by the underlying cpp implementation).

src/boost_histogram/_internal/axis.py
henryiii: Yes, `self._ax.metadata` is basically the equivalent of `self.__dict__` without `self._ax`. `self.metadata` is just an item in that dict, with the (backwards compatible) feature that it defaults to None if not present.

src/boost_histogram/_internal/axis.py
henryiii: I was trying to be backward compatible, but I think that's actually handled by setstate/getstate properly, so probably yes.

src/boost_histogram/_internal/axis.py
henryiii: I originally allowed both, but I think (as mentioned above) it now is truly always a dict, so I can simplify a bit! Metadata is allowed as a kwarg and is printed in the repr for backward compatibility, but we could remove that if you want. Since Boost.Histogram has a metadata slot, I think it's okay to allow easy access to ax.metadata this way?

src/boost_histogram/_internal/hist.py
henryiii: If you pass memo, it's a deep copy. (copy and deep copy were separate implementations before). I _think_ memo can be None if this is the outer level of deep copy, so I use a tag here that can't ever be passed as memo.

src/boost_histogram/_internal/hist.py
henryiii: Okay, can do that, is a bit simpler. I think dicts are common in Python for state, but that works too.

src/boost_histogram/_internal/axis.py
henryiii: That would be a large, heavy rewrite, and would replicate a lot of the logic for computing axes in boost-histogram, rather than just letting Boost.Histogram do the work. I'm not against it either, but I like the simplicity of using Boost.Histogram's computation (since it can handle this) instead of replicating logic.

src/boost_histogram/_internal/axis.py
HDembinski: I think you can change this without breaking backward compatibility, but perhaps a few unit tests could be added that check for possible corner cases?

src/boost_histogram/_internal/axis.py
HDembinski: How about this: all unrecognized kwargs are added to the dict? This means you can write `bh.axis.Regular(..., name="foo", title="bar")` and the old `bh.axis.Regular(..., metadata="foo")` would still work, but metadata is now just one of those unrecognized kwargs.

HDembinski: If a keyword conflicts with an axis method, the keyword needs to be rejected, however.

src/boost_histogram/_internal/axis.py
HDembinski: Ok, no need to rewrite things if things are working as they are now. Apparently I misunderstood.

Context: What I like about cpp not seeing the Python metadata is the clean border between the C++ world and the Python world. Life-time issues and (deep) copy issues are are more easily handled in this way.

src/boost_histogram/_internal/hist.py
HDembinski: It also allows for a clean cut between old and new state format.

src/boost_histogram/_internal/axis.py
henryiii: This is working well, and much simpler, I just am fixing some of the unit tests that checked a raw C++ class (the NumPy one) along with the wrapped axes, expecting them to have the same API.

src/boost_histogram/_internal/axis.py
henryiii: I don't think I like that idea, for three reasons: a) mixing keyword arguments and `**kwargs` is ugly; if you mistype a keyword argument (`grow` instead of `growth`, for example), instead of immediately complaining loudly, it silently works but does the wrong thing (saves `.grow=True` and doesn't activate growth); b) you can't add any keyword arguments later without changing the API (or in downstream libraries like Hist; if we add a `flow=True` shortcut that would cause two different behaviors), and c) changing axis classes will not cause unused keywords to throw an error, but will store them instead. For example, setting `circular=True` on a `IntCategory` would silently "work" (different form of point a).  You can also always turn this on later, but you can't remove it once added. The fact you could set title/name is a plus. Having a tight API on the upstream library is likely better?

src/boost_histogram/_internal/hist.py
henryiii: I can't provide references either, other than that the default state is the `__dict__`, which is, well, a dict ;). But tuples work well here for us, for the reasons you gave above.

henryiii: I think pybind11 heavily favors tuples.

henryiii:
@HDembinski, FYI, Boost.Histogram's axes complain if the default constructor for the metadata is explicit on MSVC only (`metadata_t meta = {}`).

HDembinski:
> @HDembinski, FYI, Boost.Histogram's axes complain if the default constructor for the metadata is explicit on MSVC only (`metadata_t meta = {}`).

Ok, noted, but I am not sure what you mean, could you point me to the compiler message?

henryiii:
Until it is deleted, it is here: https://github.com/scikit-hep/boost-histogram/runs/1084034405

Copied for posterity:

```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.27.29110\bin\HostX86\x64\cl.exe /c /nologo /Ox /W3 /GL /DNDEBUG /MD -Iinclude -Iextern/assert/include -Iextern/config/include -Iextern/core/include -Iextern/histogram/include -Iextern/mp11/include -Iextern/pybind11/include -Iextern/throw_exception/include -Iextern/variant2/include -Ic:\cibw\python\python.3.8.4\tools\include -Ic:\cibw\python\python.3.8.4\tools\include "-IC:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.27.29110\ATLMFC\include" "-IC:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.27.29110\include" "-IC:\Program Files (x86)\Windows Kits\NETFXSDK\4.8\include\um" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\ucrt" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\shared" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\um" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\winrt" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\cppwinrt" /EHsc /Tpsrc/register_axis.cpp /Fobuild\temp.win-amd64-3.8\Release\src/register_axis.obj /std:c++14 /EHsc /bigobj
  register_axis.cpp
  include\bh_python/regular_numpy.hpp(23): error C2512: 'metadata_t': no appropriate default constructor available
  include\bh_python/regular_numpy.hpp(23): note: Constructor for struct 'metadata_t' is declared 'explicit'
  include\bh_python/axis.hpp(113): error C2512: 'metadata_t': no appropriate default constructor available
  include\bh_python/axis.hpp(113): note: Constructor for struct 'metadata_t' is declared 'explicit'
  extern/histogram/include\boost/histogram/axis/regular.hpp(218): error C2512: 'metadata_t': no appropriate default constructor available
  extern/histogram/include\boost/histogram/axis/regular.hpp(217): note: Constructor for struct 'metadata_t' is declared 'explicit'
  error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Tools\\MSVC\\14.27.29110\\bin\\HostX86\\x64\\cl.exe' failed with exit status 2
  Building wheel for boost-histogram (PEP 517): finished with status 'error'
  ERROR: Failed building wheel for boost-histogram
Failed to build boost-histogram
ERROR: Failed to build one or more wheels
```

src/boost_histogram/_internal/axis.py
henryiii: I'll go ahead and merge this, since it's blocking further progress. Please let me know if you want **kwargs (or any other changes), and I'll get them in (in a separate PR).

PR #461, opened 2020-09-29
ci: fix recent test failure

henryiii:


henryiii:
Reported here: https://github.com/pytest-dev/pytest/issues/7814

I disabled and then reenabled the GHA plugin; that was not the issue.

PR #470, opened 2020-11-13
chore: updates

### LINKED ISSUE #468: Severe bug in axis.Variable
This should never possibly be happening:
```
import boost_histogram as bh
import numpy as np
bh.axis.Variable([1, 2, np.inf])
```
This produces `Variable([1, nan, inf])`

henryiii:
Includes a fix by updating Boost.Histogram, closes #468.

- docs: black formatting for notebooks
- feat: update to pybind11 2.6.1
- fix: update boost-histogram to fix inf in Variable issue
- ci: restore NumPy testing

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -17,9 +17,6 @@ on:
 
 
 env:
-  CIBW_ENVIRONMENT_MACOS: PIP_ONLY_BINARY=numpy
-  CIBW_TEST_EXTRAS: test
-  CIBW_TEST_COMMAND: "pytest {project}/tests"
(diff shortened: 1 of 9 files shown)


henryiii:
Note on status: testing on PyPy is slightly broken, causing the full wheel build to fail now that we include PyPy. Once that is fixed, this will be ready to merge. (A test on PyPy2 passes that is supposed to fail on Python 2 - it seems PyPy2 can load Python 3 pickles)

henryiii:
Okay, finally wheel builds are passing again. Tests remain deactivated for PyPy, and we no longer build manylinux1 wheels for pypy, as that's probably a bit of a waste, manylinux2010 is just fine. (Reminder: we build modified manylinux1 wheels and proper manylinux2010 wheels; most packages do not need both, but instead just have the lowest they are comfortable with).

PR #474, opened 2020-11-19
ci: use GHA for alternate archs

henryiii:
Travis has become too hard to work with. Emulate ARM using GHA and cibuildwheel 1.8's support for that, and note that PowerPC (was available before) & IBM-Z are available if requested - it's literally just a one line change, but don't want to waste CI resources if no one is using them. Also uses the new CIBW_TEST_SKIP to allow a few of the PyPy wheels to be tested. NumPy now provides ARM wheels so those are fully tested.

Generated a few very minor fixes to cibuildwheel as well to make mistakes a little more nicely handled, based on my mistakes here. :) https://github.com/joerick/cibuildwheel/pull/588 , https://github.com/joerick/cibuildwheel/pull/589

I'm also revisiting the support matrix slightly. `manylinux1` for Python 3.9 has been dropped, just like NumPy dropped it. I added Universal2 macOS to the README, and dropped the statement about requesting Conda + Python 2; pretty sure we are well beyond where a Conda 2.7 build could be requested.


henryiii:
NumPy doesn't provide a aarch64 wheel for Python 3.5, which is why the 3.5 job is so slow.

henryiii:
This is failing because I don't have TEST_SKIP and tests are on, but otherwise it works.

.github/workflows/wheels.yml
```diff
@@ -18,13 +18,16 @@ on:
       - published
 
 env:
-  CIBW_ENVIRONMENT: "PIP_ONLY_BINARY=numpy SETUPTOOLS_SCM_PRETEND_VERSION=${{ github.event.inputs.overrideVersion }}"
   SETUPTOOLS_SCM_PRETEND_VERSION: ${{ github.event.inputs.overrideVersion }}
+  CIBW_ENVIRONMENT: "PIP_ONLY_BINARY=numpy SETUPTOOLS_SCM_PRETEND_VERSION=${{ github.event.inputs.overrideVersion }}"
+  CIBW_TEST_EXTRAS: test
+  CIBW_TEST_COMMAND: "pytest {project}/tests"
+  CIBW_TEST_SKIP: "pp*macos* *universal2:arm64"
```
henryiii, line 25:
```suggestion
  CIBW_TEST_SKIP: "pp*macos* pp*win* pp27-* *universal2:arm64"
```

PR #475, opened 2020-11-25
feat: settable accumulator storage views

henryiii:
The view support has been refactored to allow setting with arrays on all storages, rather than just Weight() as implemented in #375. This also moves the code for supporting this directly to the view, simplifying the histogram and avoiding custom checks for storages. This uses the actual constructor from Boost.Histogram, so it avoids the internal Boost.Histogram data structure being required/exposed for Uproot 4 making TProfiles. Also fixed a bug with a missing `const` that kept the WeightedMean conversion from working for non-scalars (non-const cannot be vectorized).

Note that the histogram is now directly settable too, since it sends the request on to the view. This was manually patched in for Weight storage, but now it is properly implemented for all storages via the View.

Will be very useful in Uproot, as it has to set this when converting from ROOT.

Examples:

```python
h = bh.Histogram(bh.axis.Integer(0, 2), storage=bh.storage.Mean())
h[...] = [[3,2,1], [3,2,1]] # Sets count=3, value=2, variance=1 for both cells
h.view()[...] = [[3,2,1], [3,2,1]] # Also works on views

# Works with WeighedMean too:
# Assume w is a WeighedMean View
w[0] = [9, 1, 2, 3]
assert w.sum_of_weights[0] == 9
assert w[0].sum_of_weights_squared == 1
assert w.value[0] == 2
assert w[0].variance == 3
```

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,31 @@
 # What's new in boost-histogram
 
+# Version 0.12
+
+### Version 0.12.0 (WIP)
+
+Pressing forward to 1.0.
+
+
+#### User changes
+
+* You can now set all complex storages, either on a Histogram or a View with an (N+1)D array [#475][]
+
+#### Bug fixes
+
+* Fixed issue if final bin of Variable histogram was infinite by updating to Boost 1.75 [#470][]
+* NumPy arrays can be used for weights in `bh.numpy` [#472][]
+
+#### Developer changes
+
+* Bumped to pybind11 2.6.1 [#470][]
+* Black formatting used in notebooks too [#470][]
+
+[#470]: https://github.com/scikit-hep/boost-histogram/pull/470
+[#472]: https://github.com/scikit-hep/boost-histogram/pull/472
+[#475]: https://github.com/scikit-hep/boost-histogram/pull/475
+
+
 ## Version 0.11
 
 ### Version 0.11.1
--- a/src/boost_histogram/_internal/hist.py
+++ b/src/boost_histogram/_internal/hist.py
@@ -708,26 +708,26 @@ class Histogram(object):
         value = np.asarray(value)
         view = self.view(flow=True)
 
-        # Shortcut: allow raw arrays for WeightedSum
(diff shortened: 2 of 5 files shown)


henryiii:
@jpivarski This is unlikely to land in time for Uproot4, so I'll make a PR today that manually sets the internal data structure of boost-histogram's WeightedMean, and then after the boost-histogram next release, I'll make a PR that bumps the minimum version required and drops the usage of the internal structure.

@HDembinski, I still need to add tests, but does this look reasonable? Basically, view[A, B, C] = array[A, B, C, N] where N is 2, 3, or 4 will set Weight, Mean, or WeightedMean respectively, using the constructor arguments (and histograms just use the view setter). If someone _really_ wants to set `_sum_of_weighted_deltas_squared`, they can still dig into the dtype and build it manually, just like we do when unpicking.

HDembinski:
I haven't looked into the implementation of to_boost() but just going by that name, I would expect it to provide a boost-histogram already.

henryiii:
Any boost-histogram subclass (like Hist) can be called as `hist.Hist(uproot_hist)` and it calls `self.__class__(uproot_hist.to_boost())`. (All boost-histogram subclasses already support casting in this way) We could instead make a hidden api, like `_boost_histogram_`, so that a library that supports conversion to a histogram doesn't have to add a public method.

HDembinski:
Ok, that makes sense.

HDembinski (commented):
I trust you on the implementation, looks good to me.

HDembinski:
The nitpick I have left is that the title suggests that this patch adds settable views, but it looks like it is doing way more than that.

henryiii:
It's doing two things - I really should pull it apart into two PRs. I'm in the HSF training hackathon until Friday, then I'll be focusing on getting these PRs updated (adding tests and such).

henryiii:
I pulled out the to_boost part, to make sure it gets a proper discussion and name if needed; so this is now just settable views (and histograms now can be set with Mean and WeightedMean storages too, since they now just defer to the view when setting with a N+1 storage of the right shape).

PR #476, opened 2020-11-25
feat: PlottableProtocol support

### LINKED ISSUE #423: DRAFT: Plotting (and later fitting) Protocol
I'm working on defining a Protocol for plotting histograms. To do so, I need some way to access the values and optionally variances so that mplhep and others can decide on what to plot. Here is one possible suggestion:

```python
from numpy.typing import ArrayLike # requires NumPy master
from typing import Protocol, Optional, Tuple, Union, Iterable

class PlottableAxis(Protocol):
    label: str # May be removed soon

    edges: ArrayLike

     # For non-categorical axes, this returns None
    categories: Union[Iterable[int], Iterable[str], None]

class PlottableHistogram(Protocol):
    axes : Tuple[PlottableAxis]
   
    # Values returns the array or the values array for specialized accumulators
    def values(self, flow=False) -> ArrayLike: ...

    # Variance returns the variance if applicable, otherwise None
    # If counts is none, variance returns NaN for that cell (mean storages)
    def variances(self, flow=False) -> Optional[ArrayLike]: ...
```

We can look at labeling in a later draft, but there are a few key points:

* `.values()` is like `view()` for simple storages, and like `.view().value` for complex ones, providing plotting libraries a consistent way to access the central values for plots.
* `.variances()` returns None if a storage does not provide information for variances, allowing a plotting library to use `if variances := h.variances():` to skip plotting error bars, or do something reasonable. If it exists for that storage, it returns an array.
* `.axes[I].categories` is an array of labels if this is a Category storage, otherwise it is None.

I didn't add `.counts()`, but maybe that should be included too?

This is assuming the master object *is* the API. You could also have a `__histogram_dict__` that returns something like what you see above - then it's much easier to add to existing libraries like Physt and doesn't affect/change the public API, but on the flip side, it's duplicated, all or nothing, and doesn't provide API consistency for users. (Double underscores for clarity - you really aren't supposed to add anything with double underscores, so for an API choice, it might not be ideal).

PS: This is post 0.10.0, possibly post 1.0, so no rush here, feel free to come up with something better.

PPS: Initially, boost-histogram and Uproot4 would implement this, and Hist would get it for free since it is a boost-histogram.

henryiii:
Adds PlottableProtocol support to boost-histogram.

See the Uhi repo, such as this PR: https://github.com/henryiii/uhi/pull/2 and later updates.

TODO:

- [x] Return None for variance if a simple storage is set with weights or manually changed via `h[...] = `.
- [x] Change tests to use traits instead of options
- [x] Add to changelog, with upgrade warning (options -> traits)
- [x] Add to docs
- [x] Check existing docs to see if there are some places to show `.values()` and such instead of `.view()`.

We might want to loosen the protocol slightly to specify a Sequence is returned for `__getitem__` on non-discrete axis; it does not have to be exactly a Tuple. Specifically, something like uproot that does not support non-discrete axis could make this a NumPy 2D array with a few extra properties, and that should still pass the Protocol. (With a mix of axis types, this would be still representable by an AwkwardArray, technically (edit: except for needing to flatten the 1 element return).

We should also ideally do a static check against typing/plottable.py, but that likely will wait till we have more types and the UHI repo gets a few other useful bits and moves to scikit-hep. I've vendored in the Protocol for the docs for now.

Closes #423.

TODO 2:
- [x] check and match uproot for mean variance
- [x] Add docs cross links
- [x] Invalidate weights if inplace operations are used


HDembinski:
It looks like you first add properties directly on interface layer of the axis class and then wrap that data in a Option object on the Python layer on top of the interface. Wouldn't it be easier to make the Option object directly wrap the integer returned by `get_options`?

henryiii:
The point of the changes (other than trying to simplify the C++ class sitting on top of a single unsigned value that had to have a pickle function, copy/deepcopy, etc.) was to make "discrete" available in the .options structure (which is now probably closer to `traits` than `options`?). Do you like these traits bundled in with "options" here? In Python, you can't set any of these, so they all are really just output "traits" and options is a way to organize them.

A few ideas (referring to the Python Histogram, since the rest is internal detail that we can change):

1. Remove `.options` and just make these properties on `.axis` (and change the Protocol, we don't have much more time to make changes, as Uproot is released Monday, but this is small and can be done easily). Reduces the available namespace on the axis a bit, but do you really want a user to be able to add `ax.growth = ...`?
2. Keep `.options` but add traits directly on the axis. Provide `.circular` also on the axis, because only having `.options.circular` in the protocol seems really strange.
3. Inject just `.discrete` into `.options` - this was hard to do in the current state, as is messed with pickling. Also it is nice to be able to ask all the Boost.Histogram trait questions, even if they are not part of the protocol?
4. Bundle all traits into `.options` (the current state of the PR's second commit). I included all three of the Boost.Histogram traits.
5. Put them in `.traits`. Or `.properties`.
6. Put them in `.traits.options`. This is the closest to Boost.Histogram, but is probalby too structured for an ideal Protocol, so probably needs to be mixed with one of the choices above for the Protocol.

I'm pretty open to any direction, just started playing with it last night. I originally liked 4, though I like 1 better than I did originally in the Protocol discussion. @jpivarski didn't really like the name "options", since he thought that sounded like input options, rather than "properties" that you query. (In Boost.Histogram, of course, they are something you can input, while they are preset in boost-histogram).

Thoughts?

jpivarski:
If we're voting, I'd vote for `circular` and `discrete` to be properties directly on the axis, without an intermediate object. I don't think the number of these is going to grow so much that we regret not putting it in a separate namespace, and the separate namespace is cumbersome and makes it just a little harder to find these properties.

We can also drop `"median"`. It was just an indication of where this might grow, but it doesn't need to be defined into there's a need for it. But as you pointed out, these need to be strings or some very basic type. An implementation could return enumerations that are equal to the strings they represent.

henryiii:
There are two things - the Protocol, and boost-histogram. I think for the Protocol @HDembinski was neutral on the nesting, I was for but have changed to be against, and you are against - so let's remove .options from the Protocol. Then Hans can pick whatever direction he'd like to go with boost-histogram. Probably option 1 or 2 as listed above?

HDembinski:
I was neutral on options, but they have been growing on me. The options all logically belong together, so it makes sense to have them grouped. It also makes tab-completion on the axis less cluttered. Only library developers will query options. Normal users who write scripts won't have the need to query them, they know what options they have set. Since they are not queried frequently, they can and should sit deeper in the interface hierarchy.

HDembinski:
> The point of the changes in the second commit (other than trying to simplify the C++ class sitting on top of a single unsigned value that had to have a pickle function, copy/deepcopy, etc.) was to make "discrete" available in the .options structure (which is now probably closer to `traits` than `options`?). Do you like these traits bundled in with "options" here? In Python, you can't set any of these, so they all are really just output "traits" and options is a way to organize them.

In C++, `discrete` is a trait that the compiler can figure out at compile-time by inspecting the axis interface. In Python it makes sense to have it .options for the lack of a better place.

>     4. Bundle all traits into `.options` (the current state of the PR's second commit). I included all three of the Boost.Histogram traits.

This is my preferred compromise between exactness and convenience. I think we wouldn't even worry about the distinction between options and traits if these were not technically implemented differently in C++. For Python, the distinction seems irrelevant, so grouping them all in .options, which describes what the axis can do, makes sense.

HDembinski:
I would prefer to not to discuss things here and in the uhi repo simultaneously.

henryiii:
Okay, that makes sense - we'll go with option 4 then. Thanks! Don't worry about answering twice, I can update UHI's proposal from the discussion here.

henryiii:
The `__dict__` in `__slots__` trick is perfect for wrapping a C++ class. You can keep the wrapped item in `__slots__`, so it's not in your `__dict__`. Used that to simplify the Histogram now too. Finished the rest of the items, I think this is ready!

Reminder for anyone watching this - there are a few more items to work on before the next release, and I want to help get mplhep and histoprint ready for this before or around the time of a new release.

henryiii:
@HDembinski feel free to check it over! Check the docs changes, for example.

In the current state, the _only_ think that invalidates a simple storage's variance is a weighed fill. Specifically, there are three ways someone can set values:

```python
h.fill(1, weight=0.5)    # 1. Should invalidate .variances()
h[...] = array           # 2. ?
h.view()[...] = array    # 3. Cannot invalidate .variances()
```

We can't control option 3, but that's not an issue - if someone messes with the view, they are playing with raw access to our memory anyway. The question is what to do for line 2 - I think for consistency, we should follow line 3; but it's a close call, very recently I would have said any set operation should invalidate variances as well. Happy to go either way if you have a preference on line 2.

For a library that produces histograms like uproot/coffea/etc (which is the most common place I would expect to see memory set), then you can always do a single weight 0 fill to invalidate the variances if you know the data you are loading has had lossy weighed fills on it.

HDembinski:
> @HDembinski feel free to check it over! Check the docs changes, for example.
> 
> In the current state, the _only_ thing that invalidates a simple storage's variance is a weighed fill. Specifically, there are three ways someone can set values:
> 
> ```python
> h.fill(1, weight=0.5)    # 1. Should invalidate .variances()
> h[...] = array           # 2. ?
> h.view()[...] = array    # 3. Cannot invalidate .variances()
> ```
> 
> We can't control option 3, but that's not an issue - if someone messes with the view, they are playing with raw access to our memory anyway. The question is what to do for line 2 - I think for consistency, we should follow line 3; but it's a close call, very recently I would have said any set operation should invalidate variances as well. Happy to go either way if you have a preference on line 2.

I agree with you here generally. I also find it difficult to decide on 2), but agree with your assessment. We don't know what the user assigns. They might assign counts, so this operation could leave one with a histogram that has perfectly valid variances. Since we don't want to offer a public method to "restore" the variances once they are gone, we should err on the side of rather providing while they are invalid rather than removing them when they are actually valid.
 
> PS: In place operations like scaling also can change the values, didn't include that above, currently they do not invalidate the `.variances()` either.

Ideally, (in-place) scaling should switch the storage from Double() to Weight() before doing the scaling operation, then the variance afterwards would be correct. If that's not possible .variances() should be invalided by that operation.

docs/usage/histogram.rst
```diff
@@ -16,10 +16,15 @@ All storages support a ``weight=`` parameter, and some storages support a ``samp
 
 The summing accumulators (not ``Mean()`` and ``WeightedMean())``) support threaded filling. Pass ``threads=N`` to the fill parameter to fill with ``N`` threads (and using 0 will select the number of virtual cores on your system). This is helpful only if you have a large number of entries compared to your number of bins, as all non-atomic storages will make copies for each thread, and then will recombine after the fill is complete.
 
+Data
+^^^^
+
+The primary value from a histogram is always available as ``.value()``. The variance is available as ``.variances()``, unless you fill an unweighed histogram with weights, which will cause this to be return None, since the variance is no longer computable (use a weighted storage instead if you need the variances). The effective counts are available as ``.counts()``; this
+
```
HDembinski, line 23: Incomplete sentence?

Please just write the "The counts are available as ...". And then "If the histogram is weighted, ``.counts()`` returns  the effective counts see docstring for details.

docs/usage/plotting.rst
```diff
@@ -0,0 +1,40 @@
...
+objects.
+
+Using the protocol:
+
+Plotters should only depend on the methods and attributes listed below. In short, they are:
+
+* ``h.kind``: The bh.Kind of the histogram (COUNT or MEAN)
+* ``h.values()``: The value (as given by the kind)
+* ``h.variances()``: The variance in the value (None if an unweighed histogram was filled with weights)
+* ``h.counts()``: The effective counts
```
HDembinski, line 19: Please change to "How many values the bin received or their effective number if the histogram is weighted".

HDembinski: I am open to the exact wording, but you cannot explain .counts() with the word counts.

HDembinski: Also, there is no such thing as an effective count for an unweighted histogram.

docs/usage/plotting.rst
```diff
@@ -0,0 +1,40 @@
...
+accepts an object that follows the ``PlottableProtocol`` can plot boost-histogram
+objects.
+
+Using the protocol:
+
+Plotters should only depend on the methods and attributes listed below. In short, they are:
+
+* ``h.kind``: The bh.Kind of the histogram (COUNT or MEAN)
+* ``h.values()``: The value (as given by the kind)
+* ``h.variances()``: The variance in the value (None if an unweighed histogram was filled with weights)
```
HDembinski, line 18: This wording is not correct, because .variances() does not give you the variance of the mean if kind == MEAN it gives the variances of the samples. You need to divide that by .counts() to get the variance of the mean.

docs/usage/plotting.rst
```diff
@@ -0,0 +1,40 @@
...
+* ``h.counts()``: The effective counts
+* ``h.axes``: A Sequence of axes
+
+Axes have:
+
+* ``ax[i]``: A sequence of lower, upper bin, or the discrete bin value (integer or sting)
+* ``len(ax)``: The number of bins
+* ``ax.traits.circular``: True if circular
+* ``ax.traits.discrete``: True if discrete (Integer or Category axes)
```
HDembinski, line 27: "True if the bin represents a single value (e.g. Integer or Category axes) instead of an interval (e.g. Regular axes)."

docs/usage/quickstart.rst
```diff
@@ -84,11 +84,17 @@ See :ref:`usage-indexing`.
...
-You can use ``hist.view()`` to get
-a Numpy array (or a RecArray-like wrapper for non-simple storages).
-Most methods like ``.view()`` offer an optional keyword
-argument that you can pass, ``flow=True``, to enable the under and
-overflow bins (disabled by default).
+You can use ``hist.values()`` to get a Numpy array from any histogram. You can
+get the variances with ``hist.variances()``, though if you fill an unweighted
+storage with weights, this will return None, as you no longer can compute the
+variances correctly (please use a weighted storage if you need to). You can
+also get the effective counts with ``.counts()``; this will return counts even
```
HDembinski, line 91: Again please replace "effective counts" with just "counts" and then add an explanation what effective counts are in case of a weighted histogram or better just refer to one detailed explanation somewhere in the docs. Nobody understands what effective counts are unless they are familiar with the concept, and many people are not familiar with it.

docs/usage/quickstart.rst
```diff
@@ -107,6 +113,9 @@ you can set either values or arrays at a time:
     hist[bh.underflow] = 0 # set the underflow bin
     hist2d[3:5, 2:4] = np.eye(2) # set with array
 
+For non-simple storages, you can add an extra dimension that matches the
+constructor arguments of that accumulator.
```
HDembinski, line 117: Can you add an example? I am not sure what you mean.

include/bh_python/register_axis.hpp
```diff
@@ -114,6 +114,41 @@ py::class_<A> register_axis(py::module& m, Args&&... args) {
                  }
              })
 
+        .def_property_readonly("traits_underflow",
+                               [](const A& self) {
+                                   return static_cast<bool>(
+                                       bh::axis::traits::options(self)
```
HDembinski, line 120: `bh::axis::traits::options(self)` just calls `.options()` on a dynamic axes. You don't need to go through the extra layer of abstraction here, I think. Calling `self.options()` directly is more efficient during compilation.

HDembinski: Some additional explanation: The `axis::traits` exist as an indirection layer so that Boost.Histogram can accept any class that supports the minimal required Axis interface. But here you know for sure that `self` is a `boost::histogram::axis::variant`, so you don't need the extra layer of indirection.

setup.cfg
```diff
@@ -44,6 +44,7 @@ package_dir =
 install_requires =
     numpy >=1.13.3
     typing >= 3.5; python_version < '3.5'
+    enum34 >= 1.1; python_version < '3.4'
```
HDembinski, line 47: Are we still supporting 3.4?

src/boost_histogram/_internal/axis.py
```diff
@@ -155,30 +157,47 @@ def _repr_kwargs(self):
...
-            if not self.options.overflow:
+            if not self.traits.overflow:
                 ret += ", overflow=False"
 
         return ret
 
     @property
     def options(self):
         """
+        DEPRECATED: use .trails instead.
```
HDembinski, line 175:
```suggestion
        DEPRECATED: use .traits instead.
```

src/boost_histogram/_internal/axis.py
```diff
@@ -155,30 +157,47 @@ def _repr_kwargs(self):
...
         Return the options.  Fields:
           .underflow - True if axes captures values that are too small
           .overflow  - True if axes captures values that are too large
                        (or non-valid for category axes)
           .growth    - True if axis can grow
           .circular  - True if axis wraps around
         """
+        warnings.warn("DEPRECATED: use .traits instead", FutureWarning)
```
HDembinski, line 183: You could steal and use the deprecated decorator from iminuit.

src/boost_histogram/_internal/hist.py
```diff
@@ -779,10 +819,123 @@ def __setitem__(self, index, value):
...
+            return view.value
+
+    def variances(self, flow=False):
+        # type: (bool) -> Optional[ArrayLike]
+        """
+        Returns the estimated variance of the accumulated values. The sum of squared
+        weights for weighted histograms, the variance of samples for profiles, etc.
+        For an unweighed histogram where kind == "COUNT", this should return the same
+        as values if the histogram was not filled with weights, and None otherwise.
+        If counts is equal to 1 or less, the variance in that cell is undefined if
```
HDembinski, line 878:
```suggestion
        If counts is less than 2, the variance in that cell is undefined if
```

HDembinski, line 878: Perhaps add a note that this condition must be used exactly as described here. counts <= 1 is not the same as counts < 2, when counts returns effective counts.

src/boost_histogram/_internal/hist.py
```diff
@@ -779,10 +819,123 @@ def __setitem__(self, index, value):
...
+        histogram or profile and an effective number of entries (defined below)
+        for a weighted histogram or profile. An exotic generalized histogram could
+        have no sensible .counts, so this is Optional and should be checked by
+        Consumers.
+
+        For a weighted histogram, counts is defined as sum_of_weights ** 2 /
+        sum_of_weights_squared. It is equal or less than the number of times
+        the bin was filled, the equality holds when all filled weights are equal.
+        The larger the spread in weights, the smaller it is, but it is always 0
+        if filled 0 times, and 1 if filled once, and more than 1 otherwise.
```
HDembinski, line 916: Add "It can be used to compute the error on the mean like the ordinary counts, if kind == MEAN".

src/boost_histogram/_internal/hist.py
```diff
@@ -779,10 +819,123 @@ def __setitem__(self, index, value):
...
+
+        If kind == "MEAN", counts (effective or not) can and should be used to
```
HDembinski, line 918: I would switch the order of the last two paragraphs.

src/boost_histogram/_internal/traits.py
```diff
@@ -0,0 +1,47 @@
...
+        underflow=False,
+        overflow=False,
+        circular=False,
+        growth=False,
+        continuous=False,
+        inclusive=False,
+        ordered=False,
+    ):
+        for name in _traits:
+            object.__setattr__(self, name, locals()[name])
```
HDembinski, line 32: Why not `setattr(self, name, locals()[name]`?

tests/plottable.py
```diff
@@ -0,0 +1,154 @@
...
+        kind == "MEAN".
+
+        If kind == "MEAN", the counts can be used to compute the error on the mean
+        as sqrt(variances / counts), this works whether or not the entries are
+        weighted if the weight variance was tracked by the implementation.
+        """
+
+    def counts(self) -> Optional[ArrayLike]:
+        """
+        Count returns the number of entries in each bin for an unweighted
```
HDembinski, line 129: Don't repeat the docstring here. They will get out of sync.

tests/test_axis.py
```diff
@@ -820,9 +849,13 @@ def test_init(self):
         with pytest.raises(TypeError):
             bh.axis.Boolean(1)
 
+    def test_traits(self):
         ax = bh.axis.Boolean()
         assert isinstance(ax, bh.axis.Boolean)
-        assert ax.options == bh.axis.options()
+
+        # Note: this axis not inclusive, since we don't have a dedicated bool
```
HDembinski, line 856: This is bad, the Boolean axis should behave like the Boost.Histogram one, it should be inclusive. One of the reasons to use an explicit Boolean axis is to safe memory when there are really only two alternatives. It does not matter that there is no bool fill. A fill with integers will be cast to boolean values, which is as it should be when you explicitly make a Boolean axis.

HDembinski (commented):
I looked through all of it. I made some comments here and there, if those can be resolved it is good to go. I am concerned about the Boolean axis, but that is another topic to discuss.

tests/test_axis.py
HDembinski: I think we decided to make the Boolean axis just a wrapper over the Integer axis in Python, but it should still behave the same way as in Boost.Histogram.

include/bh_python/register_axis.hpp
```diff
@@ -114,6 +114,41 @@ py::class_<A> register_axis(py::module& m, Args&&... args) {
...
+                                       & bh::axis::option::growth);
+                               })
+        .def_property_readonly(
+            "traits_continuous",
+            [](const A& self) { return bh::axis::traits::continuous(self); })
+        .def_property_readonly(
+            "traits_inclusive",
+            [](const A& self) { return bh::axis::traits::inclusive(self); })
```
HDembinski, line 146: I am not sure whether you need to mirror this in Python. This trait is used internally in Boost.Histogram to make the code a little faster when it is possible.

src/boost_histogram/_internal/axis.py
```diff
@@ -155,30 +157,47 @@ def _repr_kwargs(self):
...
+        """
+        Get traits for the axis - read only properties of a specific axis.
+        """
+        return Traits(
+            self._ax.traits_underflow,
+            self._ax.traits_overflow,
+            self._ax.traits_circular,
+            self._ax.traits_growth,
+            self._ax.traits_continuous,
+            self._ax.traits_inclusive,
```
HDembinski, line 197: Perhaps leave that out for now? What use would it have?

henryiii:
Looks great, thanks for the detailed review, will work on all the points soon. I think I agree with everything except for a couple of minor details that I'll comment on.

> Ideally, (in-place) scaling should switch the storage from Double() to Weight() before doing the scaling operation, then the variance afterwards would be correct.

I don't think, if you ask for a simple storage, you should ever [^1] implicitly change to a different one. If a user wanted variance information, they should chose `Weight()` to begin with, or we could have a `.as(bh.storage.Weight)` explicit conversion. One of the reasons for this is that the memory is invalidated if you change storages, so old `.view()`'s stop working. But mostly, if I ask for a storage of `Double()`, I want storage to be a single Double each unless I ask otherwise, even if the memory was not invalidated.

[^1]: The exception to this is `Unlimited()`, and there, maybe also automatically converting to `Weight()` if filled with a weight or scaled. Or we could add `Auto()`, which would do this. But in both cases, they should be consistent with `.fill`, and a user should be able to opt into a "changing" backend, rather than trying to fill a simple array of value and having it suddenly doubling in size without an explicit request.

>  If that's not possible .variances() should be invalided by that operation.

I think this is best. So we want setting to not invalidate (as a user could and probably often is setting a "normally" filled histogram), but scaling / addition and such to invalidate? I think that sounds fine to me.

docs/usage/plotting.rst
henryiii: I thought we decided in the PlottableProtocol that `.variances()` is always the variance of `.value()`, so that a "dumb" plotter could just plot values and variances and ignore the `.kind` and get something reasonable. It is also simpler if ".values" is the value of the Kind, and the ".variances" is the variances of the Kind. So in this case, the wording is correct, but the value is incorrect, it should be calculated as `.view()["variance"] / .view()["count']`? And a user wishing to get the sample variance would multiply by the `.counts()`?

I could easily be wrong here, though, either/both about what we decided and what is best. @jpivarski what does Uproot return for `.variances()` here for a TProfile?

docs/usage/histogram.rst
henryiii: Sounds good, but we should link or provide the explanation in the docs somewhere (possibly linking to the docstring in the API docs, if double, would be best); docs shouldn't ask users to look at the docstring _in_ Python; a user is not always in a REPL.

docs/usage/quickstart.rst
henryiii: We need to make sure we cross link a bit more, and yes. :)

include/bh_python/register_axis.hpp
henryiii: Fantastic, thanks!

src/boost_histogram/_internal/axis.py
henryiii: I'll take a look, thanks!

src/boost_histogram/_internal/traits.py
henryiii: I think this was... I have no idea why I wrote it this way. Maybe a leftover from when I had some workarounds for Axis (which are no longer needed due to the `__dict__` in `__slots__` simplification), which for some reason I then put here? 🤦 Will clean up, thanks!

tests/plottable.py
henryiii: This is just an vendored copy of the file from `henryiii/uhi` so that we don't have to depend on something in my namespace that isn't published to PyPI yet. Once I clean up UHI, it will be a repo that helps with testing UHI implementations, and will include these protocols for testing PlottibleProtocol Produces and Consumers, we can drop this file and just use this (testing requirement only).

That's why it's a duplicate, though - it's only put here for testing purposes.

tests/test_axis.py
henryiii: It does behave the same way - it is "inclusive" of bools - but since it's an integer axis to avoid adding 50% extra compile complexity on things like `.fill`, it's not "truly" inclusive (as in, not all integers). We need this to not be a true Bool axis from Boost.Histogram because all axis are sliceable - if someone wants to do `h[0:1]`, that's valid in boost-histogram, it's a false-only axis. You helped me come up with this axis type, IIRC.

Now, we could override `.inclusive` on our custom Bool axis, I believe - not knowing what optimizations are done here, I didn't change that - but maybe that's actually better. Inclusive would be "true" if length 2, and "false" if this was length 1.

src/boost_histogram/_internal/axis.py
henryiii: Yes, probably best. I was just adding things I saw in the interface to Boost.Histogram that looked remotely useful, since `continuous` was useful and missing.

henryiii:
I am "mostly" off during this week, by the way. Most of the time I'm in front of a computer will be aimed at several older projects I maintain personally and am not paid to work on, so work on the Scikit-HEP and related projects will be very slow this week (except if I can help push along the Universal2 support in cibuidwheel, that's something I would like to have available soon as possible, ideally before the next boost-histogram release).

docs/usage/quickstart.rst
henryiii: Added one.

docs/usage/quickstart.rst
henryiii: Holding this open till I set up some cross linking.

src/boost_histogram/_internal/axis.py
henryiii: Fixed an issue present before the decorator with broken tab completion, and then added it. Made a few changes to the decorator, will quickly outline them in a comment box.

src/boost_histogram/_internal/deprecated.py
```diff
@@ -0,0 +1,30 @@
...
+                "{0} is deprecated: {1}".format(
+                    self._name or func.__name__, self._reason
+                ),
+                category=FutureWarning,
+                stacklevel=2,
+            )
+            return func(*args, **kwargs)
+
+        decorated_func.__doc__ = "DEPRECATED: " + self._reason + "\n" + func.__doc__
+        return decorated_func
```
henryiii, line 30: Quick outline of changes from iMinuit:
* I needed a settable name so that I could implement a workaround for properties and tab completion splitting out the deprecation warning even when unrelated.
* I'm using [`functools.wraps`](https://www.geeksforgeeks.org/python-functools-wraps-function/) because it gets lots of properties right, including the signature, at least for `help()` and `inspect.signature`. It's not as good as the [decorator](https://pypi.org/project/decorator/) or [wrapt](https://pypi.org/project/wrapt/) libraries, but I think it's okay for this.
* I'm using FutureWarning, as that's targeted at changes affecting users. DeprecationWarnings is for library authors using your library (who happen to remember to turn them on, they tend to be hidden). [Nice explanation here](https://deprecated.readthedocs.io/en/latest/tutorial.html#deprecation-warning-classes).
* I'm including the original doc string at the end (the user can delete it if they want to).

By the way, [this series of 10 articles](https://github.com/GrahamDumpleton/wrapt/tree/master/blog#readme) on wrapping functions by the wrapt author is fantastic, at least it is if I remember correctly, I think I last read it around the time he wrote it. Not needed for a simple internal decorator like this, but it's why I'd use a proper library for anything public.

henryiii:
@HDembinski I've fixed everything, I hope to merge soon, we can iterate before release. I went with matching the description we agreed on in UHI: variance is the variance of the value, regardless of the Kind. I don't know if that's best, that's just how I understood our agreement, and how I would expect this to be most useful for plotting; "dumb" plotters should get as close as possible by just plotting value and variance; that's what we worked on for the return value from variance when the storage is unweighed, as well.

Happy to hear arguments in either direction, though.

Maybe implementing this in mplhep will help clear up this confusion?

I'm also preparing "UHI" so we don't have to duplicate documentation and the protocol.

HDembinski:
Sorry for the late reply, I am drowning in work right now. I am not sure whether we understand the same under "always return the variance". The rationale in Boost.Histogram was to always provide the variance that requires least computation. In case of the mean accumulator, the variance of the samples is computed. Analog for the weighted_mean accumulator. Since variance() here is a secondary interface, you don't have to follow that lead. It is ok to return the variance of the mean instead of the variance of the samples. I am not sure what the average user wants. For statistical analysis you want the variance of the mean, not the variance of the samples. Returning the variance of the mean is more consistent wrt other storages.

I don't feel strongly either way, I am leaning slightly toward returning variance of the mean, but as long as it is correctly documented what .variance() returns both options are ok.

henryiii:
Invalidating the variance on an inplace operation was a one line fix. :)

henryiii:
Let's get this in, then iterate a bit as needed. It is the variance of the mean at the moment. I'd like to compare with ROOT in the tests, but converting is not completely trivial. This PR is big enough. :)

PR #477, opened 2020-11-26
feat: __dict__ major simplification

henryiii:
This removes the whole facade and just really makes this a `__dict__` class. You can now manipulate __dict__ directly, it's completely natural, and C++ just caries a pointer to it. If you need to clear the dict, set it, etc., you can.

### DIFF
--- a/src/boost_histogram/_internal/axis.py
+++ b/src/boost_histogram/_internal/axis.py
@@ -304,6 +304,9 @@ class Regular(Axis):
             The full metadata dictionary
         """
 
+        # Inheriting an axis and forgetting to add __slots__ should be an error
+        assert not hasattr(self, "__weakref__"), "Axis subclasses must have __slots__!"
+
         with KWArgs(kwargs) as k:
             metadata = k.optional("metadata")
             transform = k.optional("transform")
@@ -411,6 +414,10 @@ class Variable(Axis):
         __dict__: Optional[Dict[str, Any]] = None
             The full metadata dictionary
         """
+
+        # Inheriting an axis and forgetting to add __slots__ should be an error
+        assert not hasattr(self, "__weakref__"), "Axis subclasses must have __slots__!"
+
         with KWArgs(kwargs) as k:
             metadata = k.optional("metadata")
             __dict__ = k.optional("__dict__")
@@ -492,6 +499,10 @@ class Integer(Axis):
         __dict__: Optional[Dict[str, Any]] = None
             The full metadata dictionary
         """
+
+        # Inheriting an axis and forgetting to add __slots__ should be an error
+        assert not hasattr(self, "__weakref__"), "Axis subclasses must have __slots__!"
+
         with KWArgs(kwargs) as k:
             metadata = k.optional("metadata")
             __dict__ = k.optional("__dict__")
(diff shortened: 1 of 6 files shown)


henryiii:
@HDembinski I need to implement this on all axes types, but before I do, does this look reasonable? It makes the "this is a normal `__dict__` class" facade more believable and makes it much easier to do things like remove all metadata by clearing `__dict__`.

HDembinski:
Looks ok to me.

henryiii:
Major simplification in last commit (please look at it by itself!) I came across a mention in the [Python docs](https://docs.python.org/3/reference/datamodel.html#notes-on-using-slots) that you can have a mixed `__slots__` and `__dict__` class! The `__slots__` part is not in the `__dict__`, so no issues with recursiveness!

This fixed a bug with `__slots__` missing on Category axes, causing them to misbehave when assigning metadata - but now it's all moot anyway because you can forget `__slots__` and it still works!

src/boost_histogram/_internal/axis.py
```diff
@@ -30,35 +30,55 @@ def _isstr(value):
...
-        other._ax = copy.copy(self._ax)
-        return other
+    def __setattr__(self, attr, value):
+        if attr == "__dict__":
+            self._ax.metadata = value
+        object.__setattr__(self, attr, value)
 
-    def __getattr__(self, item):
-        if item == "_ax":
-            return Axis.__dict__[item].__get__(self)
```
henryiii, line 42: This sort of thing is no longer needed! :)

src/boost_histogram/_internal/axis.py
```diff
@@ -553,7 +592,9 @@ def _repr_args(self):
 @set_module("boost_histogram.axis")
 @register({ca.category_int, ca.category_int_growth})
 class IntCategory(BaseCategory):
-    @inject_signature("self, categories, *, metadata=None, growth=False")
+    __slots__ = ()
```
henryiii, line 595: This was a bug. But it wouldn’t be anymore!

src/boost_histogram/_internal/axis.py
```diff
@@ -30,35 +30,55 @@ def _isstr(value):
...
+        return {"_ax": self._ax}
 
-    def __dir__(self):
-        metadata = list(self._ax.metadata)
-        return sorted(dir(type(self)) + metadata)
+    def __copy__(self):
+        other = self.__class__.__new__(self.__class__)
+        other._ax = copy.copy(self._ax)
+        other.__dict__ = other._ax.metadata
+        assert other.__dict__ is other._ax.metadata
```
henryiii, line 80: Note to self: I can probably remove these now

src/boost_histogram/_internal/axis.py
```diff
@@ -30,35 +30,55 @@ def _isstr(value):
...
+
+        if __dict__ is not None and metadata is not None:
+            raise KeyError(
+                "Cannot provide metadata by keyword and __dict__, use __dict__ only"
             )
-            raise AttributeError(msg)
+        elif __dict__ is not None:
+            self.__dict__ = __dict__
+        elif metadata is not None:
+            self.__dict__["metadata"] = metadata
```
henryiii, line 63: Note to self: check this, might be made slightly simpler.

henryiii:
@HDembinski You've already agreed to the basic idea, so I'll merge at some point if I need to to work on the next PR without diverging too much - but feel free to let me know if I need to change / pull / add anything even after merging if needed!

henryiii:
I'm going to try to work on #476 today, and axis changes (if there are any) might interact with this. So merging!

PR #479, opened 2020-12-07
fix: slicing projection with one-sided slices

### LINKED ISSUE #478: Bug in h[start::bh.sum] from v0.8 onwards
A bug was discovered in the cut and sum syntax which is present from v0.8 onwards.

Demo:
```py
import boost_histogram as bh
import numpy as np

h1 = bh.Histogram(bh.axis.Integer(0, 5))
h1.view(True)[:] = 1

start = 0
sum_bh = h1[start :: bh.sum]
sum_np = np.sum(h1.view(True)[start + 1 :])

print("np", sum_np, "bh", sum_bh)
```
In v0.7, this gives the correct answer `np 6.0 bh 6.0`, the overflow bin is added to the sum. In v0.8+ the overflow bin is not counted `np 6.0 bh 5.0`. Since no upper limit is given, the expected behavior is that the overflow bin is counted.

This bug affects analysis results, so very severe.

I am not sure whether this was introduced in C++ code or Python wrapper code. Could be related to #373

HDembinski:
Fixes #478 

* updated boostorg/histogram to latest develop with bug-fix
* added test
* Updated the rest of boost to 1.75

The core bug is fixed in boostorg/histogram, but there still seems to be a bug when `bh.loc` is used and the value is past the upper end of the axis. The computed index then should be size + 1 to make the crop work correctly.

### DIFF
--- a/tests/test_histogram_indexing.py
+++ b/tests/test_histogram_indexing.py
@@ -205,6 +205,29 @@ def test_mix_value_with_slice_2():
     assert_array_equal(h2.shape, (5, 2, 2))
 
 
+def test_one_sided_slice():
+    h = bh.Histogram(bh.axis.Regular(4, 1, 5))
+    h.view(True)[:] = 1
+
+    assert h[sum] == 6  # 4 (internal bins) + 2 (flow bins)
+    assert h[-1:5:sum] == 6  # keeps underflow, keeps overflow
+
+    # check that slicing without bh.sum adds removed counts to flow bins
+    assert_array_equal(h[1:3].view(True), [2, 1, 1, 2])
+
+    assert h[0::sum] == 5  # removes underflow, keeps overflow
+    assert h[:4:sum] == 5  # removes overflow, keeps underflow
+    assert h[0:4:sum] == 4  # removes underflow and overflow
+
+    assert h[bh.loc(1) :: sum] == 5  # remove underflow
+    assert h[: bh.loc(5) : sum] == 5  # remove overflow
+    assert h[bh.loc(1) : bh.loc(5) : sum] == 4  # removes underflow and overflow
+
+    assert h[bh.loc(0) :: sum] == 6  # keep underflow
+    assert h[: bh.loc(10) : sum] == 6  # keep overflow
+    assert h[bh.loc(0) : bh.loc(10) : sum] == 6
+
+
 def test_repr():
     assert repr(bh.loc(2)) == "loc(2)"
     assert repr(bh.loc(3) + 1) == "loc(3) + 1"
--- a/tests/test_histogram_indexing.py
+++ b/tests/test_histogram_indexing.py
@@ -210,7 +210,7 @@ def test_one_sided_slice():
     h.view(True)[:] = 1
 
     assert h[sum] == 6  # 4 (internal bins) + 2 (flow bins)
-    assert h[-1:5:sum] == 6  # keeps underflow, keeps overflow
+    assert h[bh.tag.at(-1) : bh.tag.at(5) : sum] == 6  # keeps underflow, keeps overflow
 
     # check that slicing without bh.sum adds removed counts to flow bins
     assert_array_equal(h[1:3].view(True), [2, 1, 1, 2])
@@ -224,8 +224,8 @@ def test_one_sided_slice():
     assert h[bh.loc(1) : bh.loc(5) : sum] == 4  # removes underflow and overflow
 
     assert h[bh.loc(0) :: sum] == 6  # keep underflow
-    assert h[: bh.loc(10) : sum] == 6  # keep overflow
-    assert h[bh.loc(0) : bh.loc(10) : sum] == 6
+    assert h[: bh.loc(10) + 1 : sum] == 6  # keep overflow
+    assert h[bh.loc(0) : bh.loc(10) + 1 : sum] == 6
 
 
 def test_repr():
--- a/src/boost_histogram/tag.py
+++ b/src/boost_histogram/tag.py
@@ -107,9 +107,6 @@ class at(object):
         self.value = value
 
     def __call__(self, axis):
-        if self.value < -2:
-            raise IndexError("Index cannot be less than -1")
-
         return self.value
 
 


henryiii:
Just for clarification, it should be capped at size + 1? Should it be capped at -1 too? And should the capping happen in the Python layer, or in the C++ wrapper? In the C++ wrapper would keep the various UHI locators much simpler.

HDembinski:
Indices do not have to be capped, they are capped automatically in `reduce`.

HDembinski:
Ideally, the value to index conversion should happen in C++, so that it is guaranteed to be consistent. I think the current implementation always converts the value to an index before passing it to `reduce`.

henryiii:
Yes, it does - so the problem is that `.index(past_end)` returns `size` instead of `size + 1`?

henryiii:
Actually, no, this is ideal, I think. `loc(past_end)` should not include the overflow bin. The overflow bin is infinite. Only leaving this off (None) should include the overflow bin.

HDembinski:
Remembering more about the details: The problem is not what `.index` returns. The rules for what happens when doing slice over a value range is not perfectly compatible with what `.index` returns. There is special code in `reduce` to handle this.

HDembinski:
> Actually, no, this is ideal, I think. `loc(past_end)` should not include the overflow bin. The overflow bin is infinite. Only leaving this off (None) should include the overflow bin.
> 
> `h[0::sum]` includes the overflow bin, `h[0:anything_finite:sum]` does not, that's in the UHI spec.

Then it is a) inconsistent with how Boost Histogram works and b) it is inconsistent, because the current implementation handles the underflow cases differently to the overflow case.

HDembinski:
We said at some point that a value range specified should include all bins which *overlap* with that value range. That should include the flow bins.

henryiii:
It's not as clear as I thought, since it only refers to standard indexing:

https://boost-histogram.readthedocs.io/en/latest/usage/indexing.html#slicing

But triggering an infinitely wide bin by accident seems quite dangerous, especially when you can't do this accidentally by integer indexing (-1 is invalid), but you can via `loc`? Floating point operations could include a bin you were not intending, and an infinite bin you might not even know about would be pretty bad. I think it would be much better to be explicit here if you want to include a flow bin.

Either way, the current inconsistent behavior is wrong. :)

HDembinski:
The rule is: if the bin *overlaps* with the value range specified, it should be included. That also goes for bins which are infinitely large and I don't think that this is anything to worry about. You can make sure that the overflow bin is excluded by picking the upper value inside the last bin of the axis up to the upper edge of that bin.

henryiii:
Think about it carefully, making the best decision for Python users, then let me know. I worry if someone selects an upper edge of `bh.loc(6.0000000001)` on a 0-6 axis and suddenly get an overflow bin. I could see an easy mistake being "I know my histogram is has an upper limit of about 10, so I'll just use `bh.loc(20)` as an upper limit to be safe". Or maybe the limit is a variable. I can't easily imagine a case where you would have a variable upper or lower edge but would prefer to have it suddenly include infinity. Most cases are "I want a range" OR "I want all above/below x".  Overflow bins _in integrations_ only being included if explicitly requested nicely fits under "explicit is better than implicit", I would think.

henryiii:
Everywhere else in boost-histogram the flow bins are either hidden (carried along), natural (sum over None endpoints includes everything, like `::sum`), or explicit (`bh.underflow/bh.overflow`, `flow=true`).

HDembinski:
> Think about it carefully, making the best decision for Python users, then let me know. I worry if someone selects an upper edge of `bh.loc(6.0000000001)` on a 0-6 axis and suddenly get an overflow bin in a sum. I could see an easy mistake being "I know my histogram is has an upper limit of about 10, so I'll just use `bh.loc(20)` as an upper limit to be safe". Or maybe the limit is a variable. I can't easily imagine a case where you would have a variable upper or lower edge but would prefer to have it suddenly include infinity. Most cases are "I want a range" OR "I want all above/below x". Overflow bins _in integrations_ only being included if explicitly requested nicely fits under "explicit is better than implicit", I would think.

It has to be documented properly, of course, but it is important to have simple rules that people can remember. The flow bins are special in that you normally don't see them, but they should be handled just like other bins as much as possible (from a logical point of view and also from a programming point of view).

Note that what you worry about is only a problem when there are actually entries in the flow bins. It is hard to say whether the user wants to keep those or not when he/she uses an upper value beyond the axis range. Maybe keeping the overflow is exactly the intention?

HDembinski:
I need to move fast with the patch in boostorg/histogram to get this still included in the upcoming release in two days.

henryiii:
This is perfectly fine with Boost.Histogram, I'm just looking at the UHI slicing syntax. Let's pick the include syntax for now, but revisit it to see if it's more consistent/user friendly to make overflow/underflow _always_ explicit when reducing via UHI reducer before a boost-histogram release.

henryiii:
> So you would prefer including the flow bins when values are passed only if the value passed is infinite, right?

Yes, so these would be the only ways to apply a reducer to a flow bin:

```python
h[start:None:sum] # More usually written as h[start::sum]
h[start:bh.loc(infty):sum]
```

At least that is what I think the original intention was, and I think it would be more explicit?

HDembinski:
I like the simple consistent rule, but I do see your point and I worry about one thing: I don't remember whether the current implementation of regular axis is consistent so that `axis.index(axis.value(i))` always returns `i`. If we pick my approach and `axis.value(len(axis))` happens to be too large by a tiny amount, the overflow bin would be included even though the intention was probably to exclude it.

henryiii:
For rules, `h[start:stop]` isn't affected, since it always bring along the underflow/overflow bins, since it returns a new histogram. So we are left with `h[select]`, but that's a different construct anyway; if you write `h[bh.loc(too_big)]`, the choices are error or return the overflow bin; since you asked for a bin, it is quite natural to return a bin, even the overflow bin. You also are not asking for a range - including infinity when you request finite range is much more worrisome because infinity makes the width of the range infinite. So I think this is not inconsistent?

Also, this, and especially generalized reducers, `h[a:b:ufunc]` when implemented, don't have an exact parallel to Boost.Histogram, so while we should consider how slices work in Boost.Histogram, but the behavior doesn't have to be a perfect match. And Boost.Histogram explicitly assumes users are knowledgeable enough to handle knowing about flow.

HDembinski:
> For rules, `h[start:stop]` isn't affected, since it always brings along the underflow/overflow bins, since it returns a new histogram. So we are left with `h[select]`, but that's a different construct anyway; if you write `h[bh.loc(too_big)]`, the choices are error or return the overflow bin; since you asked for a bin, it is quite natural to return a bin, even the overflow bin. You also are not asking for a range - including infinity when you request finite range is much more worrisome because infinity makes the width of the range infinite. You should already be checking the properties of `h.axes[0][bh.loc(too_big)]` to see what you are looking at. So I think this is not inconsistent?
> 
> Also, this, and especially generalized reducers, `h[a:b:ufunc]` when implemented, don't have an exact parallel to Boost.Histogram, so while we should consider how slices work in Boost.Histogram, but the behavior doesn't have to be a perfect match. And Boost.Histogram explicitly assumes users are knowledgeable enough to handle knowing about flow.

I cannot say much about this, because it seems to address other points. Can we stick to the topic how `h[bh.loc(a):bh.loc(b):sum]` and `h[bh.loc(a):bh.loc(b)]` should work? This is independent to what happens with `h[select]`. The logically consistent thing is that `h[bh.loc(too_big)]` gives you the overflow bin, I thought that was obvious.

henryiii:
> My mind is set on applying rules consistently that are as simple as possible, so behavior remains easy to remember.

I was just comparing to the existing rules to see if there was an inconstancy if we apply this explicit endpoint rule for reducers. I don't think so.

Also this does include `h[bh.loc(a):bh.loc(b)]` - this returns a new histogram, so the flow bins are always implicitly included and summed if needed no matter what you do here. `h[bh.loc(a):bh.loc(b)][::sum]` will always sum over the entirety of h regardless of what you select in the first step.

HDembinski:
I cannot say much about this, because it seems to address other points. Can we stick to the topic how `h[bh.loc(a):bh.loc(b):sum]` and `h[bh.loc(a):bh.loc(b)]` should work?

Naturally, `h[bh.loc(a):bh.loc(b)]` and `h[bh.loc(a):bh.loc(b):sum]` should select the same bin ranges. The only difference between the two is what happens with the overflow bins. When you do `h[bh.loc(a):bh.loc(b)]`, the content of the removed bins are added to the respective flow bins, while in case of `h[bh.loc(a):bh.loc(b):sum]` we decided to crop the content of the removed bins. But both commands should select the same bin range for any values of a and b.

With `h[bh.loc(a):bh.loc(b)][::sum]` you cannot do a one-sided crop, I don't see how that relates to the discussion (also no one should do that, it looks awkward).

henryiii:
And they currently do. In fact, there is no difference in `h[bh.loc(a):bh.loc(b)]` if you pick a `b` inside the final bin, or if you pick a `b` past the final bin. You want to introduce a difference in the one with `sum` based on this distinction, however.

This affects two sided crops too, by the way, if `a<min` and `b>max`.

henryiii:
My point was `h[bh.loc(a):bh.loc(b)][::sum]` does not crop at all. The first step makes a smaller histogram, but then the second step sums over the flow bins too, so it is identical to the same action before the cropping. So the only time you truly "crop" is in the reducer expression. (It might be nice to have a way to return new histograms with flow bins turned off for an axis, but not something that currently exists, and not something that would likely ever be part of UHI, which describes conceptual manipulations, not changes in axis type specifics).


Edit on last comment: You want to introduce a difference for the upper boundary. I want to remove the distinction for the lower boundary. The current situation is of course an inconstant mess. I am assuming that whatever we decide, it will be applied to both endpoints consistently, and I just sometimes select one endpoint to discuss for clarity.

HDembinski:
> And they currently do. In fact, there is no difference in `h[bh.loc(a):bh.loc(b)]` if you pick a `b` inside the final bin, or if you pick a `b` past the final bin. You want to introduce a difference in the one with `sum` based on this distinction, however.

Not sure if you are referring to my original statement that slices select bin ranges that overlap with (a,b) or something else. I am currently listening to some talks, but can we discuss this later via Zoom?

HDembinski:
As far as I can see, either way nothing changes for `h[bh.loc(a):bh.loc(b)]`.

henryiii:
Collecting the discussion, I think the only thing we are considering changing is the following behavior.

```python
h = bh.Histogram(bh.axis.Regular(1,1,2))
h.view()[...] = 1

# Explicit endpoints
3 == h[::sum]
2 == h[:bh.loc(2.00001):sum]
2 == h[bh.loc(.99999)::sum]
1 == h[bh.loc(.99999):bh.loc(2.00001)]
1 = h[0:len:sum]

# Flow inclusive endpoints
3 == h[::sum]
3 == h[:bh.loc(2.00001):sum]
3 == h[bh.loc(.99999)::sum]
3 == h[bh.loc(.99999):bh.loc(2.00001)]
1 = h[0:len:sum]
```

You can replace an empty endpoint with None (syntactically equivalent) or possibly `bh.loc(float("infty"))`.

Note that if you were to do this in two steps, doing the slice first before the reduction, there is no difference if including an a point in the edge bin vs. in the flow bin in the "normal" slicing.

```python
1 == h[bh.loc(.99999):bh.loc(2.00001)][0:len:sum]
1 == h[bh.loc(1):bh.loc(2)][0:len:sum]

3 == h[bh.loc(.99999):bh.loc(2.00001)][::sum]
3 == h[bh.loc(1):bh.loc(2)][::sum]
```

henryiii:
Outcome of meeting:

* UHI locators need to be able to detect if they are in a slice to enable advanced tricks. They only receive the axis as an argument, so my thought would be to include a `_slice_` object on the axis (proxy) that they receive - this would even allow them to detect what the action was.
* The default `bh.loc` would avoid including overflow/underflow bins by default. Writing a custom loc function that would include these would be easy.

henryiii:
* We can (finally) fix `bh.overflow` to behave nicely in reductions with this! Currently `h[:bh.overflow:sum]` slices to one-before-the-bin, which is the final bin, not the overflow bin. For example, for a 10 bin histogram, `h[5:bh.overflow:sum]` simply reduces to `h[5:10:sum]`, and Python slicing does not include the endpoint - so no overflow bin. We can fix this if we want to now.

Does that sound useful?

HDembinski:
>     * We can (finally) fix `bh.overflow` to behave nicely in reductions with this! Currently `h[:bh.overflow:sum]` slices to one-before-the-bin, which is the final bin, not the overflow bin. For example, for a 10 bin histogram, `h[5:bh.overflow:sum]` simply reduces to `h[5:10:sum]`, and Python slicing does not include the endpoint - so no overflow bin. We can fix this if we want to now.
> 
> 
> Does that sound useful?

Yes.

HDembinski:
How do we go about with this PR? @henryiii could you finish the work by committing to the branch that I opened? I think you know best how to implement `loc`.

henryiii:
Sure, I can do that.

Fun fact: we need the extra information as described above even if we wanted to include endpoints - the reason that currently `h[:bh.loc(1e9):sum]` and `h[bh.loc(-1e9)::sum]` differ is due to Python not including the endpoint in the slice.

henryiii:
Also, this affects `h[:bh.loc(.5)]`, where this is a point inside the histogram. Currently, `bh.loc(.5)` returns the index of the bin that contains `.5`, and then Python slicing syntax takes over - and goes up to _but not including_ this bin. Before the above planned change, we had no way to do anything different - returning one more than the index containing the bin would break `h[bh.loc(.5)]`. And `h[bh.loc(.5):]`, too, for that matter. Do we want to change this now, so that `bh.loc` when not on a bin edge and when the stop of a slice will return the bin + 1?

I remember describing the first behavior in PyHEP 2019, and then I think we had a discussion and decided on the second behavior for Boost.Histogram, but since this part is controlled by normal indexes (since UHI converts all values to indexes, and then boost-histogram uniformly uses indices behind the scenes), this behavior inherits the natural Python index behavior and never changed.

HDembinski:
Slicing with indices works in C++ in the same way as in Python. In a range a:b, b is not included. In Boost.Histogram that does not matter at all, because I don't use a construct like bh.loc. You pass a slice with the slice object, which then has normal slicing behavior. A slice can only use indices. To use values you need the crop or shrink object, which only accept values. The two concepts are well separated and there is no problem that one includes the bin that contains the value and the other does not.

The design in Pyhton has become awkward because you force both values and indices to go through a Python slice. The behavior of the slice simply does not generalize to this case.

HDembinski:
Note that Jim's rule was born out of what the user most likely needs, we should not forget that. I agree that for consistency, bh.loc should not automatically add +1 when it is on the upper edge, because that is awkward. Instead, we should offer an alternative object that you can use instead of the slice, which keeps the bins that overlap.

henryiii:
This is part of UHI, the locators control the conversion to integers, giving this power to users very powerful. We also can't directly use Boost.Histogram's implementation anyway, because it doesn't support mixing endpoint types; there is no way to do `h[0:bh.loc(.5):sum]`, or `h[bh.loc(.5):len:sum]`.

The question here is do we keep the current simpler behavior or convert to the more like Boost.Histogram's  behavior?

henryiii:
Okay, that's much less disruptive. And I do think it tends to be a common need, it's often mentally simpler to request a range that is inclusive when using data coords instead of bins, hence the immediate request for a +1. What would you like to call it?

henryiii:
`bh.incloc`, for inclusive loc, perhaps? How about `bh.contains`? It's a little long but I like that it is clear, `h[bh.contains(.5):bh.contains(1.5)]` is very obvious to someone unfamiliar with our library.

HDembinski:
My proposal is to have another object for doing a slice in values. I don't care that you cannot mix values and integers then, because I don't believe that's what you want. For this PR, it does not matter how we solve that in future. Let's focus on this PR. For this PR, we do as you suggest, bh.loc gives you the index that corresponds to the value. If you want to include the upper bin you do bh.loc(value) + 1.

HDembinski:
In other words, I was happy with the result of our Zoom call, let's just implement that and worry about other ways of slicing later.

HDembinski:
I wrote up what I had in mind for a value_slice here https://github.com/scikit-hep/boost-histogram/issues/482, so that we can focus on this PR, which is strictly about the behavior of bh.loc.

henryiii:
> I was happy with the result of our Zoom call

I just was afraid that you thought the behavior of loc was different in the Zoom call (I didn't really think of it then, either).

No problem, was not going to block or make too many changes in one PR anyway. I'll get to this late today or tomorrow.

tests/test_histogram_indexing.py
```diff
@@ -205,6 +205,29 @@ def test_mix_value_with_slice_2():
     assert_array_equal(h2.shape, (5, 2, 2))
 
 
+def test_one_sided_slice():
+    h = bh.Histogram(bh.axis.Regular(4, 1, 5))
+    h.view(True)[:] = 1
+
+    assert h[sum] == 6  # 4 (internal bins) + 2 (flow bins)
+    assert h[bh.tag.at(-1) : bh.tag.at(5) : sum] == 6  # keeps underflow, keeps overflow
```
henryiii, line 213: UHI follows normal Python indexing, so since `h[-1:5:sum]` should match `h[-1:5]`, which is a valid NumPy expression, the original form of this line was incorrect. I've corrected it to use `bh.tag.at()`, which is boost-histogram style -1 to N+1 indexing.

tests/test_histogram_indexing.py
```diff
@@ -205,6 +205,29 @@ def test_mix_value_with_slice_2():
...
+    assert h[0::sum] == 5  # removes underflow, keeps overflow
+    assert h[:4:sum] == 5  # removes overflow, keeps underflow
+    assert h[0:4:sum] == 4  # removes underflow and overflow
+
+    assert h[bh.loc(1) :: sum] == 5  # remove underflow
+    assert h[: bh.loc(5) : sum] == 5  # remove overflow
+    assert h[bh.loc(1) : bh.loc(5) : sum] == 4  # removes underflow and overflow
+
+    assert h[bh.loc(0) :: sum] == 6  # keep underflow
+    assert h[: bh.loc(10) + 1 : sum] == 6  # keep overflow
```
henryiii, line 227: This (now) works exactly as expected - it goes to the overflow bin, but Python indexing does not include the upper bound, so you need to add +1 to include the overflow bin in the slice.

henryiii:
This looks good, now, will merge. I'll make a followup PR to add the smarter UHI support, which will allow locators to detect which part of a slice they are in, allowing much more powerful locators to be written by users.

PR #481, opened 2020-12-10
docs: support links and template

henryiii:
On gitter, @HDembinski's was worried that we might be splitting the community by using GitHub's Discussions, so this clarifies the role a bit and replaces the Stack Overflow link (we've only had two questions there; the barrier of opening an account and gaining reputation might be part of the issue, it's also much harder to track for me). My expectation is that any issue that is not resolved by code changes, such as a question on how to do something, should either be opened as a Discussion, or moved to Discussions - answering a question via Q&A keeps the "issue from being closed" and therefore making it harder for new users to find and learn from the answered question (as discussed at the IRIS-HEP Analysis Systems meeting today).

@matthewfeickert, this is what I'm thinking. There is a discussions badge, which helps users see that we are using Discussions, even from the ReadTheDocs page. Then there's a link in the new issue page that suggests using Q&A for questions when opening an issue. Of course, we can manually move issues over to Discussions to keep them from being "closed" when answered, as well.

- docs: restructure readme links
- docs: add issue templates


matthewfeickert:
> There is a discussions badge, which helps users see that we are using Discussions, even from the ReadTheDocs page. Then there's a link in the new issue page that suggests using Q&A for questions when opening an issue. Of course, we can manually move issues over to Discussions to keep them from being "closed" when answered, as well.


This sounds good to me. :+1: Once this gets merged in I'll take a look at how this looks for Boost.Histogram (looking quickly at this PR it seems that Issue templates have evolved quite a bit since 2018 when I put them in for `pyhf`) and then follow suit.

henryiii:
Boost.Histogram is C++, boost-histogram is this package. ;)

matthewfeickert:
> Boost.Histogram is C++, boost-histogram is this package. ;)

Noted. :) I actually appreciate you pointing this out to me, as getting these naming conventions correct can be important to libraries getting the recognition they need!

henryiii:
By the way, you can already see the [issue config in place](https://github.com/scikit-hep/hist/issues/new/choose) in Hist. There are more issue templates there, though we might trim those down in the future now that we have Discussions.

henryiii:
It's a fun naming convention when speaking. That's why I was surprised that the Rev transcriber got it right most of the time for the CHEP talk. :)

HDembinski:
Looks good to me, especially if it replaces StackOverflow which I never check.

For the record: I personally don't think we are at a stage where we need issue templates. It is not that we get flooded with poorly written issues, but I don't want to discuss this endlessly and it is not something worthwhile blocking.

henryiii:
That's fine, I agree actually - I'm not doing the same thing as in Hist, where there are lot of issue templates that Nino set up; that's just an example of what the links look like in the list. I just added a single one here because otherwise the only way to make an issue would be the small "blank issue" button at the bottom, which not everyone would be able to find. I'm perfectly fine to leave it completely blank too, I just want it to show up as a clear, green button in the list.

henryiii:
This is what it looks like inplace:

Screen Shot 2020-12-11 at 9 59 00 AM

HDembinski:
It is nice and increasingly commonplace, thanks Henry.

PR #483, opened 2020-12-18
feat: support _to_boost_

henryiii:
Pulled the change out from #475 so that it can get a proper discussion if it needs it.

Boost-histogram already supports casting - you can call `Hist(a_boost_object)`, where Hist is any boost-powered project, including Hist, and it converts. Likewise with `Histogram(a_hist_object)` . This allows uproot to participate - you can call `bh.Histogram(an_uproot_histogram)` and it converts by first calling `to_boost`, then using the standard mechanism to convert it to `bh.Histogram` (or Hist, or any other dependent project if more get added).

The main question I'd like to discuss is: should we use the current method uproot provides, `.to_boost()`, or instead provide a hidden method to do this, like `_to_boost_()`, and add it to uproot? If Physt were to add a method, I bet Jan would be fine with `to_boost()`, but not requiring a user-visible method might be better? Also, in the future, we might want to provide a way to produce a structure that boost-histogram can adapt to reduce coupling - but that's _very_ far away - sticking to a hidden method now would keep us from having to expose a public method tied to the old method if a new one was added.

Edit: As seen below, I went with `_as_boost_histogram_()`.

### DIFF
--- a/src/boost_histogram/_internal/hist.py
+++ b/src/boost_histogram/_internal/hist.py
@@ -134,8 +134,8 @@ class Histogram(object):
             return
 
         # Support objects that provide a to_boost method, like Uproot
-        elif len(axes) == 1 and hasattr(axes[0], "_as_boost_histogram_"):
-            self.__init__(axes[0]. _as_boost_histogram_())
+        elif len(axes) == 1 and hasattr(axes[0], "_to_boost_"):
+            self.__init__(axes[0]. _to_boost_())
             return
 
         # Keyword only trick (change when Python2 is dropped)
--- a/src/boost_histogram/_internal/hist.py
+++ b/src/boost_histogram/_internal/hist.py
@@ -134,8 +134,8 @@ class Histogram(object):
             return
 
         # Support objects that provide a to_boost method, like Uproot
-        elif len(axes) == 1 and hasattr(axes[0], "to_boost"):
-            self.__init__(axes[0].to_boost())
+        elif len(axes) == 1 and hasattr(axes[0], "_as_boost_histogram_"):
+            self.__init__(axes[0]. _as_boost_histogram_())
             return
 
         # Keyword only trick (change when Python2 is dropped)
--- a/src/boost_histogram/_internal/hist.py
+++ b/src/boost_histogram/_internal/hist.py
@@ -134,8 +134,8 @@ class Histogram(object):
             return
 
         # Support objects that provide a to_boost method, like Uproot
-        elif len(axes) == 1 and hasattr(axes[0], "_to_boost_"):
-            self.__init__(axes[0]. _to_boost_())
+        elif len(axes) == 1 and hasattr(axes[0], "_to_boost_histogram_"):
+            self.__init__(axes[0]. _to_boost_histogram_())
             return
 
         # Keyword only trick (change when Python2 is dropped)
--- a/src/boost_histogram/_internal/hist.py
+++ b/src/boost_histogram/_internal/hist.py
@@ -133,6 +133,11 @@ class Histogram(object):
             self._from_histogram_object(axes[0])
             return
 
+        # Support objects that provide a to_boost method, like Uproot
+        elif len(axes) == 1 and hasattr(axes[0], "to_boost"):
+            self.__init__(axes[0].to_boost())
+            return
+
         # Keyword only trick (change when Python2 is dropped)
         with KWArgs(kwargs) as k:
             storage = k.optional("storage")


henryiii:
Tests are failing because of the new setup-python release: https://github.com/actions/setup-python/issues/171. We can pin to an older one until it's fixed. (Though the new release is really exciting, it now supports all versions of PyPy, in all flavors! Well, it would, anyway, if it worked...) :)

henryiii:
(actually intended to write this on another PR, but guess it's okay here.)

henryiii:
I'm bothered by the fact this is a public method. We might come up with a better method design later, so how about we select `_as_boot_histogram_`, and that makes it easier to replace later if we have a better method without littering up downstream project's namespaces if they want to avoid adding a public method. Uproot histogram could simply do `_as_boost_histogram_ = to_boost` and start supporting this.

henryiii:
@jpivarski would you be okay if we make this `_as_boost_histogram`, and uproot gets a `_as_boost_histogram_ = to_boost` line?

jpivarski:
> @jpivarski would you be okay if we make this `_as_boost_histogram`, and uproot gets a `_as_boost_histogram_ = to_boost` line?

Sure, if you do the pull request on Uproot, I'll accept it. You'll have to coordinate versions, but I can make an Uproot release candidate immediately after your PR.

src/boost_histogram/_internal/hist.py
```diff
@@ -133,6 +133,11 @@ def __init__(self, *axes, **kwargs):
             self._from_histogram_object(axes[0])
             return
 
+        # Support objects that provide a to_boost method, like Uproot
+        elif len(axes) == 1 and hasattr(axes[0], "to_boost"):
+            self.__init__(axes[0].to_boost())
```
henryiii, line 138:
```suggestion
        elif len(axes) == 1 and hasattr(axes[0], "_as_boost_histogram_"):
            self.__init__(axes[0]. _as_boost_histogram_())
```

HDembinski:
Hidden methods are better than public methods. The long name bothers me, why not call it simply `_to_boost`.

src/boost_histogram/_internal/hist.py
```diff
@@ -133,6 +133,11 @@ def __init__(self, *axes, **kwargs):
             self._from_histogram_object(axes[0])
             return
 
+        # Support objects that provide a to_boost method, like Uproot
+        elif len(axes) == 1 and hasattr(axes[0], "_as_boost_histogram_"):
+            self.__init__(axes[0]. _as_boost_histogram_())
```
henryiii, line 138:
```suggestion
        elif len(axes) == 1 and hasattr(axes[0], "_as_boost_"):
            self.__init__(axes[0]. _as_boost_())
```

henryiii:
My thought was this will never have to be typed, and someone coming across a method called `_to_boost_histogram_` would know exactly what library to look for, while `_to_boost_` could be a bit confusing. But this will always be on something that can be viewed as a histogram, so happy with any name, really.

HDembinski:
All those arguments are valid, but for consistency/symmetry with `to_numpy` etc., I would still prefer `_to_boost`.

henryiii:
Okay, though matching `to_<package_name>` was my exactly my proposal `_to_boost_histogram_`. We don't own the [`boost` package on PyPI](https://pypi.org/project/boost/) (though might be worth asking if we could get it, it would be nice to make it a namespace package for the boost libraries :) )

jpivarski:
In following this, I've also been thinking that "`_to_<package name>_`" would be best because then there are no special cases, like shortening "`boost_histogram`" → "`boost`". The whole package name is pasted in as-is, and that makes it more consistent.

Since this is an underscored name, brevity is not important: users will not be typing it. However, it's part of a protocol for developers, so it is important what that name is, and the need for consistency is heightened. (After all, these underscores don't mean "private"—a name like `__add__`, `__array__`, or `_ipython_key_completions_` has to be _more_ stable than even a user-facing interface.)

src/boost_histogram/_internal/hist.py
```diff
@@ -133,6 +133,11 @@ def __init__(self, *axes, **kwargs):
             self._from_histogram_object(axes[0])
             return
 
+        # Support objects that provide a to_boost method, like Uproot
+        elif len(axes) == 1 and hasattr(axes[0], "_to_boost_"):
+            self.__init__(axes[0]. _to_boost_())
```
henryiii, line 138:
```suggestion
        elif len(axes) == 1 and hasattr(axes[0], "_to_boost_histogram_"):
            self.__init__(axes[0]. _to_boost_histogram_())
```

henryiii: This would be the full name version.

henryiii:
Also, `.to_numpy` and `._to_boost_histogram_` are totally different; we provide `.to_numpy()`, but we do _not_ provide `._to_boost_histogram_` - we _already are_ a boost histogram! Libraries that implement this protocol will not necessary have a `to_numpy` also. (Uproot4 does, of course)

henryiii:
@HDembinski please confirm one of the two proposed names `_to_boost_histogram_` or `_to_boost_` (keeping in mind this is technically added by downstream libraries, not us, and the author of the main downstream library in Scikit-HEP is in favor of the more descriptive name). We can release 0.13 after this goes in. I’d like to have it sit for a few weeks while I’m off. I’ve already missed several important deadlines waiting for a response from you on this issue and the other one, including the SciPy proposal, my biweekly work report, and the birth of my daughter. If you insist on complete creative control over a library I proposed, wrote 87% of, and get paid for, you could at least try to be responsive when I ask for input. I did a same day release of 0.12 when you asked for it for your analysis. If you don’t have time for boost-histogram, I would always be happy to regain creative control. ;)

PS: I finally merged the other PR, #498, since I make the requested change related to the PR, updated the explanation, and the unrelated change of dropping `dd=` should be opened as an issue if you want it (though I would rather leave it matching uproot).

HDembinski:
I am sorry for letting you down on this and I apprecitate how fast you reacted on the release. I choose _to_boost_ then.

henryiii:
Thank you @HDembinski! Sorry I snapped a bit. Not getting much sleep here. 👶  I'm going to get a 0.13 release out so it can sit for a bit while I'm off before the deprecation/Py2 cleanup for 1.0 starting in March!

PR #487, opened 2020-12-21
fix: access getitem via string was broken (and attr access uses this too)

### LINKED ISSUE #486: [BUG] Modifying existing histogram through view stopped working
**Describe the bug**

A histogram modification through its `view` like below used to work.

```python
import boost_histogram as bh
import numpy as np

bins = [0, 1, 2]
hist = bh.Histogram(bh.axis.Variable(bins), storage=bh.storage.Weight())
yields = [3, 4]
var = [0.1, 0.2]
hist[...] = np.stack([yields, var], axis=-1)

hist.view().value /= 2
```

The above works with `boost-histogram` version 0.11.1, and breaks in the current master:

```
  File "test.py", line 10, in <module>
    hist.view().value /= 2
  File "[...]/boost_histogram/_internal/view.py", line 57, in fset
    self[name] = value
  File "[...]/boost_histogram/_internal/view.py", line 49, in __setitem__
    raise ValueError("Needs matching ndarray or n+1 dim array")
ValueError: Needs matching ndarray or n+1 dim array
```

The new behavior was introduced in #475.

**Steps to reproduce**

See above, tested with 41f56ef.

henryiii:
Fixes #486. Thanks to @alexander-held for testing the latests develop branch!

If I had static types here, I would have seen that ind could be a string, which probably would have kept me from making this mistake. But I might be giving myself too much credit there...

### DIFF
--- a/src/boost_histogram/_internal/view.py
+++ b/src/boost_histogram/_internal/view.py
@@ -37,6 +37,11 @@ class View(np.ndarray):
         )
 
     def __setitem__(self, ind, value):
+        # `.value` really is ["value"] for an record array
+        if isinstance(ind, str):
+            super(View, self).__setitem__(ind, value)
+            return
+
         array = np.asarray(value)
         if (
             array.ndim == super(View, self).__getitem__(ind).ndim + 1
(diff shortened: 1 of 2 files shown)


alexander-held:
Thank you! Credit to @matthewfeickert who set up [an action](https://github.com/alexander-held/cabinetry/blob/081c2308b844880b0cbebfbfed67e466b7e45fcf/.github/workflows/dependencies-head.yml#L111-L134) to run nightly checks, which caused me to spot this behavior.

PR #492, opened 2021-01-27
chore: update to pybind11 2.6.2, update formatting

henryiii:
- chore: update to pybind11 v2.6.2
- chore: pre-commit auto-update
- chore: add pyupgrade
- style: run pyupgrade
- chore: add isort
- style: run isort

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,6 +33,16 @@ repos:
   - id: cmake-format
     additional_dependencies: [pyyaml]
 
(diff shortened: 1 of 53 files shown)


henryiii:
@jpivarski I guess I should either take the writing example out of our examples. Or, maybe better, add uproot3 and use it for writing alongside the uproot 4 reading example?

jpivarski:
The minimal change would be to add `uproot3` and use that for writing. Uproot4-writing is another project that has to compete for time with Vector... It will happen sometime in the spring.

PR #497, opened 2021-02-09
chore: cleanup a bit for pybind11 2.6.x cleanups

henryiii:
Minor cleanup I noticed.


henryiii:
Oops, guess this wasn't fixable form upstream. Nevermind.

PR #505, opened 2021-02-14
chore: drop deprecated .rank

henryiii:
Just a small part of #503 - this is the only previously deprecated part.


henryiii:
Forgot there was an extra workaround to make this show deprecation warnings, so it's a little more cleanup than I thought. :)

PR #506, opened 2021-02-15
fix typo

kgizdov:
as the title says


henryiii:
@all-contributors please add @kgizdov for platform and bug.

PR #518, opened 2021-03-08
fix: restore metadata in repr

henryiii:
Restore metadata in all the reprs, as requested by @HDembinski a while back. Now that we have a way to store "long" metadata (not nice to print) via `__dict__`, showing metadata= isn't a problem. It was partially shown before, but was missing on some axes types (category and boolean, at least). This also provides a more structured method for building the repr, and is clearly named to indicate that Hist is using it too.

- fix: metadata repr restored
- refactor: use more consistent name


HDembinski:
Sounds nice

PR #537, opened 2021-03-16
fix: Mean/WeightedMean accumulator addition fix

### LINKED ISSUE #531: [BUG] Sample variance not summed correctly
**Describe the bug**

I get incorrect results when summing bins of `Mean` storage.

**Steps to reproduce**

```python
import numpy as np
import boost_histogram as bh

values = np.random.normal(loc=1.3, scale=0.1, size=1000)
print(f"sample mean: {values.mean()}, sample variance: {values.var(ddof=1)}")

# a single bin works fine
h = bh.Histogram(bh.axis.Regular(1, -1, 3), storage=bh.storage.Mean())
h.fill(values, sample=values)
print(f"sample mean from bh with Mean storage: {h[0].value} sample variance: {h[0].variance}")

# multiple bins do not sum correctly
h = bh.Histogram(bh.axis.Regular(10, -1, 3), storage=bh.storage.Mean())
h.fill(values, sample=values)
print(f"sample mean from bh with multi-bin Mean storage: {h.sum().value} sample variance: {h.sum().variance}")

sumv2 = sum(b.value**2 * b.count + b.sum_of_deltas_squared for b in h)
sumv = sum(b.value * b.count for b in h)
n = sum(b.count for b in h)
mean = sumv / n
variance = (sumv2 - sumv**2 / n) / (n - 1)
print(f"hand-calculated from bh mean: {mean}, variance {variance}")
```

<details><summary>Original Hist code</summary>

```python
import numpy as np
import hist

values = np.random.normal(loc=1.3, scale=0.1, size=1000)
print(f"sample mean: {values.mean()}, sample variance: {values.var(ddof=1)}")

# a single bin works fine
h = hist.Hist(hist.axis.Regular(1, -1, 3, name="value"), hist.storage.Mean())
h.fill(values, sample=values)
print(f"sample mean from bh with Mean storage: {h[0].value} sample variance: {h[0].variance}")

# multiple bins do not sum correctly
h = hist.Hist(hist.axis.Regular(10, -1, 3, name="value"), hist.storage.Mean())
h.fill(values, sample=values)
print(f"sample mean from bh with multi-bin Mean storage: {h.sum().value} sample variance: {h.sum().variance}")

sumv2 = sum(b.value**2 * b.count + b.sum_of_deltas_squared for b in h)
sumv = sum(b.value * b.count for b in h)
n = sum(b.count for b in h)
mean = sumv / n
variance = (sumv2 - sumv**2 / n) / (n - 1)
print(f"hand-calculated from hist mean: {mean}, variance {variance}")
```
</details>

An example output:
```
sample mean: 1.301933250738541, sample variance: 0.009653364754162593
sample mean from bh with Mean storage: 1.3019332507385406 sample variance: 0.009653364754162598
sample mean from bh with multi-bin Mean storage: 1.301933250738541 sample variance: 0.0051521413271981365
hand-calculated from hist mean: 1.301933250738541, variance 0.009653364754163354
```
which shows agreement within expected floating-point precision loss except for the multi-bin storage which is off by a factor of ~2.

henryiii:
Closes #531, implements https://github.com/boostorg/histogram/pull/308 .

Also fixes WeightedMean. Needs backport. Also supporting `__add__` on accumulators from the Python side since that's easier to test and work with in Python.

I'm thinking releasing 1.0.1 / 0.13.1 after this sounds like a good idea.

### DIFF
--- a/tests/test_accumulators.py
+++ b/tests/test_accumulators.py
@@ -1,3 +1,6 @@
+import pytest
+from pytest import approx
+
 import boost_histogram as bh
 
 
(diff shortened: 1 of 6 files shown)


tests/test_accumulators.py
```diff
@@ -85,3 +87,47 @@ def test_mean():
...
+
+def test_sum_weighed_mean():
+    a = bh.accumulators.WeightedMean()
+    a.fill([1, 2, 3])
+
+    b = bh.accumulators.WeightedMean()
+    b.fill([5, 6])
+
+    c = bh.accumulators.WeightedMean()
+    c.fill([1, 2, 3, 5, 6])
```
HDembinski, line 121: This test is incomplete if you don't use weights.

HDembinski: To clarify, if you don't use the original Boost Histogram code, then you should fully test the implementation thoroughly. If you rely on the original Boost Histogram code, then you can be more lax and just test the interface.

henryiii: Added weights, and added a test for the histogram summing as seen in the issue this fixes. As mentioned elsewhere, we have to have a copy because Boost.Histogram makes the internal memory storage private, but we have to be able to export this to a NumPy dtype, and the Python code has to internally convert back and forth, such as making an accumulator from a View, pickling, etc. You helped me do this in #194.

PR #543, opened 2021-03-22
tests: add requirements to dev-requirements

henryiii:
Missed updating this.


henryiii:
Seems like there's an issue with pulling manylinux. Will retry tomorrow.

PR #558, opened 2021-04-24
fix possible typo in sphinx_book_theme requirement

mayou36:
docs require sphinx_book_theme==0.38.0, but this version is not out as the newest is 0.1.0. I assume it's a typo and 0.0.38 was meant (which exists).

Fixed in the setup.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ extras = {
     "docs": [
         "Sphinx~=3.0",
         "myst_parser>=0.13",
-        "sphinx_book_theme==0.38.0",
+        "sphinx_book_theme==0.0.38",
         "nbsphinx",
         "sphinx_copybutton",
     ],
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ extras = {
     "docs": [
         "Sphinx~=3.0",
         "myst_parser>=0.13",
-        "sphinx_book_theme==0.0.38",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -53,7 +53,7 @@
     "docs": [
         "Sphinx~=3.0",
         "myst_parser>=0.13",
-        "sphinx_book_theme==0.38.0",
+        "sphinx_book_theme==0.0.38",
```
henryiii, line 56:
```suggestion
        "sphinx-book-theme>=0.0.33",
```
This would match docs/requirements.txt. (We should probably pin there and keep it looser here eventually)

mayou36:
Indeed, I wasn't aware of this file. Shouldn't we simply parse the docs/requirements.txt instead of having it duplicated in two places?

henryiii:
I'm not fond of parsing files in setup.py, it's a bit fragile and can't be moved to setup.cfg (it would be easy to add to setuptools, but they don't want to add it since it's bad practice - requirements files should be pinned, and extras should be loose). But that's an option. The docs extra is mostly an (unused) convenience to make docs locally, while the requirements file is needed for readthedocs because readthedocs can't build boost-histogram in the memory/time it has available. Ideally, we should probably pin versions exactly in the requirements.txt and provide loose versions in the setup.py file.

PR #559, opened 2021-04-25
fix: scaling, add logging, fix internal repr

### LINKED ISSUE #555: [BUG] Scaling Weight storage histograms
**Describe the bug**

When multiplying a histogram with Weight storage by a scalar, I expected the Variance to be scaled by the factor squared. The reasoning being that I should reasonably expect `h.fill(..., weight=w)` to be equivalent to `h.fill(...); h *= w`, as well as it would be more ergonomic to be able to just scale `h` than individually scale `h.view().value` and `h.view().variance`.

**Steps to reproduce**

```python
import hist
import numpy as np

w = 1e-2  # e.g. a cross section for a process
x = np.random.normal(loc=0.5, scale=0.2, size=1000)

h = (
    hist.Hist.new.Reg(3, 0, 1, name="x", label="axis")
    .Weight()
    .fill(x=x, weight=w)
)

print("Fill with weight", h.view())

h = (
    hist.Hist.new.Reg(3, 0, 1, name="x", label="axis")
    .Weight()
    .fill(x=x)
)

h *= w
print("Fill then scale", h.view())

h = (
    hist.Hist.new.Reg(3, 0, 1, name="x", label="axis")
    .Weight()
    .fill(x=x)
)

h.view().value *= w
h.view().variance *= w**2
print("Fill then scale as desired", h.view())
```
producing
```
Fill with weight WeightedSumView: (value, variance)
[(1.91, 0.0191) (5.8 , 0.058 ) (2.17, 0.0217)]
Fill then scale WeightedSumView: (value, variance)
[(1.91, 1.91) (5.8 , 5.8 ) (2.17, 2.17)]
Fill then scale as desired WeightedSumView: (value, variance)
[(1.91, 0.0191) (5.8 , 0.058 ) (2.17, 0.0217)]
```
(exact values will vary with random seed)

henryiii:
Closes #555. Added two more small changes to this PR.

### DIFF
--- a/src/boost_histogram/_internal/hist.py
+++ b/src/boost_histogram/_internal/hist.py
@@ -1,4 +1,5 @@
 import copy
+import logging
 import threading
 import typing
 import warnings
@@ -48,6 +49,8 @@ _histograms: Set[Type[CppHistogram]] = {
     _core.hist.any_mean,
     _core.hist.any_weighted_mean,
 }
+
+logger = logging.getLogger(__name__)
 
 
 CppAxis = NewType("CppAxis", object)
@@ -772,6 +775,7 @@ class Histogram:
                 assert isinstance(stop, int)
                 slices.append(_core.algorithm.slice_and_rebin(i, start, stop, merge))
 
+        logger.debug("Reduce with %s", slices)
         reduced = self._hist.reduce(*slices)
 
         if not integrations:
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -2,9 +2,11 @@
 
 ## UPCOMING
 
+* Fix scaling a weighted storage [#559][]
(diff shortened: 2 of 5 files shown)


src/boost_histogram/_internal/hist.py
```diff
@@ -772,6 +775,7 @@ def __getitem__(  # noqa: C901
                 assert isinstance(stop, int)
                 slices.append(_core.algorithm.slice_and_rebin(i, start, stop, merge))
 
+        logger.debug(f"Reduce with {slices}")
```
henryiii, line 778:
```suggestion
        logger.debug("Reduce with %s", slices)
```
Easier to backport, and may be slightly optimized if unused.

henryiii:
I really like Sourcery as a one time run, but the regular messages from it are annoying. Will probably turn it off.

PR #577, opened 2021-05-18
feat: support selecting on categorical axes (experimental)

henryiii:
Followup to #576, supporting sequences of bin numbers for categorical axes (needs more tests).


henryiii:
Also needs bh.loc support.

henryiii:
I'm going to merge this as an experimental feature so we can release for PyHEP. It currently does not support adding the removed axes to the flow bins, but eventually it should. It prints an "Experimental" warning if used.

PR #580, opened 2021-05-20
fix/feat: support scaling flow bins

### LINKED ISSUE #579: [BUG] Scaling histogram doesn't affect flow bins
```py
def test_scale_flowbins():    
    w = 1e-2
    x = np.random.normal(loc=0.4, scale=0.4, size=1000)
    y = np.random.normal(loc=0.6, scale=0.4, size=1000)
    
    h = bh.Histogram(
        bh.axis.Regular(3, 0, 1),
        bh.axis.Regular(3, 0, 1),
        storage=bh.storage.Weight(),
    )
    
    h.fill(x, y, weight=w)

    ref_value = np.array([h[bh.underflow,  0].value, h[1, 0].value, h[bh.overflow,  0].value])
    h = h * 5
    scale_value = np.array([h[bh.underflow,  0].value, h[1, 0].value, h[bh.overflow,  0].value])
    return np.allclose(ref_value * 5 , scale_value)
```

henryiii:
Closes #579.

This was an intentional design decision, based on the fact that `h + 1` shouldn't add 1 to the flow bins, but scaling makes more sense to apply to all.

### DIFF
--- a/src/boost_histogram/_internal/hist.py
+++ b/src/boost_histogram/_internal/hist.py
@@ -380,7 +380,7 @@ class Histogram:
                     )
                 )
         else:
-            view = self.view(flow=False)
+            view = self.view(flow=True)
             getattr(view, name)(other)
         self._variance_known = False
         return self
--- a/tests/test_histogram.py
+++ b/tests/test_histogram.py
@@ -1178,7 +1178,7 @@ def test_add_broadcast():
 
     h2 = h + [[1]]
     assert h2.sum() == 10 * 20
-    assert h2.sum(flow=True) == 10 * 20
+    assert h2.sum(flow=True) == 12 * 22
 
     h3 = h + np.ones((10, 20))
     assert h3.sum() == 10 * 20
--- a/tests/test_histogram_indexing.py
+++ b/tests/test_histogram_indexing.py
@@ -429,3 +429,29 @@ def test_single_flow_bin():
     assert h[3::sum] == 1
 
     assert h[1:2][sum] == 5
+
+
+# issue 579
+
+
+def test_scale_flowbins():
+    w = 1e-1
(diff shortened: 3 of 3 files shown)


henryiii:
@HDembinski

Currently, we have (idea 1):

```python
h + 1 # only adds to the normal bins
h * 1 # only scales the normal bins
```

With this PR (idea 2):

```python
h + 1 # only adds to the normal bins
h * 1 # scales all bins (flow too)
```

Another idea (idea 3):

```python
h + 1 # adds to all bins (flow too)
h * 1 # scales all bins (flow too)
```

The original idea was based on the fact that we always treat flow as hidden in Python, so adding +1 would be very surprising if it added 1 to the flow bins (which is poorly defined, since they are infinite - but one could argue that adding a constant to variable bins, etc is pretty useless too). But one could argue quite well that scaling should always include the flow bins, which leads into idea 2 (implemented here) or idea 3 being clearly less surprising and better. Note for above examples, `/` acts like `*` and inplace versions are included. I'm leaning toward 2, but 3 would be more consistent.

Of course, you can do any of these explicitly using an array, then you get to pick based on the size of the array if the flow bins are included. This is just for the shortcut.

andrzejnovak:
@HDembinski Can you give feedback? It would be good to resolve this sooner or later. I am in favour of 3) for consistency reasons.

HDembinski:
Flow bins should not be treated different. If you add 1 it should also include the flow bins. Same with multiplication. The moment you add a constant to a histogram, it stops being a histogram anyway. The least surprising is to treat all bins the same. I was aware that adding used to work differently and I would have been strongly against.

HDembinski:
> @HDembinski Can you give feedback? It would be good to resolve this sooner or later. I am in favour of 3) for consistency reasons.

Seems we agree then, thanks for pinging me.

PR #594, opened 2021-06-25
chore: update Boost.Histogram to 1.77

henryiii:
Adapting to the new count implementation.

### DIFF
--- a/include/bh_python/register_histogram.hpp
+++ b/include/bh_python/register_histogram.hpp
@@ -30,22 +30,6 @@
 #include <tuple>
 #include <vector>
 
-namespace boost {
-namespace histogram {
-namespace detail {
-
-using _atomic_type = histogram<vector_axis_variant, storage::atomic_int64>;
-
-template <>
-struct has_operator_rdiv_impl<_atomic_type, _atomic_type> : detect_base {
-    using type = mp11::mp_false;
-};
-
-} // namespace detail
-} // namespace histogram
-} // namespace boost
-// bh::detail::has_operator_rdiv<histogram_t, histogram_t>{}
-
 template <class S>
 auto register_histogram(py::module& m, const char* name, const char* desc) {
     using histogram_t = bh::histogram<vector_axis_variant, S>;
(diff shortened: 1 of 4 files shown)


include/bh_python/register_histogram.hpp
```diff
@@ -30,6 +30,22 @@
...
+namespace boost {
+namespace histogram {
+namespace detail {
+
+using _atomic_type = histogram<vector_axis_variant, storage::atomic_int64>;
+
+template <>
+struct has_operator_rdiv_impl<_atomic_type, _atomic_type> : detect_base {
+    using type = mp11::mp_false;
+};
```
henryiii, line 42: @HDembinski I'm having to work around the fact that `count` claims to have a division operator, but the implementation does not actually have one, so it reports true here then fails to compile. Same issue as the one fixed for `*` in https://github.com/boostorg/histogram/pull/328. Though `*` used to work, and `/` used to be not implemented, I believe.

henryiii:
After this is merged I can test https://github.com/boostorg/histogram/pull/323.

include/bh_python/register_histogram.hpp
HDembinski: Sorry again, I am getting sloppy. I am developing a fix here:
https://github.com/boostorg/histogram/pull/329
Could you please use that version instead of 1.77?

henryiii: Sure, thanks!

henryiii: Could you sync master with develop in Boost.Histogram? I can use master, develop has been force pushed to in the past, which breaks boost-histogram if we were pointing at a commit that doesn't exist anymore and gets cleaned up.

Otherwise this works.

henryiii:
@HDembinski this is just waiting for the develop commit to land in master.

I've started working on the accumulator branch but it will take a bit to get it to work, not ready yet.

include/bh_python/register_histogram.hpp
HDembinski: Sorry, I missed that, I merged develop into master now.

HDembinski:
> @HDembinski this is just waiting for the develop commit to land in master.

Sorry, it is done now.

PR #596, opened 2021-06-30
fix: improve __getattr__ error message

amangoel185:
The error message in \_\_getattr\_\_ had `not` instead of `no`.


henryiii:
Thanks!

PR #610, opened 2021-07-17
chore: use pip instead of docker for clang-format

henryiii:
Seeing if pre-commit.ci supports this.


henryiii:
Waiting on conda support for pre-commit.ci. I think it is planned.

henryiii:
https://github.com/conda-forge/clangdev-feedstock/pull/152 should make this work.

PR #613, opened 2021-07-21
docs: fix typo in Storages

amangoel185:
Change `storage` to `storages` in [storages.rst](https://github.com/scikit-hep/boost-histogram/blob/develop/docs/usage/storage.rst).


henryiii:
@all-contributors please add @amangoel185 for docs

PR #622, opened 2021-07-30
Minor typo fix

mayou36:
I think it should be the method `values`, not `value` (typo)
Removed a "be" (too much, typo?)

Also named value -> values (variances) to be consistent (it's also counts and not count). If you dislike this changes, I'll revert them


henryiii:
Thanks!

PR #627, opened 2021-08-18
fix: support getattr in custom AxesTuple subclasses

henryiii:
Fixes an issue we are having to work around in https://github.com/scikit-hep/hist/pull/288 .

The default setter now fails if the iterable is of the wrong length.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 #### Bug fixes
 * Support custom setters on AxesTuple subclasses. [#627][]
+* Throw an error when an AxesTuple setter is the wrong length (inspired by zip strict in Python 3.10) [#627][]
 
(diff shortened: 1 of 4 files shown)


docs/CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
 
 ## Version 1.1
 
+### Version 1.1.1
+
+#### User changes
+* Python 3.10 officially supported, with wheels.
+
+#### Bug fixes
+* Support custom setters on AxesTuple subclasses. [#627][]
```
henryiii, line 11:
```suggestion
* Support custom setters on AxesTuple subclasses. [#627][]
* Throw an error when an AxesTuple setter is the wrong length (inspired by zip strict in Python 3.10) [#627][]
```

PR #631, opened 2021-08-19
fix: return False if axis do not support protocol instead of raising

### LINKED ISSUE #626: [BUG] axis equality comparison errors with different types
**Describe the bug**

The equality comparison of the axis does not return False but errors (because the `__eq__` assumes that the other object supports the axis protocol)
`bh.axis.Regular(20, -3, 3) == []`

**Steps to reproduce**

```
import boost_histogram as bh

bh.axis.Regular(20, -3, 3) == object()
```

I already fixed it (checking for the attribute such as `Histogram` does) and added a test, fine if I make a PR with the fix?

jonas-eschle:
Closes #626 

(I removed the type: ignore as mypy complained, if that was needed for something, I'll add it back)

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,8 +10,10 @@
 #### Bug fixes
 * Support custom setters on AxesTuple subclasses. [#627][]
 * Throw an error when an AxesTuple setter is the wrong length (inspired by zip strict in Python 3.10) [#627][]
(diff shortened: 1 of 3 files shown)


henryiii:
Curious, I wonder if MyPy is smarter about this now, it used to not be able to infer that a hasattr enabled later code to use that attr.

henryiii (approved):

Issue #632, opened 2021-07-20
[FEATURE] Implement __sub__


andrzejnovak:

henryiii:
Missing only from _internal/hist.py, the views already support it.

PR #636, opened 2021-08-31
feat: add boilerplate sub implementation

amangoel185:
Basic implementation for #632.

TODO:

- [x] Add test for scalar
- [x] Add test for array
- [x] Add test for Histogram (expected to throw unless Boost.Histogram adds or supports sub)
- [x] Add test for int storage < 0 after subtraction (unsigned currently) - should throw, convert to double, or storages should be signed.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -6,14 +6,24 @@
 
 #### User changes
 * Python 3.10 officially supported, with wheels.
+* Support subtraction on histograms [#636][]
 
 #### Bug fixes
 * Support custom setters on AxesTuple subclasses. [#627][]
 * Throw an error when an AxesTuple setter is the wrong length (inspired by zip strict in Python 3.10) [#627][]
 * Fix error thrown on comparison with axis and non-axis object [#631][]
+* Static typing no longer thinks `storage=` is required [#604][]
 
+#### Developer changes
+* Support NumPy 1.21 for static type checking [#625][]
+* Use newer Boost 1.77 and Boost.Histogram 1.77+1 [#594][]
+
+[#594]: https://github.com/scikit-hep/boost-histogram/pull/594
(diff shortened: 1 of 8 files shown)


amangoel185:
We need to implement `__isub__` but it has to be in accordance with Boost.Histogram

henryiii:
@HDembinski Question for you: should Hist - Hist be supported? I'd only want to support it in boost-histogram if Boost.Histogram supports it, or you have a good reason to support it only in boost-histogram. It's not hard to be explicit - `h1 - h2.view(flow=True)` for example should do it since subtracting an array should be supported.

Also, maybe we should switch to using signed storage so that subtraction (and negative weights?) work without having to change storages or throw an error. Technically, NumPy uses signed ints, IIRC.

HDembinski:
I think Hist should support what boost-histogram supports. The added benefit is that hist - hist can check whether this actually makes sense, it should fail if the axes are not the same.

henryiii:
@HDembinski Sorry, my notation was sloppy. I wasn't suggesting Hist and boost-histogram diverge, I was just asking if a histogram histogram subtraction, `bh.Histogram(...) - bh.Histogram(...)`, should be supported. I was under the impression Boost.Histogram didn't currently support negating and subtracting histograms, but looking at the code, it seems to just fine. So we should add this in boost-histogram!

henryiii:
Okay, needs tests now. @amangoel185 can you look into adding tests as outlined above? Or I might be able to later.

henryiii:
Weighted and AtomicInt64 histograms do not support subtraction, so they don't support subtraction here either.

henryiii:
@all-contributors please add @amangoel185 for code

henryiii:
After this goes in, I think we should cut a patch release - there have been a few nice fixes, and Python 3.10 support needs to ship.

Issue #644, opened 2021-09-15
Slicing a large histogram is slow


henryiii:
Hi @henryiii @HDembinski ,

I assume the following is related, if not please correct me and I open a fresh new issue...
We noticed in the scope of our analysis that `__getitem__` is a performance "hurdle" for high dimensional histograms (imagine: dataset axis of O(1000) dataset, category axis of O(100) categories and systematic axis of O(100) shifts).

I will put a snippet here, which makes the performance visible:
```python
import boost_histogram as bh

h = bh.Histogram(
    bh.axis.StrCategory([str(i) for i in range(100)]),  # e.g. datasets
    bh.axis.StrCategory([str(i) for i in range(100)]),  # e.g. categories
    bh.axis.StrCategory([str(i) for i in range(100)]),  # e.g. systematics
    bh.axis.Regular(100, 0, 500),
)

# let's fill a dummy value
h[...] = 1.0

# now the __getitem__ performance:
%timeit h[bh.loc("42"), bh.loc("42"), bh.loc("42"), :].view()
4.08 s ± 61.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit h.view()[h.axes[0].index("42"), h.axes[1].index("42"), h.axes[2].index("42"), :]
20.3 µs ± 669 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

Currently we use the second option, since on a larger analysis scale with multiple of these huge histograms this results in a difference of O(hours) and O(seconds) for histogram manipulation, such as grouping datasets to physics processes. However the first one is (obviously) a lot more convenient to use. 
I think this would be a major improvement, especially for the best usability of `hist` and `boost_histogram` for large-scale analysis. 

Best, Peter

_Originally posted by @pfackeldey in https://github.com/scikit-hep/boost-histogram/issues/149#issuecomment-919281492_

henryiii:
@pfackeldey This is a bit different, vectorizing getitem refers to something else, slices are fine. The second loop will always outperform the first loop above, since it's doing a lot less. The second loop is not even allocating new memory - it's just a view of the memory with different offsets and strides - a new histogram always needs new memory (if it has flow bins, that's logically impossible, without flow bins, it hasn't been worth adding).

I think we should do a bit of profiling here to see if anything could be improved (4 _seconds_ does seem a little overboard), and while I've been very reluctant to add too many axes types, maybe we could have a `flow=False` option for Str/IntCategory.

[...]

PR #645, opened 2021-09-15
fix: faster picking when no slices are made

henryiii:
Helps with #644. Ideally, I think the order of picking and slicing should be flipped, which would probably be faster. Empty reduce should still be swapped for a copy, though, so this code code is still needed even in that case.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,8 +1,8 @@
 # What's new in boost-histogram
 
-## Version 1.1
+## Version 1.2
 
-### Version 1.1.1
+### Version 1.2.0
 
 #### User changes
 * Python 3.10 officially supported, with wheels.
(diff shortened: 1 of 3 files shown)


henryiii:
Current benchmark info:

```
---------------------------------------------------------------------------------------------- benchmark 'Pick': 2 tests ----------------------------------------------------------------------------------------------
Name (time in us)               Min                     Max                    Mean                 StdDev                  Median                    IQR            Outliers         OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_pick_only             469.6110 (1.0)       10,963.2400 (1.0)          548.6303 (1.0)         335.7542 (1.0)          496.3605 (1.0)          58.5930 (1.0)         7;153  1,822.7210 (1.0)        1054           1
test_pick_and_lice     752,036.0140 (>1000.0)  794,440.9860 (72.46)    768,945.9046 (>1000.0)  15,980.4104 (47.60)    764,982.2400 (>1000.0)  18,759.4093 (320.16)        2;0      1.3005 (0.00)          5           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

PR #680, opened 2021-11-29
fix: strict AxesTuple constructor

### LINKED ISSUE #678: [BUG] (Discussion) Input validation/preprocessing in AxesTuple
**Describe the bug**

Currently, AxesTuple inherits the constructor behavior of tuple in that it requires an iterable. While this makes sense and is consistent, it can create a hard-to-spot bug introduced by the fact that a _single_ axis is also iterable and therefore if not an iterable of one or more axes but a single axis is given, it won't fail on instantiation. However, it will fail when trying to use any attribute.

I would propose to modify the constructor to be "axis-like" aware and convert a single axis automatically to an iterable with a single axis first.

The reason why I think it's fine to do is because the intention of `AxesTuple(axis1)` will always be actually `AxesTuple([axis1])`.

Alternatively, this could just error when constructing the `AxesTuple`

Or what do you think?

**Steps to reproduce**

```python
import boost_histogram as bh

axis1 =  hist.axis.Regular(50, -5, 5)

axes = bh.axis.AxesTuple([axis1])  # correct way, works

axes_wrong = bh.axis.AxesTuple(axis1)  # wrong! But does not throw an error here

# THIS will error
axes_wrong.edges  # or any other attribute
```

henryiii:
Closes #678.

Adding the check actually helped us, too - `.metadata` was incorrectly
returning an `AxesTuple` instead of a `tuple`, and there was an extra unneeded
`AxesTuple` construction.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## Version 1.2
 
+### Version 1.2.2
+
+#### User changes
+
+* PyPy 3.8 now supported. [#677][]
+* The GIL is released a little more often now. [#662][]
+* AxesTuple does not allow construction of non-axes. [#680][]
+
+[#662]: https://github.com/scikit-hep/boost-histogram/pull/662
+[#677]: https://github.com/scikit-hep/boost-histogram/pull/677
+[#680]: https://github.com/scikit-hep/boost-histogram/pull/680
+
 ### Version 1.2.1
 
 #### User changes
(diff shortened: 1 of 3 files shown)


henryiii:
FYI, even though this is a minor fix, Hist was relying on the buggy behavior, so the next release needs to at least do a minor bump to avoid breaking Hist. :facepalm:. See https://github.com/scikit-hep/hist/pull/368.

PR #689, opened 2022-01-17
fix: protect Cat indexing from missing values

### LINKED ISSUE #387: Indexing categories
When filling, an int or str category will put values not in the axes either in a new bin (growth), a flow bin, or ignore it if overflow is off. However, the Python index method does the same thing - for example:

```python
h = bh.Histogram(bh.axis.StrCategory(["a", "b"]))
h.fill(["c"])
print(h[bh.loc("d")]) # Returns 1.0
```
Also,  if growth was on, this will not grow the axes, though at least one user expected that in the past.
Should Python indexes into non-existing string or int categories throw a TypeError or KeyError? I think this would be less error-prone. You could still use `bh.overflow` to access the flow bin explicitly.

Unlike axes with both an underflow and overflow, the categories represent a set of items, and Python tools for this sort of thing (like a dict) generally throw errors when you try to access a non-existent item (but not when setting it, so fill is still consistent).

@HDembinski ?

andrzejnovak:
Category indexing will return `len(index)+1` for missing index, which creates messy error messages down the line.
image

This should catch it correctly

image

Closes #387.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,6 +8,9 @@ on:
     - master
     - develop
 
+env:
+  FORCE_COLOR: 3
+
 jobs:
   clang-tidy:
     name: Clang-Tidy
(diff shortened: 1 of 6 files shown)


henryiii:
Related to #387. This will likely have a performance impact on accessing histograms, should be checked.

andrzejnovak:
@henryiii fixed the tests

henryiii:
We are throwing a KeyError for a missing Category, and not throwing for Integer or other axes.

PR #692, opened 2022-01-29
chore: bump black to first non-beta version

henryiii:


henryiii:
Applying black 22.1 to 18d12998dc24b52b42f101055cba39e78517f36c caused a type ignore commit to be removed. Weird, might investigate later. I've fixed up the strange multiple union-attr's and rerun black, and it seems happy.

PR #695, opened 2022-02-07
fix: wrong calculation for WeightedMean variance NaN's via UHI

henryiii:
Brought up on Gitter. The old calculation could not have been right, and was not what is described above from the UHI.

### DIFF
--- a/src/boost_histogram/_internal/hist.py
+++ b/src/boost_histogram/_internal/hist.py
@@ -1067,11 +1067,12 @@ class Histogram:
             return view if self._variance_known else None
 
         if hasattr(view, "sum_of_weights"):
+            valid = view.sum_of_weights**2 > view.sum_of_weights_squared  # type: ignore[union-attr]
             return np.divide(  # type: ignore[no-any-return]
                 view.variance,  # type: ignore[union-attr]
                 view.sum_of_weights,  # type: ignore[union-attr]
                 out=np.full(view.sum_of_weights.shape, np.nan),  # type: ignore[union-attr]
-                where=view.sum_of_weights > 1,  # type: ignore[union-attr]
+                where=valid,
             )
 
         if hasattr(view, "count"):
--- a/tests/test_storage.py
+++ b/tests/test_storage.py
@@ -1,3 +1,5 @@
+import math
+
 import numpy as np
 import pytest
 from numpy.testing import assert_array_equal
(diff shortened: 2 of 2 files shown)


src/boost_histogram/_internal/hist.py
```diff
@@ -1071,7 +1071,7 @@ def variances(self, flow: bool = False) -> Optional["np.typing.NDArray[Any]"]:
                 view.variance,  # type: ignore[union-attr]
                 view.sum_of_weights,  # type: ignore[union-attr]
                 out=np.full(view.sum_of_weights.shape, np.nan),  # type: ignore[union-attr]
-                where=view.sum_of_weights > 1,  # type: ignore[union-attr]
+                where=view.sum_of_weights**2 > view.sum_of_weights_squared,
```
henryiii, line 1074: What the... Black, this comment was needed!

henryiii: Opened an issue at https://github.com/psf/black/issues/2873, and worked around this by moving this subexpression to a variable, which is a bit more readable, too - but it should not have deleted this comment!

src/boost_histogram/_internal/hist.py
```diff
@@ -1067,11 +1067,12 @@ def variances(self, flow: bool = False) -> Optional["np.typing.NDArray[Any]"]:
             return view if self._variance_known else None
 
         if hasattr(view, "sum_of_weights"):
+            valid = view.sum_of_weights**2 > view.sum_of_weights_squared  # type: ignore[union-attr]
```
henryiii, line 1070: Actually, I think this can just be:

```suggestion
            valid = view.sum_of_weights != view.sum_of_weights_squared  # type: ignore[union-attr]
```

This will either both be 0, or both be identical (counts==1 exactly), or this is a valid calculation. Comparing floating point numbers is a little scary, but I think this is fine.

henryiii: Also I need to add a test for this 😳

HDembinski:
I see you already merged this, so I guess I don't have to check it?

henryiii:
You can still check it, it was just a bad bug and the new behavior is at least closer to correct; we can always make a followup PR. Any ideas for further situations to test are always welcome. :)

Starting a discussion (https://github.com/scikit-hep/boost-histogram/discussions/701) was a good idea.

PR #698, opened 2022-02-07
fix: toml is not used anymore, replaced by tomli

henryiii:


henryiii:
I think this happened because pytest 7 was the last package to move from toml -> tomli, and our check for toml started failing. I'm just removing the check, since setuptools_scm now has a hard dependency on tomli.

PR #704, opened 2022-02-13
fix: pypy tests fail

henryiii:
On PyPy, linux only, there is a UserWarning on NumPy import. This _breaks_ pytest when warnings are turned into errors, even if you ignore this one, becuase it happens in the conftest.py.

```
>>>> import boost_histogram as bh
/opt/_internal/pypy3.8-7.3.7/lib/pypy3.8/site-packages/numpy/core/_add_newdocs.py:2826: UserWarning: add_newdoc was used on a pure-python object <bound method __class_getitem__ of <class 'numpy.ndarray'>>. Prefer to attach it directly to the source.
  add_newdoc('numpy.core.multiarray', 'ndarray', ('__class_getitem__',
/opt/_internal/pypy3.8-7.3.7/lib/pypy3.8/site-packages/numpy/core/_add_newdocs.py:6128: UserWarning: add_newdoc was used on a pure-python object <bound method __class_getitem__ of <class 'numpy.dtype'>>. Prefer to attach it directly to the source.
  add_newdoc('numpy.core.multiarray', 'dtype', ('__class_getitem__',
/opt/_internal/pypy3.8-7.3.7/lib/pypy3.8/site-packages/numpy/core/_add_newdocs.py:6640: UserWarning: add_newdoc was used on a pure-python object <bound method __class_getitem__ of <class 'numpy.number'>>. Prefer to attach it directly to the source.
  add_newdoc('numpy.core.numerictypes', 'number', ('__class_getitem__',
>>>>
```

This was breaking the wheel build, therefore a blocking issue for release.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,6 @@ write_to = "src/boost_histogram/version.py"
 minversion = "6.0"
 junit_family = "xunit2"
 addopts = ["-ra", "--showlocals", "--strict-markers", "--strict-config"]
-filterwarnings = ["error"]
 xfail_strict = true
 testpaths = ["tests"]
 required_plugins = ["pytest-benchmark"]


henryiii:
Since pytest dies without a very useful note why:

```pytb
ImportError while loading conftest '/project/tests/conftest.py'.
/project/tests/conftest.py:3: in <module>
    import boost_histogram as bh
/tmp/tmp.OO0mtwqra0/venv/lib/pypy3.8/site-packages/boost_histogram/__init__.py:1: in <module>
/tmp/tmp.OO0mtwqra0/venv/lib/pypy3.8/site-packages/boost_histogram/accumulators.py:1: in <module>
    from ._core.accumulators import (  # pylint: disable=import-error
E   ImportError: initialization failed
```

I used `git bisect` along with local cibuildwheel runs to find the breaking commit, which was cdbfabb1c22f5545bf3900be01f2025411e699f1 - which looked completely innocuous. I tried loading the wheel (and cibuildwheel doesn't output the wheel when you fail a test, so had to build without testing), then I saw the warnings, and traced it down to this.

PR #708, opened 2022-02-15
fix: regression in #662 causing segfaults

henryiii:
Important fix, requiring 2.3.1.

Nicer error message and avoids catching a C++ error if possible.

### DIFF
--- a/include/bh_python/register_histogram.hpp
+++ b/include/bh_python/register_histogram.hpp
@@ -190,16 +190,17 @@ auto register_histogram(py::module& m, const char* name, const char* desc) {
 
         .def("reduce",
              [](const histogram_t& self, py::args args) {
+                 auto commands
+                     = py::cast<std::vector<bh::algorithm::reduce_command>>(args);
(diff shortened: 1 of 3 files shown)


include/bh_python/register_histogram.hpp
```diff
@@ -190,16 +190,17 @@ auto register_histogram(py::module& m, const char* name, const char* desc) {
...
+                 return bh::algorithm::reduce(self, commands);
              })
 
         .def("project",
              [](const histogram_t& self, py::args values) {
+                 auto cpp_values = py::cast<std::vector<unsigned>>(values);
                  py::gil_scoped_release release;
-                 return bh::algorithm::project(self,
-                                               py::cast<std::vector<unsigned>>(values));
+                 return bh::algorithm::project(self, cpp_values);
```
henryiii, line 203: This is the actual fix, I was running Python API via `py::cast` inside a GIL-less section! I'm also now catching this error before it happens in Python, which should make limited EH environments like pyodide happier.

PR #718, opened 2022-03-10
fix: temporary patch on broken sum with empty axis

henryiii:
Fix for https://github.com/scikit-hep/hist/pull/387 that will hopefully be properly fixed upstream soon - see https://github.com/boostorg/histogram/issues/353. You can access the
accumulator type from the storage type which makes testing easier, too.

Improves typing around accumulators - they allow being called "empty" in C++, but that was not reflected in the static typing.


henryiii:
Same bug https://github.com/jupyter/nbconvert/issues/1725 showing up.

PR #758, opened 2022-06-27
Updated numpy.testing asserts to pytest.approx

gohil-jay:


src/boost_histogram/_internal/hist.py
```diff
@@ -23,6 +24,7 @@
 )
 
 import numpy as np
+from pytest import approx
```
henryiii, line 27: We can't use pytest inside the library itself. There are tools for this like `np.allclose`. Also `compare` should return True/False, not errors / asserts. Let's move the changes to hist.py into a different PR, as that's not related to using `approx` in the tests. :)

gohil-jay:
Alright, the first approx PR is here:
https://github.com/scikit-hep/boost-histogram/pull/760

I'll send another updated one (without pytest and true/false return) for the comparison once #760 is merged!

gohil-jay:
The separate PR for the comparison is ready here in this branch (without pytest and true/false return):
https://github.com/gohil-jay/boost-histogram/tree/issue157-1

PR #760, opened 2022-06-29
Updated numpy.testing asserts to pytest.approx

gohil-jay:


tests/test_pickle.py
```diff
@@ -162,7 +162,7 @@ def test_storage(benchmark, copy_fn, storage, extra):
         hist.fill(x, weight=np.arange(2 * n + 4) + 1, sample=np.arange(2 * n + 4) + 1)
 
     new = benchmark(copy_fn, hist)
-    assert_array_equal(hist.view(True), new.view(True))
+    assert hist.view(True) == approx(new.view(True))
```
henryiii, line 165: Interesting! This is failing due to the fact a single item in the array returns a custom object, which it can't compare. If we defined item, maybe it could. But a simple fix probably is to just put `np.asarray()` around both sides. Technically, before it was a non-approx comparison, but I think it's fine as approx.

```suggestion
    assert np.asarray(hist.view(True)) == approx(np.asarray(new.view(True)))
```

Not totally sure what this does with structured data, guessing it works though.

gohil-jay:
Seems alright now (?)

henryiii:
Yes, let's go with it!

PR #762, opened 2022-07-01
Added comparison function as histogram operation

gohil-jay:


src/boost_histogram/_internal/hist.py
```diff
@@ -337,6 +338,20 @@ def ndim(self) -> int:
         """
         return self._hist.rank()
 
+    def compare(self, hist2) -> bool:
```
henryiii, line 341:
```suggestion
    def compare(self, hist2: "Histogram") -> bool:
```

gohil-jay:
Updated PR: #763

PR #763, opened 2022-07-01
feat: add pytest comparison fixture

gohil-jay:
image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,9 @@ dependencies = [
     "numpy>=1.21.3",
 ]
 
+[project.entry-points.pytest11]
+boost_histogram = "boost_histogram._pytest"
+
 [project.urls]
 "Bug Tracker" = "https://github.com/scikit-hep/boost-histogram/issues"
 Changelog = "https://boost-histogram.readthedocs.io/en/latest/CHANGELOG.html"
@@ -150,6 +153,13 @@ addopts = [
 filterwarnings = [
   "error",
   "default:could not create cache path:pytest.PytestCacheWarning",
+  # Editable installs import the _boost_histogram_editable shim at interpreter
+  # startup (via .pth), so pytest cannot assertion-rewrite it once our pytest11
+  # plugin is registered. We do not want our own library rewritten anyway.
+  # pytest's fix for this (PR #10231, v7.2.0) only skips setuptools-style
+  # "__editable__*" finders, not scikit-build-core's "_<pkg>_editable" name.
+  # Fixed in pyest 7.2.0
+  "ignore:Module already imported so cannot be rewritten:pytest.PytestAssertRewriteWarning",
 ]
 xfail_strict = true
 testpaths = ["tests"]
--- a/src/boost_histogram/_pytest.py
+++ b/src/boost_histogram/_pytest.py
@@ -1,0 +1,171 @@
+"""
+A pytest plugin that produces readable explanations when an ``assert a == b``
+between two :class:`~boost_histogram.Histogram` (or axis) objects fails.
+
+It is registered as a ``pytest11`` entry point (see ``pyproject.toml``), so it
+loads automatically for any project that has ``boost-histogram`` installed. The
+hook is inert for every comparison that is not a ``==`` between two histograms
+or two axes, so it never interferes with unrelated assertions.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+import numpy as np
+
+from .axis import Axis
+from .histogram import Histogram
+
+if TYPE_CHECKING:
+    from numpy.typing import NDArray
+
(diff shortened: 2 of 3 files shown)


src/boost_histogram/_internal/hist.py
```diff
@@ -337,6 +338,24 @@ def ndim(self) -> int:
...
+                            return ""
+                        else:
+                            return "The axes [" + "\033[91m" + str(list(self.axes)) + " and " + str(list(hist2.axes)) + "\033[0m" + "] are not equal."
+                    else:
+                        return "The storage types [" + "\033[91m" + str(re.search("(?<=storage=).*", str(self.view))[0].split('(')[0]) + " and " + str(re.search("(?<=storage=).*", str(hist2.view))[0].split('(')[0]) + "\033[0m" + "] are not equal."
+                else:
+                    return "The histogram contents : \n" + "\033[91m" + str(self.variances()) + "\033[0m" + "\nand\n" + "\033[91m" + str(hist2.variances()) + "\033[0m" + "\nare not equal."
+            else:
+                return "The histogram contents : \n" + "\033[91m" + str(self.view()) + "\033[0m" + "\nand\n" + "\033[91m" + str(hist2.view()) + "\033[0m" + "\nare not equal."
+            else:
```
henryiii, line 356: No else after return, please! This would look much cleaner. Also, the final else block is misaligned. (CC @jpivarski this is an example of else after return. :)

I would then add "not" in front of each if, and place the return immediately after. Now you'll have a non-indented linear chain of checks, rather than this sideways pyramid. :)

src/boost_histogram/_internal/hist.py
```diff
@@ -337,6 +338,24 @@ def ndim(self) -> int:
...
 
+    def compare(self, hist2: "Histogram") -> bool:
+        if (np.allclose(self.view().shape, hist2.view().shape)):
+            if (np.allclose(self.view(), hist2.view())):
+                if (np.allclose(self.variances(), hist2.variances())):
+                    if (re.search("(?<=storage=).*", str(self.view))[0].split('(')[0] == re.search("(?<=storage=).*", str(hist2.view))[0].split('(')[0]):
+                        if (list(self.axes)==list(hist2.axes)):
+                            return ""
+                        else:
+                            return "The axes [" + "\033[91m" + str(list(self.axes)) + " and " + str(list(hist2.axes)) + "\033[0m" + "] are not equal."
```
henryiii:
```suggestion
                            return f"The axes [\033[91m{list(self.axes)} and {list(hist2.axes)}\033[0m] are not equal."
```

F-strings were made for this. :) Also I'd pull out the color formatting into a variable, like `red` and `reset`. Actually, there are lots of concerns with color formatting (needs to be disablable and such), at least rich can auto-format this, so probably leave the color formatting off for the moment.

src/boost_histogram/_internal/hist.py
gohil-jay: Alright makes sense! Will make the changes :)

henryiii: FYI, what I described is called "guard style". You check for the failure and immediately break the control flow (usually via `return`); everything after that is now guarded from that condition. It keeps logic simpler and reduces indentation.

Ruby actually has syntax for it, `return <value> if <condition>` can be written on one line, and the inverse is `return <value> unless <condition>`. Python never got custom syntax, since `if <condition>: return <value>` isn't actually any longer.

src/boost_histogram/_internal/hist.py
henryiii: I bet if you do `import rich; rich.print(h.compare(h2))` this would still be colorful. ;)

src/boost_histogram/_internal/hist.py
gohil-jay: Interesting... Thanks for the insight!

src/boost_histogram/_internal/hist.py
gohil-jay: Yes I was going to do something similar (with termcolor), but then though I'd do it without importing a library..

gohil-jay:
@henryiii, pre-commit throws few errors... Any feedback/suggestion?

henryiii:
It's complaining because you are using the result of re.search without checking to see if it has a result first. I think the best fix, though, is to use `._storage_type`, which contains the storage type of the histogram, rather than trying to parse it out of the repr.

gohil-jay:
I've been looking for something like `_storage_type` since the beginning (couldn't find it on docs or book [here](https://boost-histogram.readthedocs.io/_/downloads/en/latest/pdf/))! But alright now it should work :)

gohil-jay:
Seems alright now (?)

henryiii:
Note: the goal is to support `pytest_assertrepr_compare` https://docs.pytest.org/en/latest/how-to/assert.html#defining-your-own-explanation-for-failed-assertions

PR #771, opened 2022-07-28
fix: minor optimizations for ufuncs

henryiii:
Discovered minor improvements when preparing teaching material for next week.

### DIFF
--- a/src/boost_histogram/_internal/view.py
+++ b/src/boost_histogram/_internal/view.py
@@ -103,84 +103,106 @@ class WeightedSumView(View):
     def __array_ufunc__(
         self, ufunc: Ufunc, method: str, *inputs: Any, **kwargs: Any
     ) -> "np.typing.NDArray[Any]":
-
-        # This one is defined for record arrays, so just use it
-        # (Doesn't get picked up the pass-through)
-        if ufunc is np.equal and method == "__call__" and len(inputs) == 2:
-            return ufunc(np.asarray(inputs[0]), np.asarray(inputs[1]), **kwargs)  # type: ignore[no-any-return]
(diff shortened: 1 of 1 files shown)


src/boost_histogram/_internal/view.py
```diff
@@ -103,104 +103,130 @@ class WeightedSumView(View):
     def __array_ufunc__(
         self, ufunc: Ufunc, method: str, *inputs: Any, **kwargs: Any
     ) -> "np.typing.NDArray[Any]":
-
-        # This one is defined for record arrays, so just use it
-        # (Doesn't get picked up the pass-through)
-        if ufunc is np.equal and method == "__call__" and len(inputs) == 2:
-            return ufunc(np.asarray(inputs[0]), np.asarray(inputs[1]), **kwargs)  # type: ignore[no-any-return]
+        # Avoid infinite recursion
+        raw_inputs = [np.asarray(x) for x in inputs]
```
henryiii, line 107: This allows us to avoid special casing `==`; it was recusing before.

src/boost_histogram/_internal/view.py
```diff
@@ -103,104 +103,130 @@ class WeightedSumView(View):
...
-            and len(inputs) == 1
+            and len(raw_inputs) == 1
             and ufunc in {np.negative, np.positive}
         ):
-            (result,) = kwargs.pop("out", [np.empty(self.shape, self.dtype)])
+            (result,) = (
+                kwargs.pop("out")
+                if "out" in kwargs
+                else [np.empty(self.shape, self.dtype)]
+            )
```
henryiii, line 119: Avoids a temporary allocation if `out` is set.

PR #780, opened 2022-08-15
Swapped all numpy assert_allclose & assert_array_equal to pytest approx in tests

gohil-jay:
Addresses https://github.com/scikit-hep/boost-histogram/commit/2c81964d8f422396963fbbcaece650f5d787eefc commit further.

I realized that there are other asserts that are imported directly which I missed in previous commit, so this covers all of them.

### DIFF
--- a/tests/test_axis.py
+++ b/tests/test_axis.py
@@ -5,7 +5,6 @@ import copy
 
 import numpy as np
 import pytest
-from numpy.testing import assert_allclose, assert_array_equal
 from pytest import approx
 
 import boost_histogram as bh
@@ -253,8 +252,8 @@ class TestRegular(Axis):
         a = bh.axis.Regular(2, 1.0, 2.0)
         ref = [1.0, 1.5, 2.0]
         for i in range(2):
-            assert_allclose(a.bin(i), ref[i : i + 2])
-            assert_allclose(a[i], ref[i : i + 2])
+            assert a.bin(i) == approx(ref[i : i + 2])
+            assert a[i] == approx(ref[i : i + 2])
 
         assert a[-1] == a[1]
         with pytest.raises(IndexError):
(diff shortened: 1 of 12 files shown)


tests/test_axis.py
```diff
@@ -793,7 +792,7 @@ def test_getitem(self, ref, growth):
     def test_iter(self, ref, growth):
         Cat = bh.axis.StrCategory if isinstance(ref[0], str) else bh.axis.IntCategory
         a = Cat(ref, growth=growth)
-        assert_array_equal(a, ref)
+        assert a == approx(ref)
```
henryiii, line 795:
```suggestion
        assert np.asarray(a) == approx(ref)
```

Quick fix. But is it possible to control how pytest treats approx?

HDembinski:
Why do you replace the numpy assserts with pytest.approx? The messages produced by numpy.assert_allclose etc are better than those from pytest.approx, and numpy.assert_allclose works on array-like types, too.

gohil-jay:
I believe this came up while working on the comparison issue but I don't remember the exact details on why! @henryiii do you want to chime in here?

henryiii:
The pytest.approx output has native array support and is much better, cleaner, integrated into pytest. The only downside is you do have to make sure it is an array, it can't cast to array for you (as it's general).

Given a histogram view array([2., 1., 1., 2.]) compared against a wrong expected [2, 1, 2, 2]:

Old — `numpy.testing.assert_allclose`
```
assert_allclose(h.view(), [2, 1, 2, 2])
E       AssertionError:
E       Not equal to tolerance rtol=1e-07, atol=0
E
E       Mismatched elements: 1 / 4 (25%)
E       Mismatch at index:
E        [2]: 1.0 (ACTUAL), 2 (DESIRED)
E       Max absolute difference among violations: 1.
E       Max relative difference among violations: 0.5
E        ACTUAL: array([2., 1., 1., 2.])
E        DESIRED: array([2, 1, 2, 2])
```

New — `== pytest.approx(...)`
```
assert h.view() == approx([2, 1, 2, 2])
E       assert array([2., 1., 1., 2.]) == approx([2 ± 2... 2 ± 2.0e-06])
E
E         comparison failed. Mismatched elements: 1 / 4:
E         Max absolute difference: 1.0
E         Max relative difference: 1.0
E         Index | Obtained | Expected
E         2     | 1.0      | 2 ± 2.0e-06
```

Issue #157, opened 2019-10-18
PyTest diffs


henryiii:
I *believe* we can add a custom diff printer for PyTest such that we can see what is actually different when a histogram is not the same as the other. Look into this and see if in fact it is possible.

PR #781, opened 2022-08-15
Storage function to fetch storage type

gohil-jay:
Addresses issue faced while solving https://github.com/scikit-hep/boost-histogram/issues/157 issue.

Simply added a call function that returns storage type.

This is also needed to correctly create an empty copy an arbitrary histogram:

```python
new_hist = bh.Histogram(*old_hist.axes, storage=old_hist.storage_type())
```

Backward compatibility shim needed for a few releases since things like Hist already are having to use this; it needs to be officially public API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ histograms can be plotted via any compatible library, such as [mplhep][].
   * `*=`: Multiply by a scaler (not all storages) (`hist * scalar` and `scalar * hist` supported too)
   * `/=`: Divide by a scaler (not all storages) (`hist / scalar` supported too)
   * `.kind`: Either `bh.Kind.COUNT` or `bh.Kind.MEAN`, depending on storage
-  * `.storageType`: Fetch the histogram storage type in string
+  * `.storage_type`: Fetch the histogram storage type in string
   * `.sum(flow=False)`: The total count of all bins
   * `.project(ax1, ax2, ...)`: Project down to listed axis (numbers). Can also reorder axes.
   * `.to_numpy(flow=False, view=False)`: Convert to a NumPy style tuple (with or without under/overflow bins)
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ histograms can be plotted via any compatible library, such as [mplhep][].
   * `*=`: Multiply by a scaler (not all storages) (`hist * scalar` and `scalar * hist` supported too)
   * `/=`: Divide by a scaler (not all storages) (`hist / scalar` supported too)
   * `.kind`: Either `bh.Kind.COUNT` or `bh.Kind.MEAN`, depending on storage
-  * `.storage_type`: Fetch the histogram storage type in string
+  * `.storage_type`: Fetch the histogram storage type
   * `.sum(flow=False)`: The total count of all bins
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -125,6 +125,7 @@ histograms can be plotted via any compatible library, such as [mplhep][].
   * `*=`: Multiply by a scaler (not all storages) (`hist * scalar` and `scalar * hist` supported too)
   * `/=`: Divide by a scaler (not all storages) (`hist / scalar` supported too)
   * `.kind`: Either `bh.Kind.COUNT` or `bh.Kind.MEAN`, depending on storage
+  * `.storageType`: Fetch the histogram storage type in string
```
henryiii, line 128:
```suggestion
  * `.storage_type`: Fetch the histogram storage type
```

src/boost_histogram/_internal/hist.py
```diff
@@ -345,6 +345,9 @@ def view(
         """
         return _to_view(self._hist.view(flow))
 
+    def storageType(self) -> str:
+        return str(self._storage_type).rsplit(".", maxsplit=1)[-1][:-2]
```
henryiii, line 349:
```suggestion
        return cast(self, self._storage_type, Storage)
```

src/boost_histogram/_internal/hist.py
```diff
@@ -345,6 +345,9 @@ def view(
         """
         return _to_view(self._hist.view(flow))
 
+    def storageType(self) -> str:
```
henryiii, line 348:
```suggestion
    def storageType(self) -> Storage:
```

src/boost_histogram/_internal/hist.py
henryiii, line 349:
```suggestion
    _storage_type = storage_type
```

And rename `_storage_type` to `storage_type` (change elsewhere too).

gohil-jay:
Now all left to do is to run this by @HDembinski !

HDembinski:
Please use the review feature, to make sure I am getting the notification. I have no objections.

src/boost_histogram/_internal/hist.py
```diff
@@ -594,7 +594,7 @@ def _axis(self, i: int = 0) -> Axis:
         return cast(self, self._hist.axis(i), Axis)
 
     @property
-    def _storage_type(self) -> Type[Storage]:
+    def storage_type(self) -> Type[Storage]:
         return cast(self, self._hist._storage_type, Storage)  # type: ignore[return-value]
```
henryiii, line 598:
```suggestion
        return cast(self, self._hist._storage_type, Storage)  # type: ignore[return-value]
        
    # Backward compat
    _storage_type = storage_type
```

henryiii:
Looks like moving to windows-2022 hits a ICE. 🤦

src/boost_histogram/_internal/hist.py
```diff
@@ -594,9 +594,12 @@ def _axis(self, i: int = 0) -> Axis:
...
 
+    # Backward compat
+    _storage_type = storage_type
+
```
HDembinski, line 602: This should raise a deprecationwarning.

Issue #734, opened 2022-04-11
[BUG] Investigate providing a better error message if sample= is missing


henryiii:
The current error message for a Mean storage type is not very helpful if you forget to pass `sample=` to fill. We probably can improve it.

We'll need a release soon, we need to ship the updated pybind11 version for PyPy 3.9 support.

PR #782, opened 2022-08-15
Better error for empty AND incorrect sample

gohil-jay:
Addresses https://github.com/scikit-hep/boost-histogram/issues/734 issue.

Added better error message for
- empty sample
- incompatible type sample
- incorrect dimension sample

### DIFF
--- a/src/boost_histogram/_internal/hist.py
+++ b/src/boost_histogram/_internal/hist.py
@@ -89,14 +89,12 @@ def meanStorageSampleCheck(sample: Optional[ArrayLike]) -> None:
     if sample is None:
         raise TypeError("Sample key-argument (sample=) needs to be provided.")
     seqs = (collections.abc.Sequence, np.ndarray)
+    msg1 = f"Sample key-argument needs to be a sequence, {sample.__class__.__name__} given."
     if isinstance(sample, str) and not isinstance(sample, seqs):
-        raise ValueError(
-            f"Sample key-argument needs to be a sequence, {sample.__class__.__name__} given."
-        )
+        raise ValueError(msg1)
+    msg2 = f"Sample key-argument needs to be 1 dimensional, {np.array(sample).ndim} given."
     if np.array(sample).ndim != 1:
-        raise ValueError(
-            f"Sample key-argument needs to be 1 dimensional, {np.array(sample).ndim} given."
-        )
+        raise ValueError(msg2)
 
 
 def _arg_shortcut(item: Union[Tuple[int, float, float], Axis, CppAxis]) -> CppAxis:
--- a/src/boost_histogram/_internal/hist.py
+++ b/src/boost_histogram/_internal/hist.py
@@ -24,7 +24,7 @@ from typing import (
 )
 
 import numpy as np
-import pytest
+
 import boost_histogram
 from boost_histogram import _core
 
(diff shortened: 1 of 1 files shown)


gohil-jay:
Solve:
- Pylint issue
- 'None' data type

src/boost_histogram/_internal/hist.py
```diff
@@ -85,6 +85,22 @@ def _fill_cast(
     return value
 
 
+def meanStorageSampleCheck(sample):
```
henryiii, line 88:
```suggestion
def meanStorageSampleCheck(sample: Optional[ArrayLike]) -> None:
```

src/boost_histogram/_internal/hist.py
```diff
@@ -85,6 +85,22 @@ def _fill_cast(
...
+    assert type(sample) != type(
+        None
+    ), "Sample key-argument (sample=) needs to be provided."
```
henryiii, line 91:
```suggestion
    assert sample is not None, "Sample key-argument (sample=) needs to be provided."
```

src/boost_histogram/_internal/hist.py
```diff
@@ -85,6 +85,22 @@ def _fill_cast(
...
+    assert isinstance(
+        sample, (collections.abc.Sequence, np.ndarray)
+    ) and not isinstance(
+        sample, (str)
+    ), f"Sample key-argument needs to be a sequence, {sample.__class__.__name__} given."
```
henryiii, line 96:
```suggestion
    msg = f"Sample key-argument needs to be a sequence, {sample.__class__.__name__} given."
    assert isinstance(
        sample, (collections.abc.Sequence, np.ndarray)
    ) and not isinstance(
        sample, (str)
    ), msg
```

src/boost_histogram/_internal/hist.py
```diff
@@ -85,6 +85,22 @@ def _fill_cast(
...
+    assert (
+        np.array(sample).ndim
+    ) == 1, (
+        f"Sample key-argument needs to be 1 dimensional, {np.array(sample).ndim} given."
+    )
```
henryiii, line 101: I think assert is probably correct here, but this would be the alternative:

```suggestion
    msg = f"Sample key-argument needs to be 1 dimensional, {np.array(sample).ndim} given."
    if np.array(sample).ndim != 1:
        raise AssertionError(msg)
```

src/boost_histogram/_internal/hist.py
```diff
@@ -489,6 +505,11 @@ def fill(
             available threads (usually two per core).
         """
 
+        if self._hist._storage_type in {
+            _core.storage.mean,
+        }:
```
henryiii, line 510:
```suggestion
        if self._hist._storage_type in {_core.storage.mean}:
```

HDembinski:
I suggest to raise a RuntimeError (or another suitable error) not AssertionError. An assert is an executable documentation of a class invariant, something that always should be true if the class is correctly used by the developers. In other words, asserts should never be triggered by users.

henryiii:
```python
def f(*, sample): pass
f()
```

The error for if `sample=` is missing should be the same type as this error.

Similarly, find a NumPy function that requires a 1D arguement and pass it a 2D argument and copy that error style. (But not one of the ones that mistakably just work incorrectly anyway for historical reasons!)

`assert` is special because it can be disabled (`python -O`), making it unevaluated and 0 cost. So you could have lots of asserts, get your code working, then disable them in production. That was the idea, anyway - it could be useful in tight loops, for example. They are also treated as always true by MyPy. These two uses are slightly at odds, but not too much, and I think the latter use is the more popular interpretation these days.

src/boost_histogram/_internal/hist.py
gohil-jay: I guess weighted_mean has different input

henryiii: What was the problem?

gohil-jay: We expect 1D array for input in mean storage. However, I saw simple numerical inputs for weighted_mean storage in examples so we can't apply the mean_storage_sample_check for weighted_mean too!

PR #786, opened 2022-08-24
Added DeprecationWarning for storage type

gohil-jay:
As _storage_type has changed to storage_type, accessing the previous one still works (backward compatibility) but raises a deprecation warning as suggested by @HDembinski [here](https://github.com/scikit-hep/boost-histogram/pull/781#discussion_r953496647).

### DIFF
--- a/src/boost_histogram/_internal/hist.py
+++ b/src/boost_histogram/_internal/hist.py
@@ -597,8 +597,9 @@ class Histogram:
     def storage_type(self) -> Type[Storage]:
         return cast(self, self._hist._storage_type, Storage)  # type: ignore[return-value]
 
-    # Backward compat
-    _storage_type = storage_type
+    def _storage_type(self) -> Type[Storage]:
+        warnings.warn("Accessing storage type has changed from _storage_type to storage_type, and will be removed in future.", DeprecationWarning)
+        return cast(self, self._hist._storage_type, Storage)  # type: ignore[return-value]
 
     def _reduce(self: H, *args: Any) -> H:
         return self._new_hist(self._hist.reduce(*args))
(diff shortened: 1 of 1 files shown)


HDembinski:
:+1:

src/boost_histogram/_internal/hist.py
```diff
@@ -597,8 +597,12 @@ def _axis(self, i: int = 0) -> Axis:
     def storage_type(self) -> Type[Storage]:
         return cast(self, self._hist._storage_type, Storage)  # type: ignore[return-value]
```
HDembinski, line 598: One more thing, unrelated this PR: shouldn't this be a property instead of a function?

henryiii: It is (you have to expand the GitHub view to see the `@property`). However, `_storage_type` also needs to be a property.

src/boost_histogram/_internal/hist.py
```diff
@@ -597,8 +597,13 @@ def _axis(self, i: int = 0) -> Axis:
...
 
-    # Backward compat
-    _storage_type = storage_type
+    @property
+    def _storage_type(self) -> Type[Storage]:
+        warnings.warn(
+            "Accessing storage type has changed from _storage_type to storage_type, and will be removed in future.",
+            DeprecationWarning,
```
henryiii, line 604:
```suggestion
            DeprecationWarning,
            stacklevel=2,
```

It is virtually never correct to call `warnings.warn` without `stacklevel` of at least 2 - this means _using_ `_storage_type` is deprecated.

henryiii (approved):

Issue #567, opened 2021-05-01
[DOCS] .reset() doesn't clear growing axis


andrzejnovak:
Unsure if intended behaviour

andrzejnovak:
I see, so it is intended. I was trying to make a copy of a histogram (while trying to make a sum on categorical axes) to assign new bins. 

I suppose it's unusual and if some groupby gets implemented unnecessary, but I'm not sure, if it's a completely illegitimate use

henryiii:
This is clearly documented as "set counters to 0", so I think it's safe to close?

henryiii:
(The name comes from Boost.Histogram, otherwise I'd suggest renaming it to `.clear()`)

andrzejnovak:
Sorry, could you point to where it's documented? I couldn't find it with a cursory search.

henryiii:
https://github.com/scikit-hep/boost-histogram#cheatsheet and https://boost-histogram.readthedocs.io/en/latest/usage/histogram.html#operations

henryiii:
Can this be closed or is there an action that should still be taken?

HDembinski:
I don't think that `.clear` would be any more obvious than `.reset`. Important is the documentation, not the name.

PR #791, opened 2022-09-17
chore: changelog for patch release v1.3.2

gohil-jay:


docs/changelog.md
```diff
@@ -12,6 +12,10 @@
 
 ### Version 1.3.2
 
+* Added `storage_type` operator and `storage_type()` function [#781][], with pending deprecation for _storage_type [#786][] [#790][]
```
henryiii, line 15:
```suggestion
* Added `storage_type` operator and `storage_type()` function [#781][], with pending deprecation for `_storage_type` [#786][] [#790][]
```

docs/changelog.md
```diff
@@ -12,6 +12,10 @@
...
+* Better errors generated for missing or incorrect sample to mean storage [#782][]
+* Explicit reset() documentation [#783][]
+* Added Python 3.11 test wheels [#789][]
```
henryiii, line 18:
```suggestion
* Added Python 3.11 wheels [#789][]
```

docs/changelog.md
```diff
@@ -12,6 +12,10 @@
 
 ### Version 1.3.2
 
+* Added `storage_type` operator and `storage_type()` function [#781][], with pending deprecation for _storage_type [#786][] [#790][]
+* Better errors generated for missing or incorrect sample to mean storage [#782][]
+* Explicit reset() documentation [#783][]
```
henryiii, line 17:
```suggestion
* Explicit `reset()` documentation [#783][]
```

henryiii: Though I'm okay to not mention docs improvements, at least specifically. It's fine though. :)

henryiii:
I need to make the needs changelog bot smarter! This literally _is_ a changelog update. I think I set it up better on pybind11 (though there it's always after the PR, but maybe we should do that too since we aren't adding these as we go very well as it is).

docs/changelog.md
gohil-jay: Yeah I thought about that too, but then I thought we can add it to the issue to close it completely. I just realized we already closed it before with a previous PR merge (https://github.com/scikit-hep/boost-histogram/issues/567)

Issue #737, opened 2022-04-25
[BUG] Cannot get index of huge numbers in IntCategory


LarsKolk:
I ran into the problem that trying to access the index of huge number in IntCategory via `.index` leads to an key-error, saying that the value does not exist in this axis. 
From what I have tested, this seems to happen for numbers greater than `99_999_000`. The following example works as expected in version `1.2.1`, but fails in version `1.3.1`.
**Steps to reproduce**
```python
import boost_histogram as bh
h = bh.Histogram(bh.axis.IntCategory([4, 8, 15, 16, 23, 42, 99_999_001, 1_000_010_020]))
print(f"value = {h.axes[0].value(6)}")
print(f"index = {h.axes[0].index(99_999_001)}")
```
Expected output: 
```
value = 9999001
index = 6
```
Actual output: 
```python
value = 99999001

---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Input In [29], in <cell line: 2>()
      1 print(f"value = {h.axes[0].value(6)}")
----> 2 print(f"index = {h.axes[0].index(99_999_001)}")

File ~/.conda/envs/Test/lib/python3.8/site-packages/boost_histogram/_internal/axis.py:118, in Axis.index(self, value)
    115     msg = f"index({value}) cannot be a string for a numerical axis"
    116     raise TypeError(msg)
--> 118 return self._ax.index(value)

KeyError: '99999001 not in axis'
```

henryiii:
Very strange! I’ll try to look at it soon (possibly after PyCon).

PR #792, opened 2022-09-17
tests: add test for large num

henryiii:
As described in https://github.com/scikit-hep/boost-histogram/issues/737. Starting by adding a test.

### DIFF
--- a/include/bh_python/register_axis.hpp
+++ b/include/bh_python/register_axis.hpp
@@ -58,10 +58,10 @@ inline decltype(auto) axis_cast<int>(py::handle x) {
     if(py::isinstance<int>(x))
         return py::cast<int>(x);
 
-    auto val  = py::cast<float>(x);
+    auto val  = py::cast<double>(x);
     auto ival = static_cast<int>(val);
 
-    if(static_cast<float>(ival) == val)
+    if(static_cast<double>(ival) == val)
         return ival;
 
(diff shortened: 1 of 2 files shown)


henryiii:
The 32-bit floating point limit is exactly where this starts happening:

```python
>>> h = bh.Histogram(bh.axis.IntCategory([4, 8, 15, 16, 23, 42, x := 16_777_217, 1_000_010_020])); h.axes[0].index(x)
  File "<stdin>", line 1, in <module>
  File "/Users/henryschreiner/git/scikit-hep/boost-histogram/.nox/tests-3-10/lib/python3.10/site-packages/boost_histogram/_internal/axis.py", line 118, in index
    return self._ax.index(value)  # type: ignore[no-any-return]
KeyError: '16777217 not in axis'
>>> h = bh.Histogram(bh.axis.IntCategory([4, 8, 15, 16, 23, 42, x := 16_777_216, 1_000_010_020])); h.axes[0].index(x)
6
```

So where is this getting cast to a 32-bit float?

henryiii:
(That wasn't the fix, but is a small improvement. Now there are no `float` mentions in all the cpp code).

henryiii:
Also, this is working correctly in C++:

https://wandbox.org/permlink/ocx7lQe3w8DLaoYB

henryiii:
Ohh, maybe that float was the problem! Didn't fix it locally, but I might not have recompiled.

PR #798, opened 2022-09-21
chore: drop 3.6

henryiii:
Removing 3.6.


henryiii:
Not seeing anything bad happening after 1.3.2, so let's ho ahead and merge this as the start of version 1.4. If we have to backport and make a 1.3.3 due to some horrible missed bug that we really need fixed on Python 3.6, then branching is cheap. :)

PR #802, opened 2022-10-03
chore: use boost 1.80+

henryiii:
Will update numpy axis if this works well.


henryiii:
Will need to solve unpickling old files with the old axis. Also division support for histograms can be added.

PR #836, opened 2023-03-13
docs: update readthedocs config

henryiii:
Docs are broken, not sure why, checking ~~myst~~ nbsphinx.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,6 +5,11 @@
 # Required
 version: 2
 
+build:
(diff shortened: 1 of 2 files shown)


mgeier:
A quick solution would be to update your RTD settings, see https://github.com/spatialaudio/nbsphinx/issues/720#issuecomment-1467799866.

henryiii:
Thanks! That's the best solution. I might update https://scikit-hep.org/developer/reporeview to check for rtd and this newer build config option.

PR #864, opened 2023-07-03
chore: ruff moved to astral-sh

henryiii:
See https://github.com/scientific-python/cookie/pull/200.

Committed via https://github.com/asottile/all-repos

### DIFF
--- a/.github/workflows/emscripten.yaml
+++ b/.github/workflows/emscripten.yaml
@@ -29,7 +29,7 @@ jobs:
         python-version: "3.11"
 
     - name: Install pyodide-build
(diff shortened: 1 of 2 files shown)


.github/workflows/emscripten.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
         python-version: "3.11"
 
     - name: Install pyodide-build
-      run: pip install pyodide-build==0.23.1
+      run: pip install pyodide-build==0.23.1 pydantic<2
```
henryiii, line 32:
```suggestion
      run: pip install pyodide-build==0.23.1 "pydantic<2"
```

PR #887, opened 2023-09-15
feat: move to scikit-build-core

henryiii:
Moving the build system to scikit-build-core.


.github/workflows/emscripten.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
           python-version: "3.11"
 
       - name: Install pyodide-build
-        run: pip install pyodide-build==0.23.4
+        run: pip install pyodide-build==0.24.1
```
henryiii, line 32:
```suggestion
        run: pip install pyodide-build==0.23.4
```

0.24.0 and 0.24.1 are broken due to needing https://github.com/pyodide/pyodide/pull/4223 which isn't in a release yet.

PR #895, opened 2023-10-26
chore: move to Ruff formatter

henryiii:
See https://github.com/scientific-python/cookie/pull/299


henryiii:
Waiting on https://github.com/astral-sh/ruff/issues/8254.

PR #909, opened 2024-01-17
Add missing API docs for `bh.Histogram`

Saransh-cpp:
For some reason, the API doc of `Histogram` is missing from the documentation website. `Histogram` is in a private module, but it is exported in `__init__.py`. Furthermore, the module `boost_histogram` itself has no docstrings to display.

This was briefly mentioned in https://github.com/scikit-hep/hist/issues/445.

### DIFF
--- a/docs/api/boost_histogram.rst
+++ b/docs/api/boost_histogram.rst
@@ -1,8 +1,8 @@
 boost\_histogram
 ================
 
-.. automodule:: boost_histogram
-   :members:
+.. automodule:: boost_histogram._internal.hist
+   :members: Histogram
    :undoc-members:
    :show-inheritance:
 


henryiii:
Does this get removed if you rerun `nox -s build_api_docs`?

Saransh-cpp:
Oh yes, it reverts the changes. Looking into it.

Saransh-cpp:
Adding `--private` brings in every private module to the docs. We could only include `hist.py` by excluding other private files, but I can't see a way to exclude private members of `hist.py` after using `--private`.

The templates could be changed but the documentation for it is almost non-existent. The only available examples are in [the Sphinx project](https://github.com/sphinx-doc/sphinx/tree/master/sphinx/templates/apidoc), which I think were taken as it is for boost-histogram.

henryiii:
It's modified slightly, but yes, it was.

We could hack it by reading the file and then writing it after the generation step. It would still work in nox but not by hand, which IMO is fine for now.

Ideally, I'd like to get rid of the private files and instead move everything to where it is supposed to be, then use `__all__` and module level `__dir__` to hide the private parts. That wasn't supported everywhere when boost-histogram was written (in fact, 3.7 was probably pretty new). But that's for later.

henryiii:
Thanks!

PR #913, opened 2024-01-25
feat: support variable rebinning

Saransh-cpp:
XRef #208 

The current interface:
```python
In [1]: import numpy as np
   ...: 
   ...: import boost_histogram as bh
   ...: 
   ...: h = bh.Histogram(bh.axis.Regular(10, 0, 1))
   ...: h.fill(np.random.normal(size=1_000_000))
   ...: rebin = bh.tag.Rebinner(factor=2)
   ...: h[::rebin]
Out[1]: Histogram(Regular(5, 0, 1), storage=Double()) # Sum: 341605.0 (1000000.0 with flow)

In [2]: rebin = bh.tag.Rebinner(groups=[1, 2, 3])

In [3]: h[::rebin]
Out[3]: Histogram(Variable([0, 0.1, 0.3, 0.6], metadata=...), storage=Double()) # Sum: 225749.0

In [4]: s = bh.tag.Slicer()
   ...: 
   ...: h = bh.Histogram(
   ...:     bh.axis.Regular(20, 1, 3), bh.axis.Regular(30, 1, 3),
   ...: bh.axis.Regular(40, 1, 3)
   ...: )
   ...: 
   ...: h[{0: s[:: bh.rebin(groups=[1, 2, 3])]}].axes.size
Out[4]: (3, 30, 40)

In [5]: h[{0: s[:: bh.rebin(groups=[1, 2, 3])], 2: s[:: bh.rebin(g
   ...: roups=[1, 2 ,3])]}].axes[2].edges
Out[5]: array([1.  , 1.05, 1.15, 1.3 ])
```

- [ ] The code is a bit dirty and I don't know if it is perfectly optimized.
- [ ] How should the code handle flow bins?
- [ ] Is there any edge case that I am missing?

cc: @henryiii @matthewfeickert

### DIFF
--- a/src/boost_histogram/__init__.py
+++ b/src/boost_histogram/__init__.py
@@ -3,6 +3,9 @@ from __future__ import annotations
 from . import accumulators, axis, numpy, storage
 from ._internal.enum import Kind
 from ._internal.hist import Histogram, IndexingExpr
+from .tag import (
+    Rebinner as rebin,
+)
 from .tag import (  # pylint: disable=redefined-builtin
     loc,
     overflow,
@@ -36,6 +39,7 @@ __all__ = (
     "accumulators",
     "numpy",
     "loc",
+    "rebin",
     "sum",
     "underflow",
     "overflow",
--- a/src/boost_histogram/__init__.py
+++ b/src/boost_histogram/__init__.py
@@ -6,7 +6,6 @@ from ._internal.hist import Histogram, IndexingExpr
 from .tag import (  # pylint: disable=redefined-builtin
     loc,
     overflow,
-    rebin,
     sum,
     underflow,
 )
--- a/src/boost_histogram/__init__.py
+++ b/src/boost_histogram/__init__.py
@@ -3,12 +3,10 @@ from __future__ import annotations
 from . import accumulators, axis, numpy, storage
 from ._internal.enum import Kind
 from ._internal.hist import Histogram, IndexingExpr
-from .tag import (
-    Rebinner as rebin,
-)
 from .tag import (  # pylint: disable=redefined-builtin
     loc,
     overflow,
+    rebin,
     sum,
     underflow,
 )
--- a/src/boost_histogram/__init__.py
+++ b/src/boost_histogram/__init__.py
@@ -36,7 +36,6 @@ __all__ = (
     "accumulators",
     "numpy",
     "loc",
-    "rebin",
     "sum",
     "underflow",
     "overflow",
--- a/src/boost_histogram/_internal/hist.py
+++ b/src/boost_histogram/_internal/hist.py
@@ -827,6 +827,7 @@ class Histogram:
         slices: list[_core.algorithm.reduce_command] = []
         pick_each: dict[int, int] = {}
         pick_set: dict[int, list[int]] = {}
+        reduced: CppHistogram | None = None
 
         # Compute needed slices and projections
         for i, ind in enumerate(indexes):
@@ -891,7 +892,8 @@ class Histogram:
                     )
                 # rebinning with groups
                 elif len(groups) != 0:
-                    reduced = self._hist
+                    if not reduced:
+                        reduced = self._hist
                     axes = [reduced.axis(x) for x in range(reduced.rank())]
                     reduced_view = reduced.view(flow=True)
                     new_axes_indices = [axes[i].edges[0]]
@@ -914,7 +916,7 @@ class Histogram:
                     j = 1
                     for new_j, group in enumerate(groups):
                         for _ in range(group):
-                            pos = [slice] * (i)
+                            pos = [slice(None)] * (i)
                             new_view[(*pos, new_j + 1, ...)] += reduced_view[  # type: ignore[arg-type]
                                 (*pos, j, ...)  # type: ignore[arg-type]
                             ]
@@ -923,10 +925,9 @@ class Histogram:
                     reduced = new_reduced
 
         # Will be updated below
-        if slices or pick_set or pick_each or integrations:
+        if (slices or pick_set or pick_each or integrations) and not reduced:
             reduced = self._hist
-        elif len(groups) == 0:
-            logger.debug("Reduce actions are all empty, just making a copy")
+        elif not reduced:
             reduced = copy.copy(self._hist)
 
         if pick_each:
--- a/src/boost_histogram/_internal/hist.py
+++ b/src/boost_histogram/_internal/hist.py
@@ -858,12 +858,12 @@ class Histogram:
             if ind != slice(None):
                 merge = 1
                 if ind.step is not None:
-                    if ind.step.factor is not None:
+                    if getattr(ind.step, "factor", None) is not None:
                         merge = ind.step.factor
                     elif callable(ind.step):
                         if ind.step is sum:
                             integrations.add(i)
-                        elif ind.step.groups is not None:
+                        elif getattr(ind.step, "groups", None) is not None:
                             groups = ind.step.groups
                         else:
                             raise NotImplementedError
@@ -908,9 +908,8 @@ class Histogram:
                 new_axes_indices += [axes[i].edges[j + 1 : j + group + 1][-1]]
                 j = group
 
-            variable_axis = Variable(new_axes_indices)
-            variable_axis.metadata = axes[i].metadata
-            axes[i] = variable_axis
+            variable_axis = Variable(new_axes_indices, metadata=axes[i].metadata)
+            axes[i] = variable_axis._axis
             reduced_view = np.take(reduced_view, range(len(reduced_view)), axis=i)
 
             logger.debug("Axes: %s", axes)
--- a/src/boost_histogram/_internal/hist.py
+++ b/src/boost_histogram/_internal/hist.py
@@ -862,11 +862,11 @@ class Histogram:
                 if ind.step is not None:
                     if getattr(ind.step, "factor", None) is not None:
                         merge = ind.step.factor
+                    elif getattr(ind.step, "group_mapping", None) is not None:
+                        groups = ind.step.group_mapping(self.axes[i])
                     elif callable(ind.step):
                         if ind.step is sum:
                             integrations.add(i)
-                        elif getattr(ind.step, "groups", None) is not None:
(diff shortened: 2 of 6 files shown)


src/boost_histogram/tag.py
```diff
@@ -107,13 +110,49 @@ def __call__(self, axis: AxisLike) -> int:  # noqa: ARG002
...
+        }
+        for k, v in args.items():
+            if v is not None:
+                return_str = f"{repr_str}({k}={v})"
+                break
+        return return_str
+
+    def __call__(
+        self, axis: PlottableAxis
+    ) -> int | Sequence[int] | Mapping[int | str, Sequence[int | str]]:
```
Saransh-cpp, line 148:
```suggestion
    ) -> int | Sequence[int]:
```

src/boost_histogram/tag.py
```diff
@@ -4,11 +4,14 @@
 
 import copy
 from builtins import sum
-from typing import TypeVar
+from typing import Mapping, Sequence, TypeVar
```
Saransh-cpp, line 7:
```suggestion
from typing import Sequence, TypeVar
```

henryiii:
Ah, yeah, you probably have to use boost-histogram's cast system to go from C++ class to the correct Python class. I can look (hopefully by end of day or tomorrow, as I'll be teaching soon).

src/boost_histogram/_internal/hist.py
```diff
@@ -854,13 +854,15 @@ def __getitem__(self: H, index: IndexingExpr) -> H | float | Accumulator:
             if ind != slice(None):
                 merge = 1
                 if ind.step is not None:
-                    if hasattr(ind.step, "factor"):
+                    if getattr(ind.step, "factor", None) is not None:
                         merge = ind.step.factor
                     elif callable(ind.step):
                         if ind.step is sum:
                             integrations.add(i)
+                        elif getattr(ind.step, "groups", None) is not None:
```
henryiii, line 862: This should call `ind.step`, not check for the `groups` attr. The rebinner needs to be able to compute this on the fly, with the axis as an input.

src/boost_histogram/_internal/hist.py
```diff
@@ -854,13 +854,15 @@ def __getitem__(self: H, index: IndexingExpr) -> H | float | Accumulator:
...
+                            groups = ind.step.groups
```
henryiii, line 863: This needs to be used later. Probably where "merge" is used, this should also be processed?

src/boost_histogram/_internal/hist.py
```diff
@@ -885,6 +890,30 @@ def __getitem__(self: H, index: IndexingExpr) -> H | float | Accumulator:
             logger.debug("Reduce actions are all empty, just making a copy")
             reduced = copy.copy(self._hist)
 
+        # bin re-grouping
```
henryiii, line 893: I _think_ this should be moved to where "merge" is processed? I think it's safe to do this in the loop as long as it doesn't add or remove axes (which it doesn't, only possibly swaps an axis)

src/boost_histogram/__init__.py
```diff
@@ -3,10 +3,12 @@
 from . import accumulators, axis, numpy, storage
 from ._internal.enum import Kind
 from ._internal.hist import Histogram, IndexingExpr
+from .tag import (
+    Rebinner as rebin,
```
henryiii, line 7: We need to keep backward compat in API.

src/boost_histogram/_internal/hist.py
```diff
@@ -885,6 +890,30 @@ def __getitem__(self: H, index: IndexingExpr) -> H | float | Accumulator:
...
+            axes = [reduced.axis(i) for i in range(reduced.rank())]
+            reduced_view = reduced.view(flow=True)
+            new_axes_indices = [axes[i].edges[0]]
+            j: int = 0
+            for group in groups:
+                new_axes_indices += [axes[i].edges[j + 1 : j + group + 1][-1]]
+                j = group
+
+            variable_axis = Variable(new_axes_indices, metadata=axes[i].metadata)
+            axes[i] = variable_axis._axis
```
henryiii, line 908: This is a quick and easy way to get the C++ axes, just make a Python axis and pull the `._axis` out of it.

src/boost_histogram/tag.py
```diff
@@ -107,13 +110,40 @@ def __call__(self, axis: AxisLike) -> int:  # noqa: ARG002
...
-    def __init__(self, value: int) -> None:
-        self.factor = value
+    def __init__(
+        self,
+        factor: int | None = None,
+        *,
+        groups: Sequence[int] | None = None,
+    ) -> None:
+        if not sum(i is None for i in [factor, groups]) == 1:
+            raise ValueError("exactly one, a value or groups should be provided")
```
Saransh-cpp, line 126:
```suggestion
            raise ValueError("Exactly one, a factor or groups should be provided")
```

src/boost_histogram/_internal/hist.py
```diff
@@ -851,37 +851,72 @@ def __getitem__(self: H, index: IndexingExpr) -> H | float | Accumulator:
...
+                        reduced_view, range(len(reduced_view)), axis=i
+                    )
+
+                    logger.debug("Axes: %s", axes)
+
+                    new_reduced = reduced.__class__(axes)
+                    new_reduced.fill(
+                        np.add.reduceat(reduced.to_numpy()[0], new_axes_indices, axis=i)
+                    )
+                    reduced = new_reduced
```
Saransh-cpp, line 914: I am doing something wrong here; I can't figure out how to distribute the bin contents once bins are regrouped. Could someone please help me with this? Thanks!

Current status:
```py
In [1]: import boost_histogram as bh
   ...: import numpy as np
   ...: h = bh.Histogram(bh.axis.Regular(10, 0, 1))
   ...: h.fill(np.random.normal(size=1_000_000))
   ...: rebin = bh.tag.Rebinner(factor=2)
   ...: h[::rebin]  # rebin with a factor
Out[1]: Histogram(Regular(5, 0, 1), storage=Double()) # Sum: 342342.0 (1000000.0 with flow)

In [2]: rebin = bh.tag.Rebinner(groups=[1, 2, 3])

In [3]: h[::rebin] # rebin into groups (variable axis)
Out[3]: Histogram(Variable([0, 0.1, 0.3, 0.6], metadata=...), storage=Double()) # Sum: 0.0 (4.0 with flow)

In [4]: print(h[::rebin])  # doing something wrong while distributing the values
              ┌──────────────────────────────────────────────────────────────┐
[-inf,   0) 0 │                                                              │
[   0, 0.1) 0 │                                                              │
[ 0.1, 0.3) 0 │                                                              │
[ 0.3, 0.6) 0 │                                                              │
[ 0.6, inf) 4 │█████████████████████████████████████████████████████████████ │
              └──────────────────────────────────────────────────────────────┘

In [5]: rebin = bh.tag.Rebinner(groups=[1 for x in range(10)]) # rebin into same bins but introduce a variable axis

In [6]: h[::rebin] # rebin into groups (variable axis)
Out[6]: Histogram(Variable([0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1], metadata=...), storage=Double()) # Sum: 0.0 (11.0 with flow)

In [7]: print(h[::rebin])  # doing something wrong while distributing the values
               ┌─────────────────────────────────────────────────────────────┐
[-inf,   0) 0  │                                                             │
[   0, 0.1) 0  │                                                             │
[ 0.1, 0.2) 0  │                                                             │
[ 0.2, 0.3) 0  │                                                             │
[ 0.3, 0.4) 0  │                                                             │
[ 0.4, 0.5) 0  │                                                             │
[ 0.5, 0.6) 0  │                                                             │
[ 0.6, 0.7) 0  │                                                             │
[ 0.7, 0.8) 0  │                                                             │
[ 0.8, 0.9) 0  │                                                             │
[ 0.9,   1) 0  │                                                             │
[   1, inf) 11 │████████████████████████████████████████████████████████████ │
               └─────────────────────────────────────────────────────────────┘
```

cc: @matthewfeickert @andrzejnovak @alexander-held

andrzejnovak: Just birdseye and I only usually interact with `bh` with `hist`, but on L911 you're calling `fill`? I'd expect there be some value assignment rather than filling. As the yields in your demo hist are >1, would be consistent with them appearing in overflow on an 0,1 axis.

Saransh-cpp: Yes, maybe I am missing an internal function that can be used here?

rkansal47:
Thanks for this very useful feature! I was wondering if this adds (or could add) support for renaming categorical axis values as well?

henryiii:
I still need to review this and make it work on callables.

src/boost_histogram/_internal/hist.py
```diff
@@ -854,38 +855,78 @@ def __getitem__(self: H, index: IndexingExpr) -> H | float | Accumulator:
...
 
+            groups = []
             if ind != slice(None):
                 merge = 1
                 if ind.step is not None:
-                    if hasattr(ind.step, "factor"):
+                    if getattr(ind.step, "factor", None) is not None:
                         merge = ind.step.factor
+                    elif getattr(ind.step, "group_mapping", None) is not None:
+                        groups = ind.step.group_mapping(self.axes[i])
```
henryiii, line 865: TODO: I didn't update this correctly.

src/boost_histogram/tag.py
```diff
@@ -4,11 +4,14 @@
...
-from typing import TypeVar
+from typing import TYPE_CHECKING, Sequence, TypeVar
+
+if TYPE_CHECKING:
+    from uhi.typing.plottable import PlottableAxis
 
 from ._internal.typing import AxisLike
 
-__all__ = ("Slicer", "Locator", "at", "loc", "overflow", "underflow", "rebin", "sum")
+__all__ = ("Slicer", "Locator", "at", "loc", "overflow", "underflow", "sum", "rebin")
```
henryiii, line 14:
```suggestion
__all__ = ("Slicer", "Locator", "at", "loc", "overflow", "underflow", "rebin", "sum")
```

PR #917, opened 2024-03-26
fix: allow filling Int axis with unsigned integers

veprbl:


henryiii:
Could you drop a quick test in to be sure? I think it’s fine but good to be sore.

henryiii (approved):

henryiii:
Thanks!

PR #938, opened 2024-05-28
ci: bump to Pyodide 3.12

henryiii:
I expect this to fail. (Okay, not _that_ early!)

### DIFF
--- a/.github/workflows/emscripten.yaml
+++ b/.github/workflows/emscripten.yaml
@@ -31,7 +31,7 @@ jobs:
       - name: Install pyodide-build
         run: |
           pip install pyodide-build==0.26.1
-          echo EMSCRIPTEN_VERSION=$(pyodide config get emscripten_version >> $GITHUB_ENV
+          echo EMSCRIPTEN_VERSION=$(pyodide config get emscripten_version) >> $GITHUB_ENV
 
       - uses: mymindstorm/setup-emsdk@v14
         with:
--- a/.github/workflows/emscripten.yaml
+++ b/.github/workflows/emscripten.yaml
@@ -26,11 +26,11 @@ jobs:
 
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
+          python-version: "3.12"
 
       - name: Install pyodide-build
         run: |
-          pip install pyodide-build==0.25.1
+          pip install pyodide-build==0.26.0
           echo EMSCRIPTEN_VERSION=$(pyodide config get emscripten_version) >> $GITHUB_ENV
 
       - uses: mymindstorm/setup-emsdk@v14
--- a/.github/workflows/emscripten.yaml
+++ b/.github/workflows/emscripten.yaml
@@ -31,7 +31,7 @@ jobs:
       - name: Install pyodide-build
         run: |
           pip install pyodide-build==0.26.0
-          echo EMSCRIPTEN_VERSION=$(pyodide config get emscripten_version) >> $GITHUB_ENV
+          echo EMSCRIPTEN_VERSION=$(pyodide config get emscripten_version | tail -n 1 | awk '{print $NF}') >> $GITHUB_ENV
 
       - uses: mymindstorm/setup-emsdk@v14
         with:
--- a/.github/workflows/emscripten.yaml
+++ b/.github/workflows/emscripten.yaml
@@ -39,7 +39,7 @@ jobs:
           actions-cache-folder: emsdk-cache
 
       - name: Build
-        run: CFLAGS=-fexceptions LDFLAGS=-fexceptions pyodide build --exports whole_archive
+        run: pyodide build --exports whole_archive
 
       - uses: actions/setup-node@v4
         with:
--- a/.github/workflows/emscripten.yaml
+++ b/.github/workflows/emscripten.yaml
@@ -39,7 +39,7 @@ jobs:
           actions-cache-folder: emsdk-cache
 
       - name: Build
-        run: pyodide build --exports whole_archive
+        run: CXXFLAGS=-fexceptions LDFLAGS=-fexceptions pyodide build --exports whole_archive
 
       - uses: actions/setup-node@v4
         with:
--- a/.github/workflows/emscripten.yaml
+++ b/.github/workflows/emscripten.yaml
@@ -30,8 +30,8 @@ jobs:
 
       - name: Install pyodide-build
         run: |
-          pip install pyodide-build==0.26.0
-          echo EMSCRIPTEN_VERSION=$(pyodide config get emscripten_version | tail -n 1 | awk '{print $NF}') >> $GITHUB_ENV
+          pip install pyodide-build==0.26.1
+          echo EMSCRIPTEN_VERSION=$(pyodide config get emscripten_version >> $GITHUB_ENV
 
       - uses: mymindstorm/setup-emsdk@v14
         with:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -117,11 +117,6 @@ sdist.exclude = [
 write_to = "src/boost_histogram/version.py"
 
 
-[tool.pyodide.build]
-cxxflags = "-fexceptions"
-ldflags = "-fexceptions"
-
-
 [tool.pytest.ini_options]
 minversion = "6.0"
 junit_family = "xunit2"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -117,6 +117,11 @@ sdist.exclude = [
 write_to = "src/boost_histogram/version.py"
 
 
+[tool.pyodide.build]
+cflags = ["-fexceptions"]
+ldflags = ["-fexceptions"]
+
+
 [tool.pytest.ini_options]
 minversion = "6.0"
 junit_family = "xunit2"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -118,7 +118,7 @@ write_to = "src/boost_histogram/version.py"
 
 
 [tool.pyodide.build]
-cflags = "-fexceptions"
+cxxflags = "-fexceptions"
 ldflags = "-fexceptions"
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -118,8 +118,8 @@ write_to = "src/boost_histogram/version.py"
 
 
 [tool.pyodide.build]
-cflags = ["-fexceptions"]
-ldflags = ["-fexceptions"]
+cflags = "-fexceptions"
+ldflags = "-fexceptions"
 
 
 [tool.pytest.ini_options]


hoodmane:
Improvement:
```
Pyodide has suffered a fatal error. Please report this to the Pyodide maintainers.
The cause of the fatal error was:
CppException pybind11::stop_iteration: The exception is an object of type pybind11::stop_iteration at address 65499896 which does not inherit from std::exception
    at convertCppException (/home/rchatham/Documents/programming/boost-histogram/.pyodide-xbuildenv-0.26.0/0.26.0/xbuildenv/pyodide-root/dist/pyodide.asm.js:10:49457)
    at API.fatal_error (/home/rchatham/Documents/programming/boost-histogram/.pyodide-xbuildenv-0.26.0/0.26.0/xbuildenv/pyodide-root/dist/pyodide.asm.js:10:49750)
    at main (/home/rchatham/Documents/programming/boost-histogram/.pyodide-xbuildenv-0.26.0/0.26.0/xbuildenv/pyodide-root/dist/python:180:17) {
  ty: 'pybind11::stop_iteration',
  pyodide_fatal_error: true
}
```

hoodmane:
Oh well the problem I'm having is that I am building without `-fexceptions`. But also it was not correctly handling the fatal.

henryiii:
It does inherit from `std::runtime_error`: https://github.com/pybind/pybind11/blob/1a0ff405498b6aac4b57cf0d95670010e5e37973/include/pybind11/detail/common.h#L996C60-L1011

hoodmane:
But it sure looks like `*message` can only be `NULL` if 
```cpp
const __shim_type_info* catch_type =
    static_cast<const __shim_type_info*>(&typeid(std::exception));
int can_catch = catch_type->can_catch(thrown_type, thrown_object);
``` 
returns `False`. This seems to be asking if it would be caught by a `catch(std::exception x)`... Maybe the trouble is that `can_catch` misbehaves if the code that threw the exception was compiled without `-fexceptions`?

hoodmane:
Anyways now that I am building with `-fexceptions` I reproduce the error we see in CI:
```
Pyodide has suffered a fatal error. Please report this to the Pyodide maintainers.
The cause of the fatal error was:
TypeError: getWasmTableEntry(...) is not a function
    at invoke_viii (/home/rchatham/Documents/programming/boost-histogram/.pyodide-xbuildenv-0.26.0/0.26.0/xbuildenv/pyodide-root/dist/pyodide.asm.js:10:1225094)
    at wasm://wasm/00de1546:wasm-function[1873]:0x1e3b2b
    at invoke_vii (/home/rchatham/Documents/programming/boost-histogram/.pyodide-xbuildenv-0.26.0/0.26.0/xbuildenv/pyodide-root/dist/pyodide.asm.js:10:1224639)
    at wasm://wasm/00de1546:wasm-function[1872]:0x1e3792
    at invoke_ii (/home/rchatham/Documents/programming/boost-histogram/.pyodide-xbuildenv-0.26.0/0.26.0/xbuildenv/pyodide-root/dist/pyodide.asm.js:10:1223536)
    at wasm://wasm/00de1546:wasm-function[803]:0x14b50a
    at invoke_ii (/home/rchatham/Documents/programming/boost-histogram/.pyodide-xbuildenv-0.26.0/0.26.0/xbuildenv/pyodide-root/dist/pyodide.asm.js:10:1223536)
    at wasm://wasm/00de1546:wasm-function[298]:0xf638e
    at _PyEM_TrampolineCall_JS (/home/rchatham/Documents/programming/boost-histogram/.pyodide-xbuildenv-0.26.0/0.26.0/xbuildenv/pyodide-root/dist/pyodide.asm.js:10:125866)
    at wasm://wasm/0267d6e2:wasm-function[2178]:0x20af20 {
  pyodide_fatal_error: true
}
```

hoodmane:
Seems that `invoke_viii` is being called with a null pointer for some reason.

hoodmane:
So the problem is that `global.get $GOT.func.__cxa_throw` is getting a null pointer...

hoodmane:
If I fill in `GOT.__cxa_throw` manually the rest of the test suite works:
```js
const func = new WebAssembly.Function({parameters: ["i32", "i32", "i32"], results:[]}, Module.___cxa_throw);
const slot = Module.wasmTable.grow(1);
Module.wasmTable.set(slot, func);
GOT.__cxa_throw.value = slot;
```
@sbc100

hoodmane:
Actually, just calling `Module.reportUndefinedSymbols()` fixes the problem.

hoodmane:
```patch
--- a/src/library_dylink.js
+++ b/src/library_dylink.js
@@ -1143,7 +1143,9 @@ var LibraryDylink = {
     }
 
     try {
-      return loadDynamicLibrary(filename, combinedFlags, localScope, handle)
+      const result = loadDynamicLibrary(filename, combinedFlags, localScope, handle);
+      reportUndefinedSymbols();
+      return result;
     } catch (e) {
 #if ASSERTIONS
       err(`Error in loading dynamic library ${filename}: ${e}`);
```

lesteve:
Great to hear if there is a fix for this! I was wondering indeed whether this was not going to be an issue to use cibuildwheel for scikit-learn. I guess this is the issue you have in mind: https://github.com/pyodide/pyodide/issues/3865.

hoodmane:
Hmm darn that sounds different. Here symbol resolution succeeded but the code to lazily initialize the GOT failed. There it looks like it failed at symbol resolution. So there's several bugs it seems. 

We're using this same mechanism for loading shared libs at cloudflare (it's necessary for performance) so there's a lot of reason to fix it...

hoodmane:
Minimal reproduction here is:
```py
def test_metadata():
    a1 = bh.axis.Regular(1, 2, 3, underflow=False)
    a2 = bh.axis.Regular(1, 2, 3)
    a1 == a2 # fails
```
Now I guess I'll have to start minimizing the C++ code =/
Presumably I can modify pybind11 in place too?

henryiii:
You can copy it locally and swap out the find_package for add_subdirectory.

henryiii:
The repl is pretty broken on iOS in 0.26 otherwise I’d check it with the in-tree one. Can do later on a computer.

hoodmane:
I think I'm making pretty good progress on minimizing. So far have the C++ code in src down to:
```cpp
#include <bh_python/pybind11.hpp>

#include <bh_python/axis.hpp>
#include <bh_python/register_axis.hpp>

void register_axes(py::module& mod) {
    auto ax = register_axis<axis::regular_none>(mod);
    ax.def(py::init<unsigned, double, double>(), "bins"_a, "start"_a, "stop"_a);
    ax = register_axis<axis::regular_uflow>(mod);
    ax.def(py::init<unsigned, double, double>(), "bins"_a, "start"_a, "stop"_a);
}
```
trying to reduce `register_axis.hpp` right now.

hoodmane:
`axis.hpp` should directly include `#include <boost/histogram/axis/regular.hpp>`, it picks it up accidentally from `regular_numpy.hpp`

hoodmane:
Currently I have `module.cpp`:
```c++
#include <pybind11/pybind11.h>
#include <boost/histogram/axis/regular.hpp>
#include <boost/histogram/axis.hpp>

using namespace pybind11::literals; // For ""_a syntax
namespace py = pybind11;
namespace bh = boost::histogram;

using regular_none
    = bh::axis::regular<double, bh::use_default, int, bh::axis::option::none_t>;
using regular_uflow
    = bh::axis::regular<double, bh::use_default, double, bh::axis::option::none_t>;

template <class A>
py::class_<A> register_axis(py::module& m, const char* name) {
    py::class_<A> ax(m, name);

    ax.def("f",
        [](const A& self, const py::object& other) {
            return self == py::cast<A>(other);
    });
    return ax;
}


PYBIND11_MODULE(_core, m) {
    py::module mod = m.def_submodule("axis");
    mod.def("do_the_thing", [] (const py::object& a) {
        py::cast<regular_none>(a);
        return 77;
    });
    auto ax = register_axis<regular_uflow>(mod, "regular_uflow");
    ax.def(py::init<unsigned, double, double>(), "bins"_a, "start"_a, "stop"_a);
}
```
and `repro.py`:
```py
from boost_histogram._core import axis as ca
ca.do_the_thing(ca.regular_uflow(1, 2, 3))
```
Without the fix, it is a fatal error. With the fix, it raises:
```C++
RuntimeError: Unable to cast Python instance of type <class 'boost_histogram._core.axis.regular_uflow'> to C++ type 'boost::histogram::axis::regular<double, boost::use_default, int, boost::histogram::axis::option::bitset<0u>>'
```

hoodmane:
@henryiii looks like at this point I should be ready to remove boost histogram entirely. I just need suitable replacements for `regular_none` and `regular_uflow`...

hoodmane:
Okay, no `boost/histogram` anymore:
```cpp
#include <pybind11/pybind11.h>

template <class Value>
class regular {};

namespace py = pybind11;


PYBIND11_MODULE(_core, m) {
    py::module mod = m.def_submodule("axis");
    mod.def("do_the_thing", [] (const py::object& a) {
        py::cast<regular<double>>(a);
        return 77;
    });
    py::class_<regular<int>> ax(mod, "Thing");
    ax.def(py::init<>());
}
```
and:
```py
from boost_histogram._core import axis as ca
print(ca.do_the_thing(ca.Thing()))
```

hoodmane:
If I can't reduce this any further, this already wouldn't be an awful test case.

hoodmane:
Okay got the following reproducer:
```C++
#include <stdexcept>
#include "Python.h"

extern "C" __attribute__((visibility("default"))) 
PyObject *PyInit__core() { 
    try { 
        throw std::runtime_error("something bad?");
    } catch (const std::exception &e) { 
        PyErr_SetString(PyExc_ImportError, "oops");
    } 
    return nullptr; 
} 
```
I think exceptions just don't work at all...

hoodmane:
Actually, if you delete `__attribute__((visibility("default")))` the problem goes away.

henryiii:
I've been using `--exports whole_archive`, are you using that? The default pyodide-build behavior IIRC picks out just the PyInit. Not sure why anything would be affected here, as default visibility shouldn't hide other symbols by itself?

henryiii:
FYI, this works fine with the in-tree build of boost-histogram. That is based on the old setuptools build system instead of CMake, but assuming this shouldn't matter much? (the out-of-tree build system worked fine until the latest pyodide)

hoodmane:
Well it's just the out of tree CLI runner. It uses a lazier dynamic linking strategy. Normal Pyodide does pessimistic eager loads which is easier to get right but is slow. It would be nice to use the lazier approach always. But for now, I took a different tradeoff so that `python -c 'print("hi")'` doesn't load a bunch of dsos that aren't needed and when dsos are loaded we don't preresolve GOT entries that aren't needed, etc. It really matters to make this work well in the long run.

hoodmane:
Anyways, the compile flags are:
```
-O3 -fPIC -fvisibility=hidden -fexceptions -g2
```
If we remove `-fvisibility=hidden` then `__attribute__((visibility("default")))` stops mattering. The full diff between the dumped crashing object file and the noncrashing object file is:
```patch
-okay.o:	file format wasm 0x1
+crash.o:	file format wasm 0x1
 
 Section Details:
 
@@ -58,7 +58,7 @@ Data[2]:
 Custom:
  - name: "linking"
   - symbol table [count=26]
-   - 0: F <PyInit__core> func=18 [ binding=global vis=hidden ]
+   - 0: F <PyInit__core> func=18 [ binding=global vis=default ]
    - 1: D <__THREW__> [ undefined binding=global vis=default ]
    - 2: F <env.__cxa_allocate_exception> func=0 [ undefined binding=global vis=default ]
    - 3: G <env.__memory_base> global=0 [ undefined binding=global vis=default ]
```

hoodmane:
Right now I have:

build.sh:
```sh
set -x

em++ \
-isystem \
.pyodide-xbuildenv-0.26.0/0.26.0/xbuildenv/pyodide-root/cpython/installs/python-3.12.1/include/python3.12 \
-c src/module.cpp \
-o module.cpp.o \
-O3 -fPIC -fexceptions -g2 \
#-fvisibility=hidden 

em++ \
-L.pyodide-xbuildenv-0.26.0/0.26.0/xbuildenv/pyodide-root/cpython/installs/python-3.12.1/lib/ \
-o dist/boost_histogram-1.4.2.dev14+g54eb2dd.d20240603/boost_histogram/_core.cpython-312-wasm32-emscripten.so \
module.cpp.o \
-s WASM_BIGINT -s SIDE_MODULE=1 -fexceptions -g2 -O2 \

wheel pack dist/boost_histogram-1.4.2.dev14+g54eb2dd.d20240603 -d dist
```
and `module.cpp`:
```cpp
#include "Python.h"
#include <stdexcept>

extern "C"
PyObject *PyInit__core() { 
    try { 
        throw std::runtime_error("");
    } catch (const std::exception &e) { 
        PyErr_SetString(PyExc_ImportError, "oops"); 
    } 
    return nullptr; 
} 
```
and `repro.py`:
```py
print("=" * 20)
try:
    from boost_histogram._core import do_the_thing
except Exception:
    print("ok!")
```
and `run.sh`:
```sh
rm -rf .venv-pyodide
pyodide venv .venv-pyodide
.venv-pyodide/bin/pip install dist/boost_histogram-1.4.2*.whl
.venv-pyodide/bin/python repro.py
```

hoodmane:
Okay I'm being silly, of course when we don't export `PyInit__core`, we get:
```py
ImportError('dynamic module does not define module export function (PyInit__core)')
```
instead of the desired crash but this isn't an interesting observation.

hoodmane:
Emscripten bug:
https://github.com/emscripten-core/emscripten/issues/22052

.github/workflows/emscripten.yaml
```diff
@@ -26,12 +26,12 @@ jobs:
...
         with:
-          python-version: "3.11"
+          python-version: "3.12"
 
       - name: Install pyodide-build
         run: |
-          pip install pyodide-build==0.25.1
-          echo EMSCRIPTEN_VERSION=$(pyodide config get emscripten_version) >> $GITHUB_ENV
+          pip install pyodide-build==0.26.0
+          echo EMSCRIPTEN_VERSION=$(pyodide config get emscripten_version | tail -n 1 | awk '{print $NF}') >> $GITHUB_ENV
```
henryiii, line 34:
```suggestion
          pip install pyodide-build==0.26.1
          echo EMSCRIPTEN_VERSION=$(pyodide config get emscripten_version >> $GITHUB_ENV
```

.github/workflows/emscripten.yaml
```diff
@@ -26,12 +26,12 @@ jobs:
...
         with:
-          python-version: "3.11"
+          python-version: "3.12"
 
       - name: Install pyodide-build
         run: |
-          pip install pyodide-build==0.25.1
-          echo EMSCRIPTEN_VERSION=$(pyodide config get emscripten_version) >> $GITHUB_ENV
+          pip install pyodide-build==0.26.1
+          echo EMSCRIPTEN_VERSION=$(pyodide config get emscripten_version >> $GITHUB_ENV
```
henryiii, line 34:
```suggestion
          echo EMSCRIPTEN_VERSION=$(pyodide config get emscripten_version) >> $GITHUB_ENV
```

pyproject.toml
```diff
@@ -117,6 +117,11 @@ sdist.exclude = [
 write_to = "src/boost_histogram/version.py"
 
 
+[tool.pyodide.build]
+cflags = ["-fexceptions"]
+ldflags = ["-fexceptions"]
```
henryiii, line 122:
```suggestion
cflags = "-fexceptions"
ldflags = "-fexceptions"
```

pyproject.toml
```diff
@@ -117,6 +117,11 @@ sdist.exclude = [
 write_to = "src/boost_histogram/version.py"
 
 
+[tool.pyodide.build]
+cflags = "-fexceptions"
```
henryiii, line 121:
```suggestion
cxxflags = "-fexceptions"
```

henryiii:
Curious, I naively would have thought:

```diff
- run: CFLAGS=-fexceptions LDFLAGS=-fexceptions pyodide build --exports whole_archive
+ run: pyodide build --exports whole_archive
```

Along with

```toml
[tool.pyodide.build]
cxxflags = "-fexceptions"
ldflags = "-fexceptions"
```

Would have been the same, but now it's reporting

```
Failed to load lib  /home/runner/work/boost-histogram/boost-histogram/.venv-pyodide/lib/python3.12/site-packages/boost_histogram/_core.cpython-312-wasm32-emscripten.so
Error: need the dylink section to be first
    at failIf (/home/runner/work/boost-histogram/boost-histogram/.pyodide-xbuildenv-0.26.1/0.26.1/xbuildenv/pyodide-root/dist/pyodide.asm.js:10:137604)
    at getDylinkMetadata (/home/runner/work/boost-histogram/boost-histogram/.pyodide-xbuildenv-0.26.1/0.26.1/xbuildenv/pyodide-root/dist/pyodide.asm.js:10:138151)
    at loadWebAssemblyModule (/home/runner/work/boost-histogram/boost-histogram/.pyodide-xbuildenv-0.26.1/0.26.1/xbuildenv/pyodide-root/dist/pyodide.asm.js:10:147672)
    at /home/runner/work/boost-histogram/boost-histogram/.pyodide-xbuildenv-0.26.1/0.26.1/xbuildenv/pyodide-root/dist/pyodide.asm.js:10:154278
    at async Object.loadDynlib (/home/runner/work/boost-histogram/boost-histogram/.pyodide-xbuildenv-0.26.1/0.26.1/xbuildenv/pyodide-root/dist/pyodide.asm.js:10:89059)
    at async main (/home/runner/work/boost-histogram/boost-histogram/.pyodide-xbuildenv-0.26.1/0.26.1/xbuildenv/pyodide-root/dist/python:139:13)
ImportError while loading conftest '/home/runner/work/boost-histogram/boost-histogram/tests/conftest.py'.
tests/conftest.py:15: in <module>
    import boost_histogram as bh
.venv-pyodide/lib/python3.12/site-packages/boost_histogram/__init__.py:3: in <module>
.venv-pyodide/lib/python3.12/site-packages/boost_histogram/accumulators.py:3: in <module>
    from ._core.accumulators import (  # pylint: disable=import-error,no-name-in-module
E   ImportError: dynamic module does not define module export function (PyInit__core)
```

.github/workflows/emscripten.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
           actions-cache-folder: emsdk-cache
 
       - name: Build
-        run: CFLAGS=-fexceptions LDFLAGS=-fexceptions pyodide build --exports whole_archive
+        run: pyodide build --exports whole_archive
```
henryiii, line 42:
```suggestion
        run: CXXFLAGS=-fexceptions LDFLAGS=-fexceptions pyodide build --exports whole_archive
```

pyproject.toml
```diff
@@ -117,6 +117,11 @@ sdist.exclude = [
 write_to = "src/boost_histogram/version.py"
 
 
+[tool.pyodide.build]
+cxxflags = "-fexceptions"
+ldflags = "-fexceptions"
+
+
```
henryiii, line 124:
```suggestion
```

hoodmane:
>  now it's reporting ...

Well that sounds like another thing to look into...

Issue #943, opened 2024-06-17
[BUG] Integer axes do not function correctly with numpy 2.0 installed


lgray:
with numpy 1.26.4:
```
>>> import boost_histogram as bh
>>> bh.axis.Integer(0,5)
Integer(0, 5)
```

with numpy 2.0:
```
>>> import boost_histogram as bh
>>> bh.axis.Integer(0,5)
Integer(5, 5)
```

This causes test failures in coffea and incorrectness in analysis using numpy 2.0. This is urgent.

lgray:
@henryiii

henryiii:
At USCMS. Will look at as soon as I can. Let me know if you have any more debugging insights. 0 turning into 5 is really odd.

lgray:
I am at USCMS as well. Hopefully it's something simple. I don't really have time to dig that deep into it.

Issue where we found this in coffea: https://github.com/CoffeaTeam/coffea/issues/1112

henryiii:
```pycon
Python 3.12.4 (main, Jun  6 2024, 18:26:44) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import boost_histogram as bh
>>> bh.axis.Integer(0,5)
Integer(0, 5)
>>> import numpy as np
>>> np.__version__
'2.0.0'
```

lgray:
OK - looking again I had a skewed package on my laptop that was causing the issue. 

Trying to figure out why it was a problem in coffea's CI though.

lgray:
OK - this was a bit of a red herring. Now that I've fixed up my local environment the issue seems to be some strange roundoff difference in np.histogram.

PR #944, opened 2024-06-17
chore: try running tests with numpy 2

lgray:
I want to see if tests fail with the recently released numpy 2.

Related to #943

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ classifiers = [
     "Typing :: Typed",
 ]
 dependencies = [
-    "numpy",
+    "numpy>=2",
     "typing-extensions;python_version<'3.8'",
 ]
 


lgray:
@henryiii can you please let these tests run so we can see what fails in numpy2?

henryiii:
`numpy>=2.0.0b1` is already present in the tests, so pretty sure this does work.

lgray:
I do not understand how the tests are passing:
https://github.com/scikit-hep/boost-histogram/blob/develop/tests/test_axis.py#L696-L700

Requires everything to be as expected, but:
```
>>> bh.axis.Integer(1, 3).edges
array([3., 3., 3.])
```

Something is rather fishy.

lgray:
Likewise:
```
>>> a = bh.axis.Integer(-1, 3)
>>> a
Integer(3, 3)
>>> a == (-1,0,1,2)
False
```

PR #969, opened 2025-01-16
chore: update cibw config

henryiii:
Using the new ARM runners.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -110,7 +110,7 @@ jobs:
           - os: ubuntu-latest
             only: cp313t-manylinux_x86_64
           - os: ubuntu-24.04-arm
-            only: cp313-manylinux_x86_64
+            only: cp313-manylinux_aarch64
           - os: windows-latest
             only: cp38-win32
           - os: windows-latest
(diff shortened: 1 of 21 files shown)


.github/workflows/tests.yml
```diff
@@ -109,11 +109,11 @@ jobs:
         include:
           - os: ubuntu-latest
             only: cp313t-manylinux_x86_64
-          - os: ubuntu-latest
+          - os: ubuntu-24.04-arm
             only: cp313-manylinux_x86_64
```
henryiii, line 113:
```suggestion
            only: cp313-manylinux_aarch64
```

.github/workflows/tests.yml
```diff
@@ -140,7 +140,7 @@ jobs:
           name: test-wheels-${{ strategy.job-index }}
 
       - name: Check wheels
-        run: pipx run twine check wheelhouse/*
+        run: uvx twine check wheelhouse/*
```
henryiii, line 143:
```suggestion
        run: pipx run twine check wheelhouse/*
```

henryiii:
Tried to use uvx instead, but ran into #10697 on the new ARM runners.

PR #972, opened 2025-01-23
fix: make non-uniform rebinning work for `Weight()` and friends

### LINKED ISSUE #971: [BUG] non-uniform rebinning does not work with `Weight()` storage
**Steps to reproduce**

```py
import boost_histogram as bh

hist = bh.Histogram(bh.axis.Regular(10, 0.0, 1.0), storage=bh.storage.Weight())
hist.fill(0.9)

rebin = bh.rebin(groups=[1,2,3,4])
hist[::rebin].view()
```

would result in -
```py
UFuncTypeError: ufunc 'add' did not contain a loop with signature matching types (dtype([('value', '<f8'), ('variance', '<f8')]), dtype([('value', '<f8'), ('variance', '<f8')])) -> None
```

Working on it!

Also see https://github.com/scikit-hep/hist/issues/345#issuecomment-2451674493

Saransh-cpp:
Fixes #971 

Now the code loops over every `dtype.name` if the `view()` of a `Histogram` is a structured NumPy array.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -42,12 +42,12 @@ specific jobs:
 ```console
 $ nox -l # List all the defined sessions
 $ nox -s lint  # Lint only
-$ nox -s tests-3.9  # Python 3.9 tests only
+$ nox -s tests  # Tests only
 $ nox -s docs -- serve  # Build and serve the docs
 $ nox -s make_pickle  # Make a pickle file for this version
 ```
 
-Nox handles everything for you, including setting up an temporary virtual
+Nox handles everything for you, including setting up a temporary virtual
 environment for each run.
 
 ### Pip
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -2,6 +2,12 @@
 
 ## Version 1.5
 
+### Version 1.5.1
+
+Make non-uniform rebinning work for Weight() and friends [#972][]
+
+[#972]: https://github.com/scikit-hep/boost-histogram/pull/972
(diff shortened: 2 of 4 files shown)


matthewfeickert:
cc @andrzejnovak

andrzejnovak:
image

Works (and propagates to hist!). Thanks a lot!

matthewfeickert (approved):

matthewfeickert:
@Saransh-cpp I'll let you merge in the event there's any final small things you wanted to get in here.

Saransh-cpp:
Thanks! I don't have merge/write access to boost-histogram. Please feel free to merge :)

matthewfeickert:
Ah okay I'll also ask @henryiii for review and merge then.

henryiii:
Are all the fields addable? That is, for every type, can you just directly add the fields? For example, for the Mean storage, this is the definition of `__iadd__`: https://github.com/scikit-hep/boost-histogram/blob/df21e449f7d54dcffcf6e1c31b8d7f473c6f4a00/include/bh_python/accumulators/mean.hpp#L61-L77

I think we can just use this, though, by exposing it via a ufunc. I'm looking at this soon, but got distracted updating and fixing things in #973.

henryiii:
Fixed. Thanks!

matthewfeickert:
Thanks @henryiii and @Saransh-cpp!

Would it be possible to get a patch release with this out?

henryiii:
1.5.1 is out. :)

PR #975, opened 2025-01-29
chore: move some things out of internal

henryiii:
This is an attempt to reduce the complexity of boost-histogram a little now that we can use `__dir__` (Python 3.7+).


noxfile.py
Saransh-cpp, line 1: I remember adding this hack -

https://github.com/scikit-hep/boost-histogram/blob/a30cc81429ac0bd0cf4c24b73056ed1c88dea9cd/noxfile.py#L101-L113

sometime back. It can be removed with the changes in this PR.

henryiii: Thanks!

PR #977, opened 2025-01-31
feat: allow rebinning by passing edges or a new axis

andrzejnovak:
- **feat: allow rebinning by passing edges (or directly a hist.axis)**
- **fix: preserve flow when rebinning**

### DIFF
--- a/src/boost_histogram/tag.py
+++ b/src/boost_histogram/tag.py
@@ -123,6 +123,7 @@ class rebin:
     def __init__(
         self,
         factor_or_axis: int | PlottableAxis | None = None,
+        /,
         *,
         factor: int | None = None,
         groups: Sequence[int] | None = None,
@@ -144,6 +145,9 @@ class rebin:
             self.factor = factor_or_axis
         elif factor_or_axis is not None:
             self.axis = factor_or_axis
+        else:
+            msg = "Unsupported type, only int or Axis supported for the first argument"
+            raise TypeError(msg)
 
     def __repr__(self) -> str:
         repr_str = f"{self.__class__.__name__}"
--- a/src/boost_histogram/tag.py
+++ b/src/boost_histogram/tag.py
@@ -122,24 +122,28 @@ class rebin:
 
     def __init__(
         self,
-        factor: int | PlottableAxis | None = None,
+        factor_or_axis: int | PlottableAxis | None = None,
         *,
+        factor: int | None = None,
         groups: Sequence[int] | None = None,
         edges: Sequence[int | float] | None = None,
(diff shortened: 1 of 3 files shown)


andrzejnovak:
Hi @henryiii @Saransh-cpp . I've added the requested rebin by passing axis as well now. I also noticed the existing implementation was skipping flow bins so I've added that back in (keeps total sumw/sumw2 same). Should be good to review.

While testing, there's one more thing I noticed, but not sure if easily fixable. The repr changes between the original and the transformed hist. Something in the metadata copying is not 1-to-1.
image

src/boost_histogram/tag.py
```diff
@@ -112,26 +114,40 @@ def __call__(self, axis: AxisLike) -> int:  # noqa: ARG002
...
+        "axis",
+        "edges",
         "factor",
         "groups",
     )
 
     def __init__(
         self,
-        factor: int | None = None,
+        factor: int | PlottableAxis | None = None,
```
henryiii, line 125: If you want, you could make this a positional only optional arg, then assign to the correct one if it's given. Something like:

```python
self, factor_or_axis, /, factor=None, axis=None, ...)
```

Might make typing easier.

src/boost_histogram/tag.py
```diff
@@ -112,26 +114,44 @@ def __call__(self, axis: AxisLike) -> int:  # noqa: ARG002
...
+        "axis",
+        "edges",
         "factor",
         "groups",
     )
 
     def __init__(
         self,
-        factor: int | None = None,
+        factor_or_axis: int | PlottableAxis | None = None,
```
henryiii, line 125:
```suggestion
        factor_or_axis: int | PlottableAxis | None = None,
        /,
```

src/boost_histogram/tag.py
```diff
@@ -112,26 +114,44 @@ def __call__(self, axis: AxisLike) -> int:  # noqa: ARG002
...
+        ):
+            raise ValueError("Exactly one argument should be provided")
         self.groups = groups
+        self.edges = edges
+        self.axis = axis
+        self.factor = factor
+        if isinstance(factor_or_axis, int):
+            self.factor = factor_or_axis
+        elif factor_or_axis is not None:
+            self.axis = factor_or_axis
```
henryiii, line 146:
```suggestion
            self.axis = factor_or_axis
        else:
            msg = "Unsupported type, only int or Axis supported for the first argument"
            raise TypeError(msg)
```

henryiii: Oops, no, ignore this.

henryiii:
Thanks!

PR #978, opened 2025-01-31
fix: axis metadata was broken when rebinning

henryiii:
The metadata was getting an extra `{"metadata": metadata}` wrapped around it (and `None` was turning into `{}`). Seen in https://github.com/scikit-hep/boost-histogram/pull/977#issuecomment-2627746523.


henryiii:
As a followup, I think I'd like to rename `.metadata` on the C++ object, since it is not the same as `.metadata` on the Python one, and there's a high chance of confusion. It's actually the same as `__dict__`, which is likely a bad name, but maybe `.metadata_dict` or something like that would be less likely to be confused. Really anything that's not identical to `.metadata`. ;)

PR #980, opened 2025-01-31
fix: axis passthrough

henryiii:
Followup to #977. Uses the new axes if one is provided.

I wonder if we should allow axis + other args, if someone wants to use a new axis with a custom mapping?

We'll have to work out the correct mapping for flow bins for the four cases.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -2,9 +2,31 @@
 
 ## Version 1.5
 
+### Version 1.5.2
+
+Fix for axis metadata not passing though non-uniform rebinning correctly. Flow
+bins are now preserved when doing a non-uniform rebinning. Also adds the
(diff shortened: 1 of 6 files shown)


henryiii:
What would you expect to happen if you rebin an axis that doesn't have flow? Specifically:

```python
h = bh.Histogram(bh.axis.Regular(5, 0, 5, underflow=False, overflow=False))
h.fill([-1, 1.1, 2.2, 3.3, 4.4, 5.5])
hs = h[bh.rebin(edges=[0, 3, 5.0])]
```

Should `hs.axis[0]` have flow bins?

henryiii:
Actually, I think it should match `rebin(2)`, which matches the original. :)

PR #988, opened 2025-03-06
chore: use PEP 639 licenses (METADATA 2.4)

henryiii:
Supported now.


.readthedocs.yml
```diff
@@ -13,6 +13,7 @@ build:
     - asdf plugin add uv
     - asdf install uv latest
     - asdf global uv latest
+    - touch extern/histogram/LICENSE
```
henryiii, line 16:
```suggestion
```

PR #1001, opened 2025-04-15
ci: add Windows ARM

henryiii:
Adding ARM wheels.


mattip:
Cool! windows-11-arm64 runners!

henryiii:
@mattip does that mean we'll get PyPy for Windows ARM? :)

henryiii:
And numpy, actually...

mattip:
NumPy starts with OpenBlas, and someone contributed a batch file for win-arm64 so that might happen soon. Pypy has so many problems with windows i dont know if it is worth it,

PR #1002, opened 2025-04-17
chore: use `np.s_` for `Slicer`

pfackeldey:
This replaces the custom `Slicer` class logic with `np.s_`. For backward compatibility I made `Slicer()` return `np.s_` and added a note in the doc string.


henryiii:
@all-contributors please add @pfackeldey for code

pfackeldey:
Thanks @henryiii, somehow I missed the usage of it in the tests!

henryiii:
Thanks!

Issue #83, opened 2019-08-26
Support for multiple weight variations


douglasdavis:
A [feature](https://pygram11.readthedocs.io/en/stable/api/var1dmw.html) I recently implemented in pygram11 allows an array of input data to be histogrammed with multiple weight variations in a single function call. In ATLAS (and I'm sure most HEP experiments) we carry around _a lot_ of MC generator weights and derive uncertainties from many scale factor variations-- in the downstream parts of an analysis we end up comparing a histogrammed variable using one set of weights to the same variable histogrammed with a different weight variation. For a histogram with non-fixed width binning there's actually a nice speedup from avoiding repeating the binary search to grab the necessary bin index.

I'm not sure how this might fit into the object-focused design of boost-histogram, but I just wanted to mention it as an idea for a feature-- perhaps if there ends up being some kind of factory to create histograms from a function call which has the data to be histogrammed as an argument.

HDembinski:
If I understand you correctly, you want to store not only one sum of weights for each histogram cell, but several sums of weights. It is possible to achieve this with the current machinery, by writing a special cell accumulator and using the `sample` argument for the fill method.

> perhaps if there ends up being some kind of factory to create histograms from a function call which has the data to be histogrammed as an argument.

The new fill method allows you pass arrays with the input data. It is not a factory, so you first generate the histogram and then pass the data, but we could easily add a function which does these two steps with a single call.

HDembinski:
I will think about adding an example to the user guide about this. I don't have many examples yet for custom accumulators that use samples.

HDembinski:
Thinking more about it, it is probably even more straight-forward to support this. The default weight type in boost::histogram is double, but this can be replaced by another multi-weight type. The type has to be additive, multiplicative, and broadcastable.

HDembinski:
@douglasdavis How many weights do you need to keep track of simultaneously? 2, 4, 16?

henryiii:
I think @douglasdavis simply wants this:

```python
h = bh.histogram(bh.axis.regular(10,0,1), storage=bh.storage.double)
h.fill([1,2,3], weight=[1,1,1])
h.fill([1,2,3], weight=[2,2,2])
h.fill([1,2,3], weight=[3,3,3])
```

To be expressible as:

```python
h = bh.histogram(bh.axis.regular(10,0,1), storage=bh.storage.double)
h.fill([1,2,3], weights=[[1,1,1], [2,2,2], [3,3,3]])
```

[...]

PR #1008, opened 2025-05-26
Multi weight storage

Superharz:
This PR adds a multi weight storage type (called `MultiWeight`) to boost-histogram. It addresses #83 and is bases on a prototype provided by @HDembinski in https://github.com/boostorg/histogram/issues/211.

It allows one to create a histogram that can store multiple independent weights per bin (the number of weights per bin has to be specified when creating the histogram).
## Example
Create and fill a 1 dim histogram that stores 3 weights per bin:
```
import boost_histogram as bh
import numpy as np
x = np.array([1, 2])
weights = np.array([
        [1, 2, 3],
        [4, 5, 6]
    ])
h = bh.Histogram(bh.axis.Regular(5, 0, 5), storage = bh.storage.MultiWeight(3))
h.fill(x, sample = weights)

h[1] = [1, 2, 3]
h[2] = [4, 5, 6]
```

## Status of this PR
The MultiWeight histograms can be created in python and they can be filled.
Pickle also works.

The buffer and view for the histograms has yet to be implemented.


Superharz:
`h.values()` does not work yet.
However, the important part (for me) is to fill the histograms not to retrieve the data from them.
At the end, the data can be converted into a numpy array by a small loop over the histogram:

For a 8*8 2D histogram `h` with 20k weights:
```python
a = np.zeros((8,8, 20000))
for i in range(8):
    for j in range(8):
        a[i,j] = np.array(h[i,j])
```

include/bh_python/fill.hpp
```diff
@@ -212,6 +212,36 @@ void fill_impl(bh::detail::accumulator_traits_holder<true, const double&>,
...
+// for multi_weight
+template <class Histogram, class VArgs>
+void fill_impl(bh::detail::accumulator_traits_holder<false, boost::span<double>>,
+               Histogram& h,
+               const VArgs& vargs,
+               const weight_t& weight,
+               py::kwargs& kwargs) {
+    // weight is not used, "use" it once to suppress "unused variable" complaints by
+    // compiler
+    (void)weight;
```
HDembinski, line 224: You can use https://www.boost.org/doc/libs/latest/libs/core/doc/html/core/ignore_unused.html to make it self-explanatory.

include/bh_python/multi_weight.hpp
```diff
@@ -0,0 +1,271 @@
...
+#include <boost/histogram.hpp>
+#include <boost/histogram/detail/iterator_adaptor.hpp>
+#include <iostream>
+#include <memory>
+
+namespace boost {
+namespace histogram {
+
+template <class T>
+struct multi_weight_value : public boost::span<T> {
```
HDembinski, line 16: The naming is misleading. This inherits from boost::span, which is a view type, but the name `X_value` suggest it is a value type. A value type holds its data, while a view does not. So this is a multi_weight_span or multi_weight_view. Probably the latter.

include/bh_python/multi_weight.hpp
```diff
@@ -0,0 +1,271 @@
...
+            *it++ += x;
+    }
+    void operator+=(const boost::span<T> values) {
+        if(values.size() != this->size())
+            throw std::runtime_error("size does not match");
+        auto it = this->begin();
+        for(const T& x : values)
+            *it++ += x;
+    }
+    void operator=(const std::vector<T> values) {
```
HDembinski, line 48: Also here, why not use a span?

include/bh_python/multi_weight.hpp
```diff
@@ -0,0 +1,271 @@
...
+        for(const T& x : values)
+            *it++ = x;
+    }
+};
+
+template <class ElementType = double>
+class multi_weight {
+  public:
+    using element_type    = ElementType;
+    using value_type      = multi_weight_value<element_type>;
```
HDembinski, line 61: That's not a value type, because multi_weight_value is a view.

include/bh_python/multi_weight.hpp
```diff
@@ -0,0 +1,271 @@
...
+        } else {
+            w.assign(buffer_.get(), buffer_.get() + buffer_length_);
+            ar& make_nvp("buffer", w);
+        }
+    }
+
+  public:
+    std::size_t size_          = 0;
+    std::size_t nelem_         = 0;
+    std::size_t buffer_length_ = 0;
```
HDembinski, line 187: Some comments are needed to explain these sizes. It seems there should be only two, not three.

Superharz: You are correct. I added `buffer_length_ = size_ * nelem_` to not always have to type `size_ * nelem_` in the code

include/bh_python/multi_weight.hpp
HDembinski: You should make a multi_weight_base with common code for all of these, like the operators.

include/bh_python/multi_weight.hpp
```diff
@@ -0,0 +1,271 @@
...
+    }
+};
+
+template <class ElementType = double>
+class multi_weight {
+  public:
+    using element_type    = ElementType;
+    using value_type      = multi_weight_value<element_type>;
+    using reference       = value_type;
+    using const_reference = const value_type&;
```
HDembinski, line 63: That does not work, the const_reference must be its own type, similar to reference, but prevent manipulation of its content.

include/bh_python/register_histogram.hpp
```diff
@@ -211,3 +211,193 @@ auto register_histogram(py::module& m, const char* name, const char* desc) {
...
+                   py::self /= py::self);
+    def_optionally(hist,
+                   bh::detail::has_operator_rmul<histogram_t, histogram_t>{},
+                   py::self *= py::self);
+    def_optionally(hist,
+                   bh::detail::has_operator_rsub<histogram_t, histogram_t>{},
+                   py::self -= py::self);
+#ifdef __clang__
+#pragma GCC diagnostic pop
+#endif
```
HDembinski, line 290: Boost has nice macros for this, please use them.

include/bh_python/multi_weight.hpp
HDembinski: Still better to compute this on the fly. If it bothers you, you can make a method to return the product and call it capacity. It will be inlined by the optimizer.

include/bh_python/multi_weight.hpp
```diff
@@ -0,0 +1,271 @@
...
+            ar& make_nvp("buffer", w);
+        }
+    }
+
+  public:
+    std::size_t size_          = 0;
+    std::size_t nelem_         = 0;
+    std::size_t buffer_length_ = 0;
+    std::unique_ptr<element_type[]> buffer_;
+    std::unique_ptr<value_type[]> span_buffer_;
```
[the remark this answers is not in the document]
HDembinski: What I wrote in another comment: implement const_reference as another span-type, analog to your current multi_weight_value, which is also a view type. In the const_reference version, you must disallow modification, so no operator=, operator+=, etc., and only const reference access to elements.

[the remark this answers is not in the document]
HDembinski: Yes, that's a bit annoying, but writing C++ on this level is a bit like making legal contracts.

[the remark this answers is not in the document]
HDembinski: You can find some inspiration when you look into the implementation of unlimited_storage.hpp. You will see that the storage has implementations for a reference and const_reference type. Fortunately, yours will be much simpler.

include/bh_python/multi_weight.hpp
```diff
@@ -0,0 +1,271 @@
...
+
+    template <class T                                               = element_type,
+              std::enable_if_t<!std::is_arithmetic<T>::value, bool> = true>
+    void default_fill() {}
+
+    template <class T                                              = element_type,
+              std::enable_if_t<std::is_arithmetic<T>::value, bool> = true>
+    void default_fill() {
+        std::fill_n(buffer_.get(), buffer_length_, 0);
+    }
```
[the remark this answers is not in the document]
HDembinski: Ok, that is true, but I believe that the element_type in this case is always an arithmetic type, we don't use objects as weights in the library, only floating point types.

include/bh_python/fill.hpp
```diff
@@ -212,6 +212,36 @@ void fill_impl(bh::detail::accumulator_traits_holder<true, const double&>,
...
+    std::size_t buf_shape0 = static_cast<std::size_t>(buf.shape[0]);
+    std::size_t buf_shape1 = static_cast<std::size_t>(buf.shape[1]);
+    double* src            = static_cast<double*>(buf.ptr);
+    std::vector<boost::span<double>> vec_s;
+    vec_s.reserve(buf_shape0);
+    for(std::size_t i = 0; i < buf_shape0; i++) {
+        vec_s.emplace_back(boost::span<double>{src + i * buf_shape1, buf_shape1});
+    }
+    // releasing gil here is safe, we don't manipulate refcounts
+    py::gil_scoped_release lock;
```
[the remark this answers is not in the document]
Superharz: Resolved by d05f7a7

include/bh_python/multi_weight.hpp
```diff
@@ -0,0 +1,271 @@
...
+    static constexpr bool has_threading_support() { return false; }
+
+    multi_weight(const std::size_t k = 0)
+        : nelem_{k} {}
+
+    multi_weight(const multi_weight& other)
+        : nelem_{other.nelem_} {
+        reset(other.size_);
+        std::copy(
+            other.buffer_.get(), other.buffer_.get() + buffer_length_, buffer_.get());
```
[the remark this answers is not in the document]
Superharz: Resolved by 91d6601

include/bh_python/multi_weight.hpp
Superharz: Resolved by 91d6601

include/bh_python/multi_weight.hpp
```diff
@@ -0,0 +1,271 @@
...
+        buffer_length_ = n * nelem_;
+        buffer_.reset(new element_type[buffer_length_]);
+        default_fill();
+        span_buffer_.reset(new value_type[size_]);
+        std::size_t i = 0;
+        std::generate_n(span_buffer_.get(), size_, [&]() {
+            auto tmp_span = value_type{buffer_.get() + i * nelem_, nelem_};
+            i++;
+            return tmp_span;
+        });
```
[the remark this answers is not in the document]
Superharz: Resolved by 91d6601

include/bh_python/multi_weight.hpp
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    bool operator==(const boost::span<T> values) const {
+        if(values.size() != this->size())
+            return false;
+
+        return std::equal(this->begin(), this->end(), values.begin());
+    }
+    bool operator!=(const boost::span<T> values) const { return !operator==(values); }
+    void operator+=(const std::vector<T> values) {
```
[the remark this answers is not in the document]
Superharz: I have tried to address this in b2cffcf.
However, I can not template `+=` to accept a `class S` because this results in a huge compiler error message which says
```
/cvmfs/sft.cern.ch/lcg/views/LCG_107a/x86_64-el9-gcc14-opt/include/boost/histogram/detail/accumulator_traits.hpp:81:37: error: call of overloaded 'accumulator_traits_impl(boost::histogram::multi_weight_value<double>&, boost::histogram::detail::priority<2>)' is ambiguous
   81 |     decltype(accumulator_traits_impl(std::declval<T&>(), priority<2>{}));
```

include/bh_python/multi_weight.hpp
Superharz: Okay, if I understand your suggestion right, my idea would be to have
1.  `multi_weight_value` derived from a `std::vector` that adds assignment and sum operators
2.  `multi_weight_const_reference` derived from a `boost::span` that does not define any operators to modify the data
3. `multi_weight_reference` derived from `multi_weight_const_reference` that adds the assignment and sum operators

Should I better use a `std::valarray` instead of `std::vector`?

Superharz:
Hi @HDembinski,
I addressed the first batch of your comments.
For the remaining ones I need some input from your side on how to continue :)

include/bh_python/multi_weight.hpp
Superharz: Partially resolved by 91d6601.
Will be followed up in https://github.com/scikit-hep/boost-histogram/pull/1008#discussion_r2108396421.

include/bh_python/multi_weight.hpp
Superharz: @HDembinski I tried to address this with my latest commits.

HDembinski: Better not inherit, but use composition. A derivation implies an "is a" relationship, and that's not true. A multi_weight_value is not some kind of std::vector, not all of the methods of a standard vector make sense for a multi_weight_value.

HDembinski: > Btw, should it not be enough to define
> `using const_reference = const reference` (without the `&`)
> because this would automatically prohibit the use of any function that is not defined `const`?

Maybe. Usually, I had a reason why I did things in a certain way, but you might be right here. Implementing const_reference without the mutable methods and then inherit a reference from const_reference is cleaner, because then the methods that shouldn't be called aren't there. That being said, I can see the appeal of implementing only one kind of reference.

include/bh_python/multi_weight.hpp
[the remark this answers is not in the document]
HDembinski: You should get it to work with `boost::span`, because then it works with any contiguous memory container without a copy. Also it should be `const std::vector<T>& values` instead of `const std::vector<T> values`, the latter is passing by value so a copy is created even when the vector type matches exactly.

include/bh_python/multi_weight.hpp
HDembinski: Maybe we have to implement a copy-on-write mechanism here to meet the requirements, I don't know.

I think that's why this got stuck, the design wasn't clear.

HDembinski: > > Btw, should it not be enough to define
> > `using const_reference = const reference` (without the `&`)
> > because this would automatically prohibit the use of any function that is not defined `const`?
> 
> Maybe. Usually, I had a reason why I did things in a certain way, but you might be right here. Implementing const_reference without the mutable methods and then inherit a reference from const_reference is cleaner, because then the methods that shouldn't be called aren't there. That being said, I can see the appeal of implementing only one kind of reference.

Something was bothering me about this and so I checked. And I was right, you cannot implement a const reference like that, look:

```cpp
#include <boost/core/span.hpp>
#include <vector>
#include <cassert>

using mutable_span = boost::span<double>;
using const_span = const boost::span<double>; // not really const

const_span foo(const_span x) {
    x[1] = 0; // oopsie
    return x;
}

int main()
{
  std::vector<double> x = {1, 2, 3};
  
  mutable_span y = foo(x);

  y[0] = 0;

  assert(x[0] == 0); // oopsie
  assert(x[1] == 0); // oopsie
}
```

The `const` modifier on a value type (no &) doesn't prevent you from calling methods which mutate its contents. But even if that were true, there is no way in C++ to prevent initializing a mutable_span from a const_span, because you can also do:
```cpp
const double x = 3;
double y = x;
```

henryiii:
Upstream: https://github.com/boostorg/histogram/pull/411. I assume we can work on this, and move to upstream whenever that goes in and gets released. I've fixed up the clang-tidy recommendations, and started adding some of the missing things. I added `.nelem`, and added it to the repr. It also needs to support `.view`, `.values`, etc.

Superharz:
Commit 962b783 now adds functionality for the `reduce` and `project` algorithms for the multi weight storage by overloading how this storage is default constructed from a previous multi weight storage object.

This also enables slicing the histogram with a selection that only contains slices (this case uses directly the `reduce` algorithm from the c++ histograms).

Superharz:
Okay, tests now also run on Linux.

Superharz:
Should I also add a docs page solely dedicated to the `MultiWeight` storage to show some examples and maybe provide some benchmarks (e.g. how much computing time one can save with this storage in high statistics precision analysis)?

henryiii:
I'm curious, could we remap `weight=` in Python to `sample` in C++ for this storage? I'm fine with it as is, but just curious if that would work.

henryiii:
From a user's perspective, these are logically weights. They are not samples. So someone using boost-histogram would have to learn this peculiarity of implementation in order to use this feature, which they don't care about. Currently, these do the same thing:

```python
normal.fill(3, weight=2)
multi.fill(3, sample=[2])
```

We are implementing this ourselves, rather than wrapping something upstream, so I think we have flexibility to make this weight, and leave the fact we had to use `bh::sample` instead of `bh::weight` as an implementation detail. Maybe eventually this could even be fixed upstream? I'd say it's up to @HDembinski, though.

henryiii:
> argue here that this choice is not introducing other inconsistencies in the future.

That's why I think you should make the call here. From what you've said, it sounds like the detail that `sample` is used internally can be removed in the future, and we can cleanly swap to using it without a Python API change?

Superharz:
> I like the name MultiCell fine, I was more interested in the API for filling; `weight=[...]` fits with the API on all the other histograms better than using sample, which is an implementation detail. `MultiWeight` to me seems like it should behave more like the other storages with `Weight` in the name and track variances, which it does not.
> 
> So how about keeping the name `MultiCell`, but using the `weight=` keyword?

Hi @HDembinski and @henryiii, I have now implemented to use the `weight` keyword for `MultiCell` filling by simply switching the `sample` and `weight` input to the `hist.fill()` function.

This is only used on the high level `fill` function (the one in `histogram.py`), the low level `hist._hist.fill()` function still uses the `sample` keyword.

This allows the normal python user to use the `weight` keyword.

henryiii:
@all-contributors please add @Superharz for code

HDembinski:
Also from me, kudos to your endurance!

Superharz:
As a last request, I would like to kindly ask for a new boost-histogram release. I have some collaborators who would use the new histograms and it would be convenient if they would not have to self compile the module.

PR #1045, opened 2025-08-12
chore: adding Android wheels

henryiii:
Adding android wheels. Requires https://github.com/pypa/cibuildwheel/pull/2547.


mhsmith:
I don't see any obvious explanation in the log. Can the ImportError be improved to give more information?

PR #1057, opened 2025-11-05
feat: drop 3.9

henryiii:
Dropping 3.9, which is past EoL.


henryiii:
@copilot, please undo the change to StrEnum, and update the comment to state it is a Python 3.11 feature, not 3.10.

PR #1068, opened 2026-01-15
Avoid compiler warning: declaration of 'i' hides previous local declaration

Superharz:
When compiling boost-histogram on Windows, I get the following annoying warning:

```
\boost-histogram\include\bh_python\vector_string_caster.hpp(64,29): warning C4456: 
declaration of 'i' hides previous local declaration [\AppData\Local\Temp\tmpe4tm749r\build\_core.vcxproj]
    (compiling source file '/boost-histogram/src/register_axis.cpp')
        \boost-histogram\include\bh_python\vector_string_caster.hpp(59,25):
        see declaration of 'i'
```

I "fixed" it by renaming the inner iteration variable from `i -> j`

### DIFF
--- a/include/bh_python/vector_string_caster.hpp
+++ b/include/bh_python/vector_string_caster.hpp
@@ -61,10 +61,10 @@ struct type_caster<std::vector<std::string>>
             const auto n = strlen(p, step);
             std::string s;
             s.reserve(n);
-            for(std::size_t i = 0; i < n; ++i) {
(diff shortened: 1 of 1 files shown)


Superharz:
> I wonder if we can get warnings like that to fail the build on Windows in our CI?

I am sorry, but I don't have any expertise with CI builds.
I am only annoyed by warnings that are not caused by myself (as I often get enough of the later ones...)

PR #1087, opened 2026-02-03
fix: correct repr for MultiCell

henryiii:
Followup to #1008. This fixes the repr for MultiCell, which currently always shows `0`, since we can't get the actual instance, but just it's type (without the value of nelem).

I've added a `h.get_storage()` method which gets a storage object with the same parameters, an instance of `h.storage_type`. It's natural for the repr, and allows users to query with `h.get_storage().nelem`. Thoughts, @hdembinski and @Superharz? I could make it a property instead, such as `h.storage.nelem`. The main issue is it is just the `bh.storage` creation object, and doesn't actually give access to storage, so you can't do much with `h.storage`.

Edit: I realized that making this a property actually works better for pattern matching, which expects properties to match constructor arguments, so I've used `h.storage` now.

I thought about exposing `.nelem` on the histogram, but I can't properly statically type it (mypy doesn't allow `@property` and `@typing.overload` to be combined), and it's just None on most histograms.

### DIFF
--- a/src/boost_histogram/histogram.py
+++ b/src/boost_histogram/histogram.py
@@ -814,7 +814,8 @@ class Histogram(typing.Generic[S]):
         )
         return cast(self, self._hist._storage_type, Storage)  # type: ignore[return-value]
 
-    def get_storage(self) -> S:
+    @property
+    def storage(self) -> S:
         """
         New storage matching the one the histogram was constructed with.
         """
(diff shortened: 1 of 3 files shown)


Superharz:
Hi @henryiii, I am definitely in favor of having the correct number of cells contained in the representation. I also noticed that right now it always says `MultiCell(0)` regardless of the specific number of cells.

HDembinski:
Lgtm

PR #1109, opened 2026-04-11
feat: add flow=False support to histogram projections

### LINKED ISSUE #1032: [FEATURE] Add the option to disable flow when using project
When projecting to an axis of a multi-dimensional histogram, the flow bins are always included. 
This may be desired by default but it would be nice to be able to discard the flow bins. 
For example, one might want to make a cut on one axis: 
```python
h_new = h_old[10j:50j, :, :, :]
```
and then check how the other variables look like after the cut:
```python
h_new.project("axis_2")
```
This is going to be the same as `h_old.project("axis_2")` since flow is always included.
What I am suggesting is:
```python
h_new.project("axis_2", flow=False)
```
which would show what happens to `axis_2` after the cut on `axis_0`.

Rishabh-git10:
Closes #1032

Implements `flow=False` for histogram projections. Since the C++ backend doesn't natively support this, it handles the logic on the Python side by using `_core.algorithm.slice_mode.crop` to strip the flow bins from the integrated-out axes prior to calling `.project()`. 

Tests are included to verify the behavior.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## WIP
+
+#### Features
+
+- Support `flow=False` in histogram projections by @Rishabh-git10 in [#1109][]
+
 ## Version 1.7
 
 ### Version 1.7.2
@@ -7,7 +13,6 @@
 #### Features
 
 - No storage option/support for serialization, by @henryiii in [#1098][]
-- Support `flow=False` in histogram projections by @Rishabh-git10 in [#1109][]
 
 #### Fixes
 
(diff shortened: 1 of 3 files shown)


docs/changelog.md
```diff
@@ -7,6 +7,7 @@
 #### Features
 
 - No storage option/support for serialization, by @henryiii in [#1098][]
+- Support `flow=False` in histogram projections by @Rishabh-git10 in [#1109][]
```
henryiii: This needs to be moved up to a WIP section, as 1.7.2 shipped before this PR.

Rishabh-git10: Moved it up to the WIP section. Thanks!

Rishabh-git10:
@henryiii friendly ping! Anything else needed for this PR?

henryiii:
I'd like a ND test (like 2D), but not a blocker, thanks!

PR #1111, opened 2026-04-22
test: add ND projection test for flow=False

Rishabh-git10:
Follow-up to #1109. 

Adds a 3D projection test to verify `flow=False` behavior across multiple preserved dimensions, as requested by @henryiii [here](https://github.com/scikit-hep/boost-histogram/pull/1109#issuecomment-4296707176).

### DIFF
--- a/tests/test_histogram.py
+++ b/tests/test_histogram.py
@@ -632,6 +632,27 @@ def test_project():
     assert h_flow_false[0] == 2
 
 
+def test_project_nd():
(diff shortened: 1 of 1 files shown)


henryiii:
Sweet, you beat me to it. :) (I've fired up an agent investigating what would be needed to add this to C++ boost histogram, though).

Rishabh-git10:
Thanks! Let me know if I can help with anything else.

Issue #1115, opened 2026-05-09
AI review (bugs, perf, simp, best practices)


henryiii:
:robot:

I've completed a thorough review of the boost-histogram repository. Here's a consolidated summary of findings across **bugs, performance, simplifications, and best practices**:

---

## 🔴 Critical Bugs (Fix Immediately)

| # | Location | Issue |
|---|----------|-------|
| 1 | `src/boost_histogram/histogram.py:162` | `mean_storage_sample_check` has an **always-False** branch. `isinstance(sample, str) and not isinstance(sample, seqs)` is impossible because `str` is a `Sequence`. The intent to reject strings is silently broken. |
| 2 | `src/boost_histogram/histogram.py:1476` | Error message in `fill()` uses wrong variable: `value_shape[n]` should be `value_shape[value_n]`, producing a misleading/confusing error. |
| 3 | `tests/test_histogram.py:1277` | `platform.python_implementation == "CPython"` is missing `()`, so the check is always `False`. The refcount assertion runs on *all* implementations (including PyPy) where it may fail. |
| 4 | `include/bh_python/register_histogram.hpp:52` | **Memory leak** in `__deepcopy__`: raw `new` is used, and if `copy.deepcopy` on metadata throws a Python exception, the pointer is never freed. Same issue in `register_axis.hpp:210`. |
| 5 | `include/bh_python/make_pickle.hpp:177` | **O(n²) serialization**: `tuple_oarchive` concatenates tuples in a loop (`tup_ = tup_ + py::make_tuple(obj)`). For large histograms, pickle dominates CPU time. |

---

## 🟠 High-Impact Simplifications / Debt

| # | Location | Issue |
|---|----------|-------|
| 6 | `src/boost_histogram/histogram.py:170` | Deprecated `_arg_shortcut` developer tuple shortcut emits `FutureWarning`. It's been there for many releases — should be removed. |
| 7 | `src/boost_histogram/histogram.py:489` | `.metadata` access is still soft-deprecated (returns `None` with `FutureWarning`). Comment says "1.7+ will error." If the project is past 1.7, this should raise `AttributeError`. |
| 8 | `noxfile.py:119` | Pins `pylint==4.0.*` which **does not exist**. This nox session will fail on any fresh install. |
| 9 | `pyproject.toml:94-96` | `pytest-benchmark` is listed **twice** in the `test` dependency group. |
[...]

PR #1116, opened 2026-05-09
fix: several fixes from automated review

henryiii:
See #1115.

Assisted-by: OpenCode:Kimi-K2.6


tests/test_histogram.py
```diff
@@ -1307,7 +1274,7 @@ def test_axes_lifetime():
 
     ax = h.axes[0]
 
-    if platform.python_implementation == "CPython":
+    if platform.python_implementation() == "CPython":
         # 2 is the minimum refcount, so the *python* object should be deleted
         # after the del; hopefully the C++ object lives through the axis instance.
         assert sys.getrefcount(h) == 2
```
henryiii:
```suggestion
        # On Python 3.14, this can be 1
        assert sys.getrefcount(h) <= 2
```

PR #1133, opened 2026-06-08
feat: add Collector and WeightedCollector storages that keep original per-bin values

### LINKED ISSUE #353: Feature request: add collector accumulator
A simple and widely useful accumulator would be the **Collector** (name to be refined if needed). The accumulator holds a std::vector and appends sample values to the vector. Users should be able to view the contents as a numpy array.

## Motivation

The usual accumulators were designed to have a very small state, so that one can have very many of them. This accumulator is coming from other end of the spectrum, it uses the maximum amount of storage to hold all samples which ended up in a certain bin. Having the full sample of values in each bin is useful in a variety of contexts, to do unbinned fits in each bin, to compute the median, to compute a kernel density estimate.

## Technical challenges?

Do we need a custom view again for the accumulator or maybe just a buffer interface? It would be nice if a collector instance would act like a normal numpy array (at least read-only, possibly read-write), which means it should support slicing, masking, advanced indexing access, can be passed to numpy ufuncs etc.

henryiii:

## Summary

Adds the two collector storages requested in #353: `bh.storage.Collector` and `bh.storage.WeightedCollector`. Both keep the **original per-bin entries** (a variable-length list per bin) instead of aggregating them — `Collector` stores the sample values, and `WeightedCollector` stores `(value, weight)` pairs, the analogs of `weighted_sum` and `weighted_mean` from Hans's comment on the issue. The C++ `accumulators::collector<>` this builds on has been upstreamed into Boost.Histogram and is already vendored in `extern/`; the weighted variant is a small custom accumulator following the `weighted_mean` pattern.

Closes #353.

```python
import boost_histogram as bh

h = bh.Histogram(bh.axis.Regular(3, 0, 3), storage=bh.storage.Collector())
h.fill([0, 0, 1, 2, 2, 2], sample=[1., 2., 3., 4., 5., 6.])
h.view()   # object array: [array([1., 2.]), array([3.]), array([4., 5., 6.])]
h[0]       # [1.0, 2.0]

hw = bh.Histogram(bh.axis.Regular(3, 0, 3), storage=bh.storage.WeightedCollector())
hw.fill([0, 0, 1], sample=[1., 2., 3.], weight=[0.5, 1.5, 2.0])  # weight optional (defaults to 1)
hw.view()[0]            # array([(1., 0.5), (2., 1.5)], dtype=[('value', '<f8'), ('weight', '<f8')])
hw.view()[0]["value"]   # array([1., 2.])
hw[0]                   # [(1.0, 0.5), (2.0, 1.5)]
```

## Design

- `Collector` is `dense_storage<accumulators::collector<std::vector<double>>>` (double only), filled through a new `fill_impl` overload for `accumulator_traits_holder<false, const double&>`. Weighted fills raise.
- `WeightedCollector` is `dense_storage<accumulators::weighted_collector<double>>`, a custom accumulator (`include/bh_python/accumulators/weighted_collector.hpp`) holding a contiguous `std::vector<entry>` with `entry = {double value; double weight;}`. Its two `operator()` overloads mirror `weighted_mean`, so it takes the **existing** weighted fill path: `sample=` required, `weight=` optional (stored as 1.0 when omitted, scalars broadcast).
- Because per-bin data is ragged, there is **no buffer protocol**. `view()` returns a NumPy **object-dtype** array (shape = axes); each element is a copy of that bin's entries — a 1-D `float64` array for `Collector`, a 1-D structured array with `value`/`weight` fields (registered via `PYBIND11_NUMPY_DTYPE`) for `WeightedCollector`. A shared `register_collector_histogram<S>` template registers both, replacing `def_buffer` with a `bh::indexed`-based object-array builder.
- **Works** (all via the C++ backend, all concatenating where bins merge): integer/slice `__getitem__`, `project`, slicing/factor-rebin `reduce`, `h1 + h2` / `+=`, `sum`, pickle, copy/deepcopy.
- **Raises `NotImplementedError`** — the object view returns copies, so anything that writes back through it is unsupported: item assignment, array/scalar arithmetic (`h * 2`), group rebinning, integer picking on a subset of axes, and list-based selection. Threaded fills raise too (merging would scramble the collection order), as do weighted fills on plain `Collector`.
- Pickle stores a per-bin counts array plus flat content arrays (one for `Collector`; values + weights for `WeightedCollector`) — the offsets+content layout a future `to_awkward()` would need.

## Out of scope / follow-ups

- Awkward Array conversion (the serialization layout is the groundwork).
- A writable ragged view (would re-enable setitem / group rebin).
- Non-`double` element types.

Not UHI-serializable, consistent with `MultiCell`.

## Tests

- `tests/test_collector.py` (23 tests) and `tests/test_weighted_collector.py` (26 tests): fill, object-array views incl. flow and structured `value`/`weight` fields, weight-default/broadcast, 2D, indexing, `sum`/`project`/`reduce`/factor-rebin concatenation, addition, pickle round-trip, copy/deepcopy, reset, equality (incl. `Collector` vs `WeightedCollector`), structural match, and all the unsupported-op guards.
- Full suite: **1004 passed, 1 pre-existing xfail**. `prek -a` clean (ruff, clang-format, clang-tidy, mypy).


henryiii:
GPT 5.5 review from copilot cli below.

 2.0m (1.9m cached) • ↓ 6.3k (1.2k reasoning)

**Code review found one high-severity issue.**

`include/bh_python/storage.hpp:241` and the analogous `weighted_collector` load path trust serialized `counts` without checking non-negativity or that `sum(counts)` matches the serialized `values`/`weights` lengths. A corrupt or malicious pickle can make native code read past NumPy buffers during `__setstate__`.

Recommended fix: validate counts before constructing cells: reject negative counts, check accumulation overflow, and require the accumulated total to exactly equal `values.size()` and, for weighted collectors, `weights.size()`.

henryiii:

Fixed in f119f5f.

The `collector` and `weighted_collector` load paths now validate the serialized per-bin `counts` before constructing any cells: negative counts are rejected, and the accumulated total must exactly equal the value buffer length (the running comparison also guards against overflow). For `weighted_collector`, the value and weight buffers are additionally required to be the same length. Any violation throws before native code touches the buffers.

Added `test_pickle_rejects_corrupt_counts` to both `tests/test_collector.py` and `tests/test_weighted_collector.py`, parametrized over negative, too-large, and too-small counts (each tampers with the int64 counts array in real pickle bytes and asserts unpickling raises).

PR #1137, opened 2026-06-09
feat: addable Mean/WeightedMean views (#276)

### LINKED ISSUE #276: addable weight view?
I am not 100 % sure we should really do this, but it would be handy if the view of a histogram with weight storage would allow adding views and in general all the operators that also the histogram itself supports. The record array in numpy does not support addition or other operators, but since we are wrapping it, we could add support for this special record array.

henryiii:

Closes #276.

Histogram views now support the same arithmetic the histograms themselves do — "all operations a histogram supports should also be supported by accumulator views." PR #368 gave `WeightedSumView` full ufunc support; this fills the remaining gap: `MeanView` and `WeightedMeanView` previously had **no** arithmetic, only a read-only `variance` property.

At the C++ level, `Mean`/`WeightedMean` histograms already combine via `__iadd__` (the Welford/West merge), and the accumulators define `operator*=` — but the views exposed none of it, so `h1.view() + h2.view()` did not match `(h1 + h2).view()`.

### What changed (`src/boost_histogram/view.py`)

- **Hoisted the `__array_ufunc__` scaffolding onto the base `View`** (the `out=` allocation, dtype-based view/view-vs-scalar branching, reduce/accumulate, passthrough), dispatching the field math to per-view hooks: `_unary` / `_combine` / `_add_scalar` / `_scale` / `_reduce` / `_accumulate`, gated by `_SUPPORTS_SUB` / `_SUPPORTS_VIEW_PRODUCT`. `WeightedSumView` is migrated onto this machinery with **no behavior change** (its existing tests pass untouched).
- **Added a `_MeanArithmetic` mixin** shared by `MeanView` and `WeightedMeanView`:
  - `+` / `+=` / `.sum()` / `view + accumulator` → the vectorized parallel merge, mirroring the C++ `mean`/`weighted_mean` `operator+=`. Empty bins (`weight == 0`) are zero-guarded with `np.divide(..., where=...)` so they yield `0`, not `NaN` (important under `filterwarnings = error`). All fields are computed into temporaries before writing, so in-place `v += other` is correct despite `out` aliasing `v`.
  - scalar `*` / `/` → scales `value` by `s` and the deltas by `s²`, weights untouched (mirrors `operator*=`).
  - Subtraction, scalar-add, unary `-`, and view×view are rejected with clear `TypeError`s (means have no `operator-=`, and view ratios are out of scope).

### Tests (`tests/test_views.py`)

14 new tests: combine matches `(h1+h2).view()` field-for-field (Mean and WeightedMean, with empty bins), zero-count guard (no `NaN`, no warning), `view + accumulator`, scalar scaling, `.sum()`/reduce matching `h.project()`, and all rejection cases.

### Verification

- `tests/test_views.py`: 27 passed
- Full suite: 978 passed, 1 pre-existing xfail — confirms view-level `+` now matches histogram-level `+` (which runs the exact C++ merge) including empty-bin handling
- `mypy` clean; `prek -a` clean

### Out of scope

`WeightedSumView` view×view / view÷view ratios are intentionally **not** included (they'd go beyond what histograms support and assume independence) — that path raises a clean `TypeError` and can be a separate follow-up. This also lays groundwork for doing cross-storage histogram addition in Python (#261).

### DIFF
--- a/src/boost_histogram/view.py
+++ b/src/boost_histogram/view.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from collections.abc import Callable, Mapping, MutableMapping
-from typing import Any, ClassVar, Literal
+from typing import TYPE_CHECKING, Any, ClassVar, Literal
 
 import numpy as np
 
@@ -15,6 +15,11 @@ class View(np.ndarray[Any, Any]):
     __slots__ = ()
     _FIELDS: ClassVar[tuple[str, ...]]
     _PARENT: type[WeightedSum | WeightedMean | Mean]
+
+    # Whether ``a - b`` (and unary ``-``) and multiplying/dividing two views are
+    # meaningful for this view. Overridden per subclass.
+    _SUPPORTS_SUB: ClassVar[bool] = False
+    _SUPPORTS_VIEW_PRODUCT: ClassVar[bool] = False
 
     def __getitem__(self, ind: StrIndex) -> np.typing.NDArray[Any]:  # type: ignore[override]
         sliced = super().__getitem__(ind)  # type: ignore[index]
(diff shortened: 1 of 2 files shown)


henryiii:
:robot: Kimi K2.6 review :robot:

Overall, this is a well-structured and correct PR. The core logic is sound, the tests are thorough, and the refactoring to hoist `__array_ufunc__` onto `View` is clean.

### Issues (None blocking)

**1. `np.floor_divide` on mean views is allowed but untested and semantically questionable**

`_scale` matches `np.floor_divide` alongside `/` and `*`. For floating-point means, `//` truncates the mean and deltas (twice), which is not a mathematically meaningful scaling. This is inherited from the existing `WeightedSumView` ufunc match list, but the new `_scale` implementation for means implicitly accepts it. Consider explicitly rejecting `floor_divide` in `_MeanArithmetic._scale`, or at least add a test if it is intentional. This is low severity — it does not crash, but the result is mathematically suspect.

### Suggestions

**2. `_reduce` in `_MeanArithmetic` does not handle `keepdims=True`**

The `combined_b = combined if axis is None else np.expand_dims(combined, axis)` logic assumes `keepdims=False` (numpys default). If a caller passes `np.add.reduce(view, keepdims=True)`, the broadcast expansion will add a redundant dimension. `WeightedSumView` didnt handle this either (reduction was a generator wrapping `_PARENT._make`), but with the new `_MeanArithmetic._reduce` implementing custom reduction formulas, this edge case is newly exposed for `MeanView` and `WeightedMeanView`. Consider passing `**kwargs` into the `np.add.reduce` and `np.where` calls (which you do for the reductions themselves), but also accounting for `keepdims` in the `combined_b` broadcast step.

**3. The `_scale` double-ufunc idiom is clever but fragile**

```python
out[deltas] = ufunc(ufunc(view[deltas], scalar), scalar)
```

This relies on `np.multiply` / `np.divide` being idempotent in terms of operator identity for variance scaling. It works correctly for `*` (variance → `*s*s`) and `/` (variance → `/s/s`), but assumes `ufunc` is not something exotic. Since the ufunc match in `__array_ufunc__` tightly constrains what reaches `_scale`, this is safe in practice, but a brief inline comment explaining why `ufunc(ufunc(x, s), s)` equals `x * s²` (or `x / s²`) would help future readers.

### Verification

- All 27 tests in `test_views.py` pass ✅
- `mypy` and `pre-commit` were noted clean in the PR description ✅
- The Welford merge correctly zero-guards empty bins (verified by `test_mean_view_zero_count_bins`) ✅

### Conclusion

No bugs found. The PR is ready to merge once the author addresses the minor suggestions above at their discretion.

henryiii:

Thanks for the review! Addressed all three in b9ec13c:

**1. `floor_divide` on mean views** — now explicitly rejected. Floor-dividing a floating-point mean (and its deltas, twice) isn't a meaningful scaling, so `_MeanArithmetic._scale` raises `TypeError` for `np.floor_divide` regardless of operand order. Added `test_mean_view_rejects_floor_division`. (`WeightedSumView` keeps its existing `//`-by-scalar behavior — unchanged.)

**2. `keepdims` in `_reduce`** — now honored. The reduce reads `keepdims` from kwargs and passes it through, and the spread-term broadcast now uses a dedicated `keepdims=True` reduction for the combined mean instead of `np.expand_dims(combined, axis)`. That makes it correct for any `axis` (int, **tuple**, or `None`) and independent of the caller's `keepdims`. Added `test_mean_view_reduce_keepdims` (2-D, `axis=0`, checks both `(2,)` and `(1, 2)` shapes). Verified `v.sum()` still returns a scalar `Mean`/`WeightedMean` accumulator (no regression).

**3. `_scale` double-ufunc idiom** — expanded the inline comment to spell out why `ufunc(ufunc(x, s), s)` gives `x·s²` (multiply) / `x/s²` (divide), i.e. variance scales by `s²`.

All 29 `test_views.py` tests pass; full suite 980 passed / 1 xfail; mypy + prek + pylint clean.

Issue #1143, opened 2026-06-11
Codebase review: bugs, performance, and modernization findings


henryiii:
:robot: _AI text below_ :robot:

A multi-agent review of the codebase (Python layer, C++ bindings, build/tooling) for bugs, performance, simplifications, and modernizations. Every finding marked ✅ was reproduced against the current `develop` build; the rest were verified by code reading. Draft PRs addressing groups of these will reference this issue.

## A. Memory-safety bugs (C++ bindings)

- [ ] **A1** ✅ `include/bh_python/array_like.hpp:15` + `register_axis.hpp:79-90` — `array_like<int>` copies the *input's* strides onto the freshly allocated output, but `vectorize_index` writes linearly in C order. `IntCategory([1,2,3]).index(np.asfortranarray([[1,2],[3,1]]))` → `[[0,2],[1,0]]` (wrong); negative-stride input → garbage (`[0, 0, 1600087412, ...]`), heap OOB. Fix: allocate plain C-contiguous output with the input's shape.
- [ ] **A2** ✅ `include/bh_python/vector_string_caster.hpp:41-72` — walks the buffer with `p += itemsize`, ignoring strides. `StrCategory(['a','b','c']).index(np.array(['a','b','c','a'])[::2])` → `[0, 0]` instead of `[0, 2]`. This caster feeds `fill()`, so strided string arrays produce **silently wrong histograms**; `[::-1]` reads out of bounds. Fix: contiguous fallback copy.
- [ ] **A3** ✅ `register_histogram.hpp:124-130` / `histogram.hpp:77-99` — a previously returned `h.view()` dangles (read *and write* of freed memory) after a growing-axis fill reallocates storage; numpy `base` keeps the histogram alive but not the buffer. Hard to fix fully; needs at least documentation, or copying views of growable histograms.
- [ ] **A4** `register_histogram.hpp:189-205` — `reduce`/`project` release the GIL, then Boost copies axes including `py::dict` metadata → `Py_INCREF` without the GIL (the known TODO behind `PYBIND11_NO_ASSERT_GIL_HELD_INCREF_DECREF`, `pybind11.hpp:12-13`). Fix: copy axes under the GIL, release only around storage work. Related: under `py::mod_gil_not_used()` the unlimited-storage buffer conversion (`histogram.hpp:84-99`) mutates/frees the buffer on a read path with no lock — concurrent `.view()` calls race in 3.14t.
[...]

PR #1145, opened 2026-06-11
chore(build): pybind11 v3 pin, MSVC /WX, config and stub cleanups

henryiii:
Undoing C++17, pybind11 3 does not require C++17, I would know. :)

Part of #1143.

Build/tooling fixes verified against the sources:

- **E1 — pybind11 pin mismatch (CMakeLists.txt):** the FetchContent fallback pinned pybind11 v2.13.3 while `pyproject.toml` requires `pybind11>=3`, so pure-CMake builds (including the CMake CI job) built a different pybind11 major than shipped wheels. Now pins the v3.0.4 commit (`d03662f`) and uses `FIND_PACKAGE_ARGS 3` as a version floor. Also sets `CMAKE_CXX_STANDARD` to 17 — pybind11 3 mandates C++17, so wheels already compile as 17; the declared standard is now honest. (No C++ header modernization here; that is deferred.)
- **E2 — MSVC warnings-as-errors (CMakeLists.txt):** `BOOST_HISTOGRAM_ERRORS=ON` passed `/WE`, which is not an MSVC flag, so it silently did nothing. Fixed to `/WX`.
- **E3 — stub drift in `src/boost_histogram/_core/accumulators.pyi`:** `WeightedSum.fill` now declares `variance` keyword-only (matches `py::kw_only()` in `register_accumulators.cpp`); `Mean.__call__`/`WeightedMean.__call__` take a scalar `float` (the binding is `make_mean_call`, which binds a `double`), not `ArrayLike`; `_ipython_key_completions_` moved off `_BaseAccumulator` onto WeightedSum/Mean/WeightedMean — `Sum` never had it in C++.
- **E4 — pyproject.toml cleanup:** stray leading space in the Chat URL; dead sdist excludes for `extern/pybind11` (not a submodule); dead `gp311_242-*` cibuildwheel test-skip identifiers; `cp310-musllinux_*` subsumed by `cp31*-musllinux_*`; "Hypothosis" typo.
- **E5 — CI/nox:** `examples.yml` push trigger now includes `develop` (the active default branch); removed `--ignore=docs/.build` from the nox docs session (nonexistent dir — outdir is `docs/_build`, which sphinx-autobuild auto-ignores).

Verification: `uv sync --reinstall-package boost-histogram` rebuilds cleanly with C++17 + pybind11 3; pure-CMake configure fetches pybind11 v3.0.4 at the pinned commit; full test suite passes (997 passed, 1 xfail); `mypy` strict clean; `prek -a` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


CMakeLists.txt
```diff
@@ -5,10 +5,10 @@ project(BOOST_HISTOGRAM LANGUAGES CXX)
 
 include(FetchContent)
 
-# Boost histogram requires C++14
+# pybind11 3 requires C++17
 set(CMAKE_CXX_STANDARD
-    14
-    CACHE STRING "The C++ standard to compile with, 14+")
+    17
+    CACHE STRING "The C++ standard to compile with, 17+")
```
henryiii:
```suggestion
# Boost histogram requires C++14
set(CMAKE_CXX_STANDARD
    14
    CACHE STRING "The C++ standard to compile with, 14+")
```

PR #1159, opened 2026-07-14
fix: MultiCell loses contents when a growing axis resizes storage

NJManganelli:
### Problem

Filling a `MultiCell` histogram that has a growing axis (e.g. `StrCategory(growth=True)`)
silently zeroes everything filled before the axis grows. Any category that appears
after the first fill triggers a storage resize, and all previously accumulated cells
are lost. The only workaround today is to declare every category up front.

```python
import boost_histogram as bh, numpy as np

def make(cats):
    return bh.Histogram(bh.axis.StrCategory(cats, growth=True),
                        bh.axis.Regular(3, 0, 3),
                        storage=bh.storage.MultiCell(2))

grown, pre = make([]), make(["a", "b", "c"])
for chan, xs, cells in [("a",[.5,1.5,2.5],[[1,10],[2,20],[3,30]]),
                        ("b",[.5,2.5],[[4,40],[5,50]]),
                        ("c",[1.5],[[6,60]])]:
    grown.fill(chan, np.asarray(xs), weight=np.asarray(cells, float))
    pre.fill(chan,  np.asarray(xs), weight=np.asarray(cells, float))

grown.view().sum()  # 13.0 + 130.0  -> wrong, most of it zeroed on resize
pre.view().sum()    # 28.0 + 280.0  -> correct
```

### Cause

`multi_cell_reference` is a proxy reference (a `span` over a cell's elements). Its
element-wise assignment is provided only as a *template* `operator=`. A templated
assignment operator does **not** suppress the implicitly-declared copy-assignment
operator, so when both sides have the same type — `multi_cell_reference = multi_cell_reference` —
overload resolution prefers the compiler-generated copy-assignment, which just rebinds
the span instead of writing through to the referenced cells.

boost::histogram relocates cells with `*new = *old` in
`detail::storage_grower::apply` when a growing axis enlarges the storage. With the
rebinding assignment the relocation is a no-op, so every growth discards the existing
contents. (Confirmed by reproducing against pristine boost::histogram `develop`: the
cell-relocation offsets are computed correctly, but `*ns = x` never copies.)

### Fix

Provide an explicit write-through copy-assignment operator for `multi_cell_reference`,
and default the copy constructor to keep proxy rebind-on-copy semantics. This is the
same contract every other boost::histogram storage reference already honours.

Adds a regression test that fills a growing `MultiCell` histogram category by category
and checks it matches the pre-declared-categories result. Full test suite passes; clean
under `-Wall -Wextra` and clang-format.

Follow-up to #1008 (MultiCell storage).

🤖 Investigated and drafted with Claude Code; reviewed by @NJManganelli.

### DIFF
--- a/include/bh_python/multi_cell.hpp
+++ b/include/bh_python/multi_cell.hpp
@@ -37,6 +37,10 @@ struct multi_cell_reference : public multi_cell_base<T, boost::span<T>> {
     // using boost::span<T>::span;
     using multi_cell_base<T, boost::span<T>>::multi_cell_base;
 
+    // Copy construction rebinds the view (proxy-reference semantics), so that a
+    // reference refers to the same cells as the one it was copied from.
+    multi_cell_reference(const multi_cell_reference&) = default;
+
     void operator()(const boost::span<T> values) { operator+=(values); }
 
     void operator+=(const boost::span<T> values) {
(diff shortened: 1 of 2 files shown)


henryiii:
This is draft, what's the status?

PR #1175, opened 2026-09-02
fix: rank 0 histogram UBs + add windows 3.14t testing

ikrommyd:
This should be the crash we're seeing in integration tests in windows 3.14t ci. boost-histogram did not exercise rank-0 project and reduce in tests. Reproducing UB is hard and I got no windows machine. I did this only from inspection. The test that's crashing in the integration tests in hist is
```
def test_T_empty():
    hist_empty = hist.Hist()
    hist_T_empty = hist_empty.T
    assert hist_empty == hist_T_empty
```
and `.T` is `return self.project(*reversed(range(self.ndim)))` which is `self.project()` so it must be the rank 0 projection.

Adding windows 3.14t testing to the ci too which is what caught the UB in the integration tests repo.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -138,6 +138,8 @@ jobs:
             only: cp310-win32
           - os: windows-latest
             only: cp313-win_amd64
+          - os: windows-latest
+            only: cp314t-win_amd64
           - os: macos-15-intel
             only: cp314t-macosx_x86_64
           - os: macos-latest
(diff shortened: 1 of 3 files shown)


ikrommyd:
cc @henryiii

henryiii (approved):

ikrommyd:
We did not hit the seg fault in this mornings integration tests CI so that should have been the right fix indeed.

